PR #2, opened 2026-06-14
Fix 68000 register shift/rotate cycle counts

LinuxJedi:
I've been using this crate as the CPU core in an Amiga emulator, where 68000 cycle timing matters, and noticed the register shift/rotate instructions are billed slightly wrong on the 68000.

Two issues:

1. **Long base is 8, not 6.** Register shifts/rotates cost `base + 2 * count`, and the base is 6 for byte/word but 8 for long. The core used 6 for all sizes, so every `ASL/ASR/LSL/LSR/ROL/ROR/ROXL/ROXR .l` came out two clocks short.

2. **ROXL/ROXR counted the reduced rotate count.** The cycle count was taken after the count had been reduced modulo `(bits + 1)`, so e.g. a word `ROXL` by 17 was billed as a no-op (the base only) instead of `6 + 2*17`. The rotation itself does wrap every `(bits + 1)` positions because X acts as an extra bit, but the timing follows the full shift count.

Both are gated to the 68000, so 68020+ are untouched.

### Validation

Checked the long forms against the SingleStepTests `m68000` set. Before this change `ASL.l/LSL.l/ROL.l/ROXL.l` mismatched the fixture cycle count in 100% of cases; after, they match exactly. Byte/word register forms were already correct and stay correct (the residual on `ASL.w` etc. is the memory `<ea>` shift-by-one form, whose EA cost this core doesn't model for any instruction).

Added `tests/cycle_timing_shift_68000_tests.rs` covering the long base, the 68020 no-change case, and the ROXL/ROXR full-count timing. The existing SingleStepTests functional suite still passes (this only changes returned cycle counts, not results).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -296,7 +296,7 @@ checksum = "616ec5685824bcc94416c6d4a7a446eea774a31efd7062c8480ba6fd06d7a6e5"
 
 [[package]]
 name = "m68k"
-version = "0.1.13"
+version = "0.1.14"
 dependencies = [
  "cranelift-codegen",
  "cranelift-frontend",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "m68k"
-version = "0.1.13"
+version = "0.1.14"
 edition = "2024"
 license = "MIT"
 description = "A safe Rust M68000 family CPU emulator"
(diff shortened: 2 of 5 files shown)


benletchford:
Thanks @LinuxJedi, can you take a look at the failing unit test and submit a commit / fix for this?

LinuxJedi:
@benletchford looks like you beat me to it :)

PR #3, opened 2026-06-14
Data-dependent 68000 multiply/divide timing

LinuxJedi:
Follow-on from #2, same motivation (driving this core in an Amiga emulator). `MULU/MULS/DIVU/DIVS` currently return a flat cycle count regardless of the operands, but on the 68000 they're all data-dependent:

```
MULU.W = 38 + 2 * (number of 1 bits in the source word)
MULS.W = 38 + 2 * (number of bit transitions in source << 1)
DIVU.W = restoring-division loop (early-overflow path is short)
DIVS.W = signed restoring-division loop
```

These match the algorithms MAME/Musashi use and are the dominant error term (a `MULU` of a 16-bit value can be off by up to ~32 clocks with the flat 38).

The change keeps the existing flat values for 68020+ and only switches to the data-dependent path on `CpuType::M68000`.

### Validation

Against the SingleStepTests `m68000` fixtures, for register-direct source operands (no EA cost in play):

| op | before | after |
|------|--------|-------|
| MULU | 100% mismatch | 0% |
| MULS | 100% mismatch | 0% |
| DIVU | 100% mismatch | 0% |
| DIVS | 100% mismatch | 23% within +/-2, mean ~0 |

DIVS isn't bit-exact in ~1/4 of cases (the signed loop has a small data-dependent wobble that averages out), but it's a big improvement on the flat value and the overflow/early-out paths are exact. Happy to iterate on it if you'd prefer.

The all-modes numbers don't reach 0% because the remaining difference is the effective-address fetch cost, which this core doesn't add for any instruction; that's out of scope here.

Added `tests/cycle_timing_muldiv_68000_tests.rs` for the bit-count/transition formulas, the DIVU/DIVS overflow paths, and the 68020 flat-value fallback. The SingleStepTests functional suite is unaffected (results are unchanged; only the cycle count differs).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -296,7 +296,7 @@ checksum = "616ec5685824bcc94416c6d4a7a446eea774a31efd7062c8480ba6fd06d7a6e5"
 
 [[package]]
 name = "m68k"
-version = "0.1.14"
+version = "0.1.15"
 dependencies = [
  "cranelift-codegen",
  "cranelift-frontend",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "m68k"
-version = "0.1.14"
+version = "0.1.15"
 edition = "2024"
 license = "MIT"
 description = "A safe Rust M68000 family CPU emulator"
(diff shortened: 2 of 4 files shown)


benletchford (approved):
Reviewed locally after rebasing onto master. Added a DIVS timing sign-adjustment fix and the 0.1.15 version bump; CI is green.

PR #18, opened 2026-07-23
perf: inline memory operation internals

rlanday:
## Stack

Depends on #17, which depends on #16, #15, #14, and #13.

This branch starts at the head of #17. Until the lower PRs merge, GitHub will show their prerequisite commits in this PR. As each lower PR merges, this branch will be rebased so the final incremental diff contains only a90d60a.

## Summary

- require deferred address-register lookup, staging, and commit helpers to inline into generic effective-address resolution
- require binary memory-operation result application to inline into the decoded memory dispatcher
- preserve the existing fallback-before-commit contract and all guest-visible behavior

## Motivation

After #17 inlined generic effective-address resolution and location loads/stores, fresh Lemmings profiles showed the remaining small `Pending::{a,commit}` and `apply_binary_to_reg` call boundaries below `execute_mem_op`. These helpers operate on per-instruction local state and are called from operation-specific match arms, so inlining lets the optimizer specialize zero-, one-, and two-update cases and combine result handling with the surrounding dispatcher arm.

No addressing rule, alignment check, memory-window check, register update order, flag calculation, instruction count, or synthetic cycle count changes.

## Performance

Intel MacBookPro16,1, native x86-64 macOS optimized builds.

### Generic-memory benchmark

Five-run medians from the permanent benchmark added in #17:

```console
cargo bench --bench microbench -- generic-memory
```

| Workload | #17, 128a6a2 | This change, a90d60a | Change |
| --- | ---: | ---: | ---: |
| `TST.B (A0)` | 46.4 MIPS | 57.6 MIPS | +24.1% |
| `TST.B (A0)+` | 43.8 MIPS | 50.5 MIPS | +15.3% |
| Indexed `TST.B` | 40.6 MIPS | 46.6 MIPS | +14.8% |
| `ADD.W (A0),D0` | 39.4 MIPS | 46.6 MIPS | +18.3% |

The pending-address annotations alone improved all four rows by 7.6% to 19.4%. Adding binary-result inlining produced a further 4% to 9% improvement.

### Systemless and Lemmings

Matched 30-second `/usr/bin/sample` captures used Lemmings 1.5.2 level one after a five-second warm-up. Results are normalized by total main-thread samples.

| Revision | `run_steps_internal` samples | Main-thread samples | Normalized share |
| --- | ---: | ---: | ---: |
| #17, 128a6a2 | 349 | 22,142 | 1.576% |
| a90d60a | 235 | 24,372 | 0.964% |

Normalized emulator time fell by 38.8%.

Within the same captures, `execute_mem_op` fell from 211 / 22,142 samples (0.953%) to 152 / 24,372 (0.624%), a 34.5% normalized reduction. The targeted helper symbols disappeared from the after-profile.

Reproducibility hashes for the complete Systemless release binaries:

- #17 baseline: `3b32d0326572c7cbe31892f212d5a109c67376a674be736ee5f023abad3207c6`
- candidate: `e73795617b95dc3f2af098bf5f6b648245153b3ff5f0945ffb663b760ce009da`

## Rejected alternatives

- forcing the complete `execute_mem_op` dispatcher to inline regressed indexed and arithmetic benchmark rows by 2% to 6%
- forcing `brief_index` to inline gained only about 2.5% on indexed access while regressing an independent arithmetic row by about 5.6%
- forcing `test_condition` to inline regressed conditional-branch medians by about 4% to 6%

None of those experiments is included in this branch.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo clippy --all-targets -- -D warnings`
- complete `cargo test`, including 127 SingleStep categories and 78 Musashi tests
- repeated optimized `generic-memory` benchmark runs
- Systemless release build and interactive Lemmings level-one profiling

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -147,6 +147,37 @@ fn bench_loop<B: BenchBus>(
     );
 }
 
+fn bench_batch_loop(label: &str, name: &str, words: &[u16], instrs: u32) {
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (i, word) in words.iter().enumerate() {
+        bus.write_word_at((i * 2) as u32, *word);
+    }
+
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = 0;
+        cpu.set_a(5, 0x1000);
+        cpu.set_a(7, 0x8000);
+        cpu
+    };
+
+    let mut warm_cpu = prepare_cpu();
+    let warm = warm_cpu.run_batch(&mut bus, 5_000_000, &[]);
+    assert_eq!(warm.instructions, 5_000_000);
+
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    let result = cpu.run_batch(&mut bus, instrs, &[]);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(result.instructions, instrs);
+    println!(
+        "{label:9} {name:18} {:8.1} M instr/s",
+        instrs as f64 / elapsed / 1_000_000.0
+    );
+}
+
 fn bench_set<B: BenchBus>(label: &str) {
     bench_linear::<B>(label, "linear NOP", 0x4E71, 4, 40_000_000);
     bench_linear::<B>(label, "linear ADDQ", 0x5280, 4, 40_000_000);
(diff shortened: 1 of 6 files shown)


benletchford:
I reviewed and tested the full six-commit stack locally. `cargo fmt --all -- --check`, clippy with warnings denied, all-feature tests (including 127 SingleStep categories, 78 Musashi tests, and the differential fast-memory/JIT coverage), doctests, and `cargo package --locked` all pass. The performance work is well-scoped and preserves the fallback/accounting contracts. I’m marking the two draft prerequisites ready and merging the stack bottom-up.

PR #20, opened 2026-07-23
perf: JIT hot memory-source CMP loops

rlanday:
## Summary

Extend the trace JIT with the two read-only memory-source CMP forms that blocked the hottest Lemmings loops:

- `CMP.<size> (An),Dn`
- `CMP.<size> d16(An),Dn`

The native emitter performs a checked, endian-correct fast-memory load and computes NZVC while preserving X. All bounds/alignment checks happen before architectural state changes, so a side exit can safely re-execute the instruction through normal dispatch. The portable executor uses the same trace representation for wasm and differential tests.

Memory-CMP traces are deliberately admitted only when the recording closes into a true self-loop. Measurements showed that short non-self-loop regions do not amortize fast-memory validation plus the native/Rust round trip; those stay on the decoded-memory path.

This also adds:

- an opt-in `trace-profile` feature, compiled out of normal builds;
- rejected-head/blocker and successful-trace attribution;
- a complete non-self-loop round-trip benchmark; and
- Lemmings-weighted rejected-loop and self-loop benchmarks derived from a real run.

## Why these instructions

A 206,780,516-instruction Lemmings baseline reported:

| Trace head | Dynamic rejected hits | Traceable prefix | Blocker |
|---|---:|---:|---|
| `$0002AD52` | 483,003 | 5 ops | `$B210` — `CMP.B (A0),D1` |
| `$00022DA2` | 399,980 | 4 ops | `$BC6E` — `CMP.W d16(A6),D6` |
| `$0002AF40` | 407,271 | 2 ops | `$B210` — `CMP.B (A0),D1` |

Those three heads stranded about 4.83 million already-supported decoded dispatches in that run.

### Reproducible microbenchmark

Six alternating baseline/candidate runs of `lemmings-self-loops`:

| Build | Median throughput |
|---|---:|
| Baseline | 55.2 M guest instructions/s |
| This PR | 282.4 M guest instructions/s |

That is a **5.1x** improvement for the measured Lemmings instruction mix when it closes into self-loops.

Six alternating runs of the corresponding full non-self-loop validation/entry/exit/reprobe workload:

| Build | Median throughput |
|---|---:|
| Baseline | 53.05 M guest instructions/s |
| This PR | 52.7 M guest instructions/s |

The **-0.7%** difference is noise-level and confirms the topology guard keeps the unprofitable workload on its existing decoded path. The existing `classic Mac mix` row was also neutral in a paired run (284.7 vs. 288.5 M guest instructions/s).

### Real Lemmings trace coverage

The earlier baseline retired 17,587,832 of 206,780,516 guest instructions through native traces: **8.5% JIT coverage**.

A gameplay-focused candidate run retired 37,968,597 of 142,979,534 instructions through native traces: **26.6% coverage**. A much longer candidate run, including menus and between-level work, retired 172,109,406 of 915,992,784 instructions natively: **18.8% coverage**.

The longer run attributed the formerly blocked loops as follows:

| Trace | Compiled ops | Native calls | JIT-retired instructions | Average retired/call |
|---|---:|---:|---:|---:|
| `$00022DA2` | 19 | 366,996 | 3,818,294 | 10.40 |
| `$0002AF40` | 7 | 406,794 | 2,029,092 | 4.99 |
| `$0002AD48` | 7 | 191,462 | 1,311,309 | 6.85 |

`$0002AD48` is the enclosing trace that now absorbs the former `$0002AD52` CMP blocker.

### Matched host profile

I sampled the fixed pre-change binary and this release candidate back-to-back for 30 seconds at 1 ms while level 1 was active with Lemmings walking. Aggregating exclusive samples attributed to the Systemless binary:

| Exclusive samples | Baseline | This PR | Change |
|---|---:|---:|---:|
| All Systemless code | 1,090 | 728 | **-33.2%** |
| Emulator runner | 531 | 367 | **-30.9%** |
| Decoded memory executor | 359 | 209 | **-41.8%** |

The binaries were fixed before sampling (`e7379561...` baseline and `5084b025...` candidate) and used the same Lemmings 1.5.2 disk and Systemless configuration.

## Correctness and safety

- Decoder tests cover both hot opcodes and displacement capture.
- Portable tests verify CMP NZVC behavior, X preservation, and no state change on a failed window check.
- The integration test compares `run_batch` against instruction stepping for both forms.
- Native loads use the existing fast-memory bounds/alignment and big-endian helpers.
- Non-self-loop memory-CMP recordings are rejected before native compilation.

Validation completed:

- `cargo test --all-features --all-targets`
- `cargo clippy --all-features --all-targets -- -D warnings`
- `git diff --check`

After rebasing onto the `m68k 0.2.2` release, the focused integration test and full clippy pass were repeated successfully.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,12 @@ exclude = [
 
 [dependencies]
 
+[features]
+default = []
+# Opt-in trace-opportunity accounting for emulator profiling. Disabled builds
+# contain none of the counters or reporting code.
+trace-profile = []
+
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 cranelift-codegen = "0.132.0"
 cranelift-frontend = "0.132.0"
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -152,6 +152,14 @@ fn bench_batch_loop(label: &str, name: &str, words: &[u16], instrs: u32) {
 }
 
 fn bench_batch_loop_at(label: &str, name: &str, words: &[u16], instrs: u32, code_base: usize) {
+    let elapsed = measure_batch_loop_at(words, instrs, code_base);
+    println!(
+        "{label:9} {name:18} {:8.1} M instr/s",
+        instrs as f64 / elapsed / 1_000_000.0
+    );
+}
+
+fn measure_batch_loop_at(words: &[u16], instrs: u32, code_base: usize) -> f64 {
     let mut bus = LinearMemoryBus::new(0x10000);
     for (i, word) in words.iter().enumerate() {
         bus.write_word_at((code_base + i * 2) as u32, *word);
@@ -162,8 +170,11 @@ fn bench_batch_loop_at(label: &str, name: &str, words: &[u16], instrs: u32, code
         cpu.set_cpu_type(CpuType::M68040);
         cpu.set_sr(0x2700);
         cpu.pc = code_base as u32;
+        cpu.set_a(0, 0x4000);
         cpu.set_a(5, 0x1000);
+        cpu.set_a(6, 0x5000);
         cpu.set_a(7, 0x8000);
+        cpu.set_d(7, 1);
         cpu
     };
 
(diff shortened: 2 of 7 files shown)


benletchford (approved):
Reviewed the decoder, portable path, Cranelift emitter, trace admission guard, profiler gating, and differential coverage. Bounds/alignment checks precede architectural state changes, the CMP path preserves X and computes NZVC correctly, and non-self-loop memory-CMP regions remain excluded. The exact CI gate passes locally: rustfmt, clippy with `-D warnings`, all-feature fixture/test/doc suites, and locked packaging. Worth merging.

PR #21, opened 2026-07-23
perf: JIT two-instruction self-loops

rlanday:
## Stack

This PR is stacked on #20 (`perf: JIT hot memory-source CMP loops`). Review the final commit in this branch for the two-op-loop change; once #20 merges, GitHub will automatically remove the parent commit from this PR's diff.

## Summary

Admit two-instruction traces when—and only when—the final control-flow operation closes directly back to the trace head. The existing three-operation minimum remains unchanged for non-self-loop regions, and one-instruction loops remain below the admission threshold.

This targets a canonical 68k pattern:

```text
loop:  MOVE.<size> <source>,<destination>
       DBRA        Dn,loop
```

The native and portable trace executors already implement these memory moves and `DBRA`; the previous global three-op minimum prevented the complete two-op region from reaching either executor.

The PR also extends the opt-in `trace-profile` feature to count decoded memory operations by opcode and `(PC, opcode)` site. These counters and their hot-path hook are fully compiled out of normal builds. A Lemmings-weighted microbenchmark and exact `run_batch`/step integration coverage make the workload reproducible.

## Why this shape

A normal-batching/JIT Lemmings run identified a small number of two-op loops dominating the remaining decoded-memory workload:

| Inner loop site | Body | Dynamic iterations |
|---|---|---:|
| `$0002AB5E` | `MOVE.B D1,(A0)+ / DBRA` | 3,702,308 |
| `$0002AD60` + `$0002ACE8` | `MOVE.B (A4)+,(A0)+ / DBRA` | 3,877,713 |
| `$0002A9FA` + `$0002AA9E` | `MOVE.L (A1)+,(A0)+ / DBRA` | 2,405,305 |
| `$0002318E` | `MOVE.B (A0)+,(A1)+ / DBRA` | 654,377 |

Together these represented roughly 21.3 million decoded operations in the captured interval. The existing three-op JIT minimum excluded all of them even though a self-loop amortizes validation across repeated iterations.

### Reproducible weighted microbenchmark

Five alternating runs used fixed baseline and candidate binaries. The workload resets pointers and the DBRA counter in a small outer loop, keeping each measured inner loop inside fast memory indefinitely. Weights come from the real per-site profile above.

| Lemmings loop shape | Baseline median | Candidate median | Speedup |
|---|---:|---:|---:|
| `MOVE.B D1,(A0)+ / DBRA` | 54.5 MIPS | 149.9 MIPS | **2.75x** |
| `MOVE.B (A4)+,(A0)+ / DBRA` | 48.9 MIPS | 202.8 MIPS | **4.15x** |
| `MOVE.L (A1)+,(A0)+ / DBRA` | 49.2 MIPS | 205.6 MIPS | **4.18x** |
| Weighted total | 49.7 MIPS | 175.6 MIPS | **3.53x** |

Existing three-or-more-op self-loop rows remained noise-level neutral (paired weighted observations: baseline 351.6/358.4 MIPS; candidate 359.7/352.7 MIPS). Non-self-loop CMP and linear decoded-memory workloads do not enter the new admission case; their paired runs showed no consistent directional change.

### Matched Lemmings execution coverage

Baseline and candidate trace-profile runs covered nearly identical hot-site iteration counts (for example, `$23086` executed 7,012,739 vs. 6,988,354 times). That makes the structural comparison phase-matched:

| Metric | Baseline | Candidate | Change |
|---|---:|---:|---:|
| Decoded memory operations | 46,921,106 | 31,219,481 | **-15,701,625 (-33.5%)** |
| JIT-retired instructions | 18,205,771 | 33,972,058 | **+15,766,287 (+86.6%)** |

The newly admitted two-op traces retired:

| Trace head | Native-retired instructions |
|---|---:|
| `$0002AB5E` | 7,412,732 |
| `$0002A9FA` | 2,402,874 |
| `$0002AA9E` | 2,386,626 |
| `$0002ACE8` | 2,095,152 |
| `$0002318E` | 1,153,210 |
| `$0002AD24` | 192,724 |

A matched 30-second macOS `sample` also moved the decoded memory executor from 226 to 204 exclusive samples (**-9.7%**). Total runner-exclusive samples were noisy, and generated Cranelift functions appear as unresolved addresses, so the dynamic decoded/JIT counters above are the more reliable attribution.

One high-volume byte-copy path at `$0002AD60` remains decoded. It is reached through an interior-branch side exit from the existing `$0002AD5C` trace rather than by a backward edge whose target is `$0002AD60`; addressing it requires branch-path/chaining work and is intentionally outside this PR.

## Correctness and scope

- The minimum remains three ops for every non-self-loop trace.
- One-op loops remain rejected.
- Existing fast-memory bounds, alignment, self-modifying-code, instruction-budget, and watched-PC guards are unchanged.
- An integration test compares the two-op memory-copy trace with instruction stepping.
- A feature-gated profiler test asserts that a two-op self-loop actually compiles and retires native instructions.
- Decoded opcode/site counters have no code or storage in builds without `trace-profile`.

Validation completed:

- `cargo test --all-targets --features trace-profile`
- `cargo clippy --all-targets --features trace-profile -- -D warnings`
- `git diff --check`

`cargo check --target wasm32-unknown-unknown` could not run locally because that Rust target is not installed; compilation did not reach project code. The portable trace executor is covered by the feature-enabled native test suite, and CI remains authoritative for the wasm target.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,12 @@ exclude = [
 
 [dependencies]
 
+[features]
+default = []
+# Opt-in trace-opportunity accounting for emulator profiling. Disabled builds
+# contain none of the counters or reporting code.
+trace-profile = []
+
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 cranelift-codegen = "0.132.0"
 cranelift-frontend = "0.132.0"
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -152,6 +152,14 @@ fn bench_batch_loop(label: &str, name: &str, words: &[u16], instrs: u32) {
 }
 
 fn bench_batch_loop_at(label: &str, name: &str, words: &[u16], instrs: u32, code_base: usize) {
+    let elapsed = measure_batch_loop_at(words, instrs, code_base);
+    println!(
+        "{label:9} {name:18} {:8.1} M instr/s",
+        instrs as f64 / elapsed / 1_000_000.0
+    );
+}
+
+fn measure_batch_loop_at(words: &[u16], instrs: u32, code_base: usize) -> f64 {
     let mut bus = LinearMemoryBus::new(0x10000);
     for (i, word) in words.iter().enumerate() {
         bus.write_word_at((code_base + i * 2) as u32, *word);
@@ -162,8 +170,11 @@ fn bench_batch_loop_at(label: &str, name: &str, words: &[u16], instrs: u32, code
         cpu.set_cpu_type(CpuType::M68040);
         cpu.set_sr(0x2700);
         cpu.pc = code_base as u32;
+        cpu.set_a(0, 0x4000);
         cpu.set_a(5, 0x1000);
+        cpu.set_a(6, 0x5000);
         cpu.set_a(7, 0x8000);
+        cpu.set_d(7, 1);
         cpu
     };
 
@@ -179,10 +190,7 @@ fn bench_batch_loop_at(label: &str, name: &str, words: &[u16], instrs: u32, code
     let result = cpu.run_batch(&mut bus, instrs, &[0]);
     let elapsed = start.elapsed().as_secs_f64();
     assert_eq!(result.instructions, instrs);
-    println!(
-        "{label:9} {name:18} {:8.1} M instr/s",
-        instrs as f64 / elapsed / 1_000_000.0
-    );
+    elapsed
 }
 
 fn bench_one_shot_trace(head_ops: usize, instrs: u32) {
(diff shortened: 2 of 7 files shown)


benletchford (approved):
Reviewed the layer on top of #20. The two-op minimum is limited to genuine self-loop topology, one-op and non-self-loop thresholds remain unchanged, and the memory-copy differential/profiler tests cover admission and architectural behavior. Feature-disabled builds retain no profiling state. The cumulative #20/#21 result passes rustfmt, clippy with `-D warnings`, all-feature fixture/test/doc suites, and locked packaging. Worth merging.

PR #22, opened 2026-07-23
perf: adapt JIT traces to dominant branch paths

rlanday:
## Summary

Trace regions follow the first path observed through an interior conditional branch and guard that decision at runtime. That works well while the original path remains representative, but a phase change can make the opposite path dominant. The compiled trace then side-exits on nearly every call and permanently strands the rest of the loop in decoded execution.

This change makes that path prediction adapt once:

- identify partial returns caused specifically by an interior guarded branch, rather than memory-access bailouts;
- accumulate the exact number of native calls ending at each guarded exit (`full_iters + 1`);
- after at least 64 calls, re-record when at least 48 calls (75%) used the opposite path;
- carry a one-re-record budget into the replacement trace to prevent oscillation;
- reset the per-CPU record/probe filters when retiring the stale trace;
- perform adaptation bookkeeping only on guarded exits, leaving correctly predicted traces off this path;
- expose guarded-exit and adaptive-re-record counts in the opt-in trace profiler.

The policy is shared by native Cranelift traces and the portable WebAssembly trace executor.

## Why this matters for Lemmings

The hot trace at `$0002AD5C` was recorded through an uncommon branch path. Its common path side-exited after:

```text
$0002AD5C  CMP.B  (A0),D1
            Bcc    guarded opposite path
```

and left this copy loop interpreted:

```text
$0002AD60  MOVE.B (A4)+,(A0)+
$0002AD62  DBRA   D0,$0002AD5C
```

After one adaptive re-record, `$0002AD5C` becomes the four-instruction common-path loop and executes the copy/DBRA pair inside the trace.

### Reproducible phase-biased microbenchmark

`microbench trace-branch-bias` deliberately records a seven-op uncommon path, then changes state so a four-op CMP/branch/copy/DBRA loop becomes dominant. Five alternating fixed-binary runs produced:

| Build | Median throughput |
|---|---:|
| Parent (`cdc342a`, adaptation disabled) | 22.2 MIPS |
| Candidate | 165.6 MIPS |
| Change | **7.46x** |

The benchmark is permanent so future trace-policy changes keep covering this topology.

### Lemmings trace counters

The before/after runs retired comparable instruction totals (94,732,669 and 97,401,472 respectively):

| Metric | Before | After | Change |
|---|---:|---:|---:|
| `$2AD5C` compiled ops | 5 | 4 | dominant path captured |
| `$2AD5C` native calls | 3,179,114 | 3,081,598 | -3.1% |
| `$2AD5C` JIT-retired ops | 7,758,682 | 11,411,935 | **+47.1%** |
| `$2AD5C` average ops/call | 2.44 | 3.70 | **+51.6%** |
| `$2AD60 MOVE.B` decoded | 2,803,636 | 87,628 | **-96.9%** |
| `$2AD62 DBRA` decoded | 2,803,636 | 87,628 | **-96.9%** |
| All JIT-retired ops | 33,972,058 | 41,385,932 | **+21.8%** |
| All decoded memory ops | 31,219,481 | 26,455,087 | **-15.3%** |

The profiler recorded 459,380 guarded exits at `$2AD5C` over the full run and exactly one adaptive re-record. The large lifetime exit count includes later path variation; the replacement is intentionally not allowed to oscillate.

### 30-second macOS level-1 samples

Matched non-instrumented Systemless samples corroborate the trace counters:

| Top-of-stack symbol | Before | After | Change |
|---|---:|---:|---:|
| `FixtureRunner::run_steps_internal` | 444 | 184 | **-58.6%** |
| `m68k::core::mem_ops::execute_mem_op` | 204 | 80 | **-60.8%** |
| `MacMemoryBus::read_word` | 50 | 5 | **-90.0%** |
| host `memmove` | 148 | 80 | **-45.9%** |

These live-game samples are workload-sensitive, so the deterministic microbenchmark and guest instruction counters are the primary evidence; the host samples show that the improvement survives in the application.

The final policy refinement removed bookkeeping from correctly predicted entries and uses exact native calls rather than Rust entries as the dominance denominator. The committed benchmark and behavior tests exercise that final policy.

## Correctness and stability

- A 100,000-instruction differential test compares the phase-changing batch/JIT path against `step()`, including PC, SR, all address registers, and all data registers.
- A profiler test proves the biased seven-op trace is recorded twice, adaptively replaced once, and ends as the four-op loop.
- A deliberately alternating interior branch proves a 50/50 path does **not** re-record.
- A one-re-record budget prevents repeated path flipping even if a program later changes phase again.
- Memory-check partial exits are excluded from branch adaptation.

Validation completed:

```text
cargo test --all-targets --features trace-profile
cargo clippy --all-targets --features trace-profile -- -D warnings
git diff --check
```

All passed.

## Stack

This PR depends on #21 (`perf: JIT two-instruction self-loops`), which in turn depends on #20. Because the upstream base is `master`, GitHub temporarily displays the parent commits too. The new commit for this PR is:

```text
c9026c3 perf: adapt JIT traces to dominant branch paths
```

Review only the new layer with:

```text
git diff cdc342a...c9026c3
```

Once the parent PRs merge, this PR's visible diff will collapse to the adaptive-path commit.

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -9,7 +9,7 @@ use super::cpu::{CpuCore, NFLAG_SET};
 use super::execute::RUN_MODE_BERR_AERR_RESET;
 use super::mem_ops::{BitSource, DecodedMemOp, FastEa};
 use super::memory::AddressBus;
-use super::op_cache::{BitOp, CachedRunResult, DecodedSimpleOp};
+use super::op_cache::{BinaryOp, BitOp, CachedRunResult, DecodedSimpleOp};
 use super::types::{CpuType, Size};
 #[cfg(not(target_family = "wasm"))]
 use cranelift_codegen::Context;
@@ -233,6 +233,15 @@ pub(crate) enum JitTraceOp {
         size: Size,
         src: JitEa,
         dst: JitEa,
+    },
+    /// Read-only ALU operation from fast memory to a data register. The
+    /// initial decoder deliberately admits only CMP with `(An)`/`d16(An)`
+    /// sources; the generic shape leaves room for other read-only forms.
+    AluMemToReg {
+        op: JitBinaryOp,
+        size: Size,
+        src: JitEa,
+        dst: u8,
     },
     /// Hot Classic Mac memory forms using the A5-relative global-data
     /// convention. These require extension words, so they are represented
@@ -540,6 +549,8 @@ impl TraceJit {
                     execute_portable_trace(cpu, &trace.ops, trace.code_start, trace.code_end);
                 cycles_total += (packed as u32) as i64;
                 let ops_done = (packed >> 32) as u32;
+                #[cfg(feature = "trace-profile")]
+                super::trace_profile::note_native_call(pc, cpu_type, ops_done);
                 retired += ops_done;
                 if ops_done < ops_len {
                     // A memory check bailed before its op, or a guarded
@@ -582,6 +593,8 @@ impl TraceJit {
                     cpu_type,
                     ops: Vec::with_capacity(TRACE_MAX_OPS),
                 });
+                #[cfg(feature = "trace-profile")]
+                super::trace_profile::note_recording(pc, cpu_type);
                 cpu.trace_recording = true;
                 None
             }
@@ -631,6 +644,17 @@ impl TraceJit {
         }
     }
 
+    #[cfg(feature = "trace-profile")]
+    fn is_rejected(&self, pc: u32, cpu_type: CpuType) -> bool {
+        matches!(
+            &self.slots[trace_cache_index(pc)],
+            TraceSlot::Rejected {
+                pc: rejected_pc,
+                cpu_type: rejected_type,
+            } if *rejected_pc == pc && *rejected_type == cpu_type
+        )
+    }
+
     fn reject_recording(&mut self, cpu: &mut CpuCore) {
         if let Some(recording) = self.recording.take() {
             let idx = trace_cache_index(recording.start_pc);
@@ -661,6 +685,8 @@ impl TraceJit {
         let idx = trace_cache_index(recording.start_pc);
         let start_pc = recording.start_pc;
         let cpu_type = recording.cpu_type;
+        #[cfg(feature = "trace-profile")]
+        let recorded_ops = recording.ops.len();
         self.slots[idx] =
             match self.compile_decoded_ops(cpu, start_pc, cpu_type, recording.ops, Some(exit_pc)) {
                 Some(trace) => TraceSlot::Compiled(trace),
@@ -672,6 +698,10 @@ impl TraceJit {
                     }
                 }
             };
+        #[cfg(feature = "trace-profile")]
+        if matches!(self.slots[idx], TraceSlot::Compiled(_)) {
+            super::trace_profile::note_compiled(start_pc, cpu_type, recorded_ops);
+        }
     }
 
     fn record_executed<B: AddressBus>(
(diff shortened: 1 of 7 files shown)


src/core/trace_jit.rs
```diff
@@ -554,7 +601,32 @@ impl TraceJit {
...
+            {
+                // A guarded exit ends this Rust entry, so `full_iters + 1`
+                // is the exact number of native calls since entry. Updating
+                // only here keeps correctly predicted traces off the
+                // adaptation hot path while still distinguishing dominant
+                // side exits from genuinely alternating branches.
+                let calls = trace
+                    .adaptive_calls
+                    .get()
+                    .saturating_add(full_iters.saturating_add(1));
```
benletchford: `full_iters + 1` counts only calls in the current entry that ends in a guarded exit. Successful calls from earlier entries are never added. For a non-self-loop trace, each successful entry runs once and leaves this block without updating `adaptive_calls`, while every rare mismatch adds one call and one exit. Consequently 64 mismatches—even if separated by thousands of successful calls—look like 64/64 and trigger a rerecord. I confirmed this with a guard that exits 1% of the time; it still consumed the rerecord budget.

benletchford (changes requested):
The adaptive rerecord idea is worthwhile, but the current denominator does not implement the stated 75% policy for non-self-loop traces. I reproduced this with a four-op non-self-loop trace whose interior guard exits only once every 100 iterations: after 50,000 instructions the profiler still reports `adaptive_rerecords = 1`. Please include successful trace calls across entries in the denominator, or otherwise scope the policy so rare exits cannot consume the one-rerecord budget. A regression with a low-frequency guard exit should stay at zero rerecords; the existing dominant and 50/50 cases should continue to behave as documented.

src/core/trace_jit.rs
rlanday: Fixed in a2dcfbd. The adaptation window now counts every native trace call across Rust entries, including successful predictions; a partial final call is derived from the loop state rather than adding a counter operation to every native invocation. Guarded exits increment only the numerator. After the one permitted rerecord, the new trace clears its adaptive-accounting flag so it stays off this path.

I added `rare_non_self_loop_guard_exit_is_not_rerecorded`, a deterministic four-op non-self-loop trace with an approximately 1% guard mismatch. At 50,000 guest instructions, the previous code fails because `recording_attempts` reaches 2; the corrected code has one recording attempt, zero adaptive rerecords, four compiled ops, and more than 64 guarded exits. The existing dominant-path and alternating-path cases continue to pass.

Validation: default library tests (20/20), all-feature library tests (27/27), all-feature lib/bench Clippy with `-D warnings`, and rustfmt all pass. The existing `trace-branch-bias` benchmark remained within measurement noise (five-run median 296.0 M instr/s before vs 290.7 M instr/s after, -1.8%, with overlapping runs). A matched 30-second active Lemmings Level 1 sample at the same Systemless commit showed no application regression: runner-exclusive samples fell 342 to 164 and `execute_mem_op` samples fell 152 to 65. The latter is encouraging but I am treating it as workload-sensitive supporting evidence, not a claimed speedup from this correctness fix.

benletchford (approved):
The denominator fix addresses my requested change: successful trace calls now contribute across Rust entries, guarded exits remain the numerator, and adaptive accounting is disabled after the one rerecord. I reproduced the rare ~1% case at 50,000 instructions with zero rerecords; the dominant and 50/50 regressions also pass. Full rustfmt, locked all-target/all-feature Clippy with -D warnings, locked tests, and package verification pass locally, matching the successful GitHub checks.

PR #30, opened 2026-07-25
perf: JIT data-only MOVEM.W postincrement loops

rlanday:
## Summary

- add an atomic decoded fast path for data-register-only `MOVEM.W (An)+,<register list>`
- add matching native Cranelift and portable/WASM trace operations
- let Lemmings' hot palette-lookup loop compile through seven indexed byte moves and its closing `DBRA`
- add a permanent benchmark that reproduces the complete real loop, including the `MOVEM` instruction that prevented the prior synthetic benchmark from seeing the bottleneck

This is stacked on and depends on #29.

## Why this shape is narrow

The trace decoder accepts only word-sized postincrement MOVEM loads whose extension mask contains data registers only. Excluding address-register destinations avoids the architectural aliasing cases where the updated base register also appears in the register list.

The decoded, portable, and native paths all check the extension word, alignment, address-mask wrap, and complete contiguous source span before changing any 68k register. A failed check therefore returns to exact dispatch with no partial architectural state.

## Real workload evidence

The opportunity profile identified Lemmings PC `$00493010`, opcode `$4C98`, as a permanently rejected trace head. It is:

```text
MOVEM.W (A0)+,D1-D7
MOVE.B 0(A2,D1.W),(A1)+
MOVE.B 0(A2,D2.W),(A1)+
...
MOVE.B 0(A2,D7.W),(A1)+
DBRA D0,loop
```

The previous benchmark started after the `MOVEM`, so it measured traceable indexed moves but missed the instruction that kept the real loop out of the JIT. The new `lemmings-movem-loop` benchmark includes the complete nine-instruction inner loop and an outer reset path.

## Performance

Seven same-machine release runs on an Intel MacBook Pro, comparing unchanged #29 to this branch:

| Build | Runs (M guest instr/s) | Median |
|---|---|---:|
| #29 | 32.1, 32.5, 32.8, 31.2, 31.7, 31.1, 31.3 | 31.7 |
| This branch | 268.9, 275.5, 283.0, 271.3, 281.8, 280.7, 283.5 | 280.7 |

That is an **8.85x** median speedup for the complete loop.

In a real Lemmings trace-counter run, `$00493010` compiled as a nine-op trace and:

- retired 809,667 guest instructions
- used 1,764 native calls (459 instructions/call on average)
- had zero guard exits
- removed the formerly hot decoded `$12F2` sites at `$493014` through `$49302C` after warm-up

In a matched 30-second native macOS process sample, `execute_mem_op` top-of-stack samples fell from 52 on #29 to 34 on this branch. The broader application profile is noisy, so the PR does not infer a precise whole-app speedup from that single sample.

## Correctness and portability

Tests cover:

- extension-mask decode restrictions
- signed word loads into data registers
- preservation of condition codes
- PC and postincrement address updates
- atomic window misses and address-mask-wrap bails
- native result/cycle/retirement accounting

Validation performed:

- `cargo test --lib`
- full integration test suite
- `cargo check --target wasm32-unknown-unknown`
- `cargo check --target aarch64-apple-darwin`

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -295,6 +295,29 @@ fn blocked_self_loop(prefix_ops: usize, blocker: &[u16]) -> Vec<u16> {
     words
 }
 
+/// Isolate the largest rejected trace from the Lemmings profile: a hot
+/// region executed 24 traceable operations before `ASR.W #1,D7`, then seven
+/// more before `LSL.L #3,D0`. The surrounding ADDQs are synthetic; this
+/// benchmark measures the cost of rejecting versus compiling that topology,
+/// while the application profile measures its end-to-end importance.
+fn bench_immediate_shift_trace() {
+    let mut words = vec![0x5280; 24];
+    words.push(0xE247);
+    words.extend(std::iter::repeat_n(0x5280, 7));
+    words.push(0xE788);
+    let bytes_after_back_branch = (words.len() + 1) * 2;
+    let back_disp = -(bytes_after_back_branch as i16);
+    assert!((-128..=-1).contains(&back_disp));
+    words.push(0x6000 | (back_disp as u8 as u16));
+    bench_batch_loop_at(
+        "batch",
+        "shift blockers p24/32",
+        &words,
+        200_000_000,
+        0x7000,
+    );
+}
+
 /// Replay the three dominant rejected-trace shapes observed during a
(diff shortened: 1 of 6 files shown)


benletchford:
Superseded by #32, which carries the reusable MOVEM.W trace work on current `master` with a conventional commit and generic benchmark/comment naming. #32 passed the full CI and packaging gate and has been merged.

PR #50, opened 2026-08-01
feat: calibrate 020 taken-branch refill and model result forwarding from real hardware

LinuxJedi:
## Summary

Two 020 timing calibrations against a real 68EC020 (stock Amiga 1200 at 14.19 MHz) running an E-clock-referenced timing test disk ([Copperline's `timing-test/`](https://github.com/CopperlineHQ/Copperline/tree/main/timing-test)), whose loop rows isolate individual instruction costs.

### 1. `TAKEN_BRANCH_REFILL` drops from 2 to 1

Three independent `dbra`-loop rows measure a cached taken `dbra` at 7 clocks per iteration (the MC68020UM section-8.2 cache-case 6 plus 1), and the `move`, shift, `mulu`, paired-op and DIV rows all agree with the same refill once the loop branch is accounted for:

| loop row | real A1200 | refill = 1 | refill = 2 |
|---|---|---|---|
| dbra only (x3 rows) | 7 clocks | 7 | 8 |
| move.w Dn,Dn + dbra | 9 | 9 | 10 |
| lsl.l #8,Dn + dbra | 11 | 11 | 12 |
| mulu.w #imm,Dn + dbra | 36 | 36 | 37 |
| beam advance across a 64x DIVU loop under 6-bitplane display DMA | 888 cck | 886 | 906 |

The previous value of 2 was calibrated against FS-UAE (it reproduced FS-UAE's columns exactly); the same disk on the real machine shows FS-UAE over-bills every taken branch by one clock.

### 2. Result forwarding on register-to-register MOVE pairs

The disk's two register-pair rows discriminate a forwarding path: the RAW-dependent pair `move.w d2,d0 + move.w d0,d1` runs at **10 clocks** per loop iteration on real silicon against **11** for the independent pair `move.w d2,d0 + move.w d3,d1` - the dependent pair is *faster*, which only result forwarding can produce (the source is still in the execution unit's result latch, skipping the register-file read micro-cycle). Modelled as a one-clock refund on a register-to-register MOVE whose source is the register the immediately preceding register-to-register MOVE wrote, cached case only; wider forwarding is deliberately not modelled without hardware data. The latch register is CpuCore state (`serde(default)`), cleared by any other instruction and by exceptions.

With both changes patched into Copperline, every one of the disk's 22 confidently-read real-hardware rows lands within 1%, most within 1-2 E-clock ticks.

## Testing

- Full suite passes (SST fixtures + Musashi submodules); new unit test covers the forwarding refund, its RAW discrimination, latch clearing, and the uncached case.
- `cargo fmt --check` clean.
- End-to-end regression: TRSI Cubic Dream on A1200 (the demo whose crash introduced the refill; needs >33 frames of loader margin) runs clean past 140 s with both changes.

## Note for consumers

`CpuCore` gains a serialized field, so bincode-style snapshots of the core change shape with this release.

### DIFF
--- a/src/core/cpu.rs
+++ b/src/core/cpu.rs
@@ -84,6 +84,11 @@ pub struct CpuCore {
     pub dtt1: u32,
     /// Instruction Register (current opcode)
     pub ir: u32,
+    /// Data register left in the execution unit's result latch by the
+    /// previous instruction (a register-to-register MOVE), for the 68020
+    /// result-forwarding refund; see `timing_020.rs`.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub(crate) result_latch_020: Option<u8>,
 
     // ========== FPU Registers (68881/68882/68040) ==========
     /// FPU Data Registers (FP0-FP7) - 80-bit extended precision.
@@ -501,6 +506,7 @@ impl CpuCore {
             dtt0: 0,
             dtt1: 0,
             ir: 0,
+            result_latch_020: None,
             fpr: [FloatX80::default(); 8],
             fpiar: 0,
             fpsr: 0,
--- a/src/core/cpu.rs
+++ b/src/core/cpu.rs
@@ -607,6 +607,7 @@ impl CpuCore {
 
     /// Set CPU type and configure appropriate masks/timing.
     pub fn set_cpu_type(&mut self, cpu_type: CpuType) {
+        self.clear_execution_pipeline_state();
         self.cpu_type = cpu_type;
         self.is_pre_68020 = matches!(
             cpu_type,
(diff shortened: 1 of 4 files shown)


benletchford:
Thanks for the detailed real-hardware measurements and implementation. I merged this and released it in `m68k` 0.5.0.

I made one correctness fix before merging: the new `result_latch_020` could survive execution boundaries that occur outside `cycles_020()`. In particular, vectored exceptions/interrupts/trace, surfaced or HLE-handled traps, reset and CPU-model changes, STOP/HALT, and transitions through the fast batch execution path could leave a stale destination register in the latch. A later unrelated register-to-register `MOVE` could then incorrectly receive the one-clock forwarding refund.

I fixed this by centralising pipeline-state clearing and invoking it at each of those boundaries, with regression tests covering the cases. The hardware-calibrated taken-branch refill and the measured register-to-register `MOVE` forwarding behaviour were retained; the correction only prevents that forwarding state leaking beyond the immediately adjacent instruction window.

The final fix was included in the squash merge as 0b9117a. Thanks again — the hardware evidence was especially useful.

PR #57, opened 2026-08-02
perf: JIT guarded indexed scans

rlanday:
## Summary

Compile a common compiler-shaped guarded record scan as one native trace, with matching portable execution for non-JIT targets.

The change adds decoded, portable, and native trace support for the instruction/effective-address forms needed by the measured region:

- register-source `MULU.W` and `MULS.W`;
- immediate data-register ALU operations (`ORI`, `ANDI`, `SUBI`, `ADDI`, `EORI`, and `CMPI`);
- absolute-short/absolute-long `MOVE` and `MOVEA` sources;
- `(An)` and `d16(An)` `LEA`;
- checked indexed-memory `TST`;
- the existing conditional branches and register operations that join them into a trace.

Source, tests, benchmark names, and diagnostics remain application-neutral.

## Motivation and source shape

Profiling EV Override found a hot bounded scan that the existing JIT rejected after its first instruction. Its 68k structure corresponds to code like:

```c
struct Record {
    uint8_t prefix[42];
    int16_t state;
    uint8_t suffix[12];
};

for (int16_t i = 0; i < 128; ++i) {
    struct Record *records = *record_table;
    if (records[(uint16_t)i].state > 0)
        uncommon_path();
}
```

The compiler realizes that loop approximately as:

```text
MOVEQ   #56,D0               ; sizeof(Record)
MULU.W  D3,D0                ; byte offset = (uint16_t)i * 56
MOVEA.L record_table,A0      ; load the current record array
LEA     42(A0),A0            ; offsetof(Record, state)
TST.W   0(A0,D0.L)           ; records[i].state
BLE     latch                ; predicted common path
... uncommon path ...
latch:
ADDQ.W  #1,D3
CMPI.W  #128,D3
BLT     scan
```

Supporting only one of these missing forms does not admit the region: the trace recorder simply stops at the next unsupported instruction. They are therefore submitted together around one measured, permanent benchmark.

This pattern did not dominate the earlier Lemmings profiles. Its hot loops used simpler addressing/arithmetic combinations that the trace machinery already admitted; EV Override exposed this particular multiply + absolute pointer + indexed predicate combination.

## Correctness

- Immediate and absolute extension words are captured when a trace is recorded.
- Indexed `TST` uses the checked fast-memory window and exits before committing flags if the access is invalid.
- Portable and native paths preserve atomic fallback: an invalid guarded memory operation leaves the architectural state for that operation unchanged so the interpreter can execute it.
- `MULU.W` and `MULS.W` update the 32-bit destination and condition codes like the interpreter.
- MC68000 variable multiply timing is retained; later supported CPUs use their fixed timing model. Native trace headroom uses a conservative maximum for the MC68000 case.
- Address masking, alignment checks, self-modifying-code checks, and trace cycle/instruction accounting remain in force.

The benchmark repeatedly reaches the terminal bound and restarts, rather than replacing the bounded scan with an unrealistic infinite loop.

## Microbenchmarks

Both binaries were built separately from plain m68k 0.5.2 (`55f1a2d`) and this commit, with isolated Cargo target directories. SHA-256 hashes were checked to ensure that Cargo had not reused one artifact for both trees. Five runs were alternated; medians are shown.

### Native JIT (`--features jit`)

| Build | Runs (M guest instr/s) | Median |
|---|---|---:|
| m68k 0.5.2 | 54.4, 52.8, 54.1, 53.6, 52.2 | 53.6 |
| this change | 530.4, 527.5, 538.2, 529.5, 527.0 | 529.5 |

**9.88×** median throughput.

### Portable path (no `jit` feature)

| Build | Runs (M guest instr/s) | Median |
|---|---|---:|
| m68k 0.5.2 | 56.2, 56.0, 56.2, 57.7, 55.9 | 56.2 |
| this change | 122.1, 124.7, 125.1, 122.1, 119.6 | 122.1 |

**2.17×** median throughput. This is also the execution path compiled for WebAssembly.

## Whole-application profile

The application comparison used Systemless commit `65883b1` (the head of benletchford/systemless#277) for both binaries. Only the m68k dependency changed: plain 0.5.2 versus this standalone commit. The same stationary EV Override save was sampled three consecutive times for 15 seconds per build on an Intel Mac.

| Top-of-stack samples | 0.5.2 runs (median) | Candidate runs (median) | Change |
|---|---|---|---:|
| emulator runner | 594, 501, 592 (592) | 631, 471, 563 (563) | **−4.9%** |
| instruction dispatch | 46, 38, 38 (38) | 32, 23, 32 (32) | **−15.8%** |
| generic memory-op dispatch | 353, 291, 319 (319) | 318, 241, 336 (318) | −0.3% |

The isolated loop improvement is necessarily larger than the whole-app result because rendering, audio, other guest regions, and host event work remain unchanged.

## Trace attribution and CMPA ablation

An opportunity-profile ablation isolated this commit on top of a CMPA-only build:

- before this change, trace head `00EEFFC0` accumulated 986,765 rejected backward hits and stopped after a one-operation prefix at `MULU.W`;
- after this change, the same head compiled as a 9-operation trace and retired 12,684,006 guest instructions in 12,841 native calls (987.77 guest instructions per native call).

CMPA is intentionally not included here:

- the successfully compiled guarded scan contains no CMPA;
- CMPA-only produced no reproducible whole-application improvement in the matched scene;
- a different loop containing CMPA remained uncompiled even with both experiments: its blocker moved from `ADDI` after 14 admitted operations to memory `ADDQ.L #2,(A7)` after 26.

That experiment remains local until it demonstrates an independent application-level benefit or completes another measured region.

## Testing

- `cargo test --lib --all-features` — 142 passed
- isolated native and portable microbenchmarks above
- paired Systemless whole-application profile above

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -747,6 +747,74 @@ fn bench_trace_branch_bias() {
     );
 }
 
+/// Measure a compiler-shaped bounded record scan with a predicted interior
+/// branch and a conditional loop latch. An outer two-instruction loop resets
+/// the index so the benchmark repeatedly exercises both the hot path and the
+/// terminal exit instead of turning the scan into an unbounded synthetic loop.
+///
+/// The measured 68k sequence is equivalent to this C-shaped loop:
+///
+/// ```c
+/// struct Record {
+///     uint8_t prefix[42];
+///     int16_t state;
+///     uint8_t suffix[12];
+/// };
+///
+/// for (int16_t i = 0; i < 128; ++i) {
+///     struct Record *records = *record_table;
+///     if (records[(uint16_t)i].state > 0)
+///         uncommon_path();
+/// }
+/// ```
+fn bench_guarded_indexed_scan() {
+    const CODE_BASE: u32 = 0x6000;
+    const INSTRS: u32 = 100_000_000;
+    let words = [
+        0x7600, // outer: MOVEQ #0,D3
+        0x7038, // scan: MOVEQ #56,D0
+        0xC0C3, // MULU.W D3,D0
+        0x2079, 0x0000, 0x4000, // MOVEA.L $00004000,A0
+        0x41E8, 0x002A, // LEA 42(A0),A0
+        0x4A70, 0x0800, // TST.W 0(A0,D0.L)
+        0x6F02, // BLE.S latch (common path)
+        0x4E71, // uncommon-path placeholder
+        0x5243, // latch: ADDQ.W #1,D3
+        0x0C43, 0x0080, // CMPI.W #128,D3
+        0x6D00, 0xFFE2, // BLT.W scan
+        0x7600, // MOVEQ #0,D3
+        0x60DC, // BRA.S scan
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    bus.write_long(0x4000, 0x5000);
+
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0xF000);
+        cpu
+    };
+
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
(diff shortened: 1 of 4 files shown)


benletchford (approved):
Reviewed the decoded, portable, and native paths, including extension capture, address masking/alignment, atomic fast-memory bailout, self-modifying-code guards, flag behavior, multiply timing, and the focused benchmark. GitHub CI is green and the full all-feature test suite also passes locally. I found no blocking correctness or design issues.

PR #59, opened 2026-08-03
perf: JIT packed lookup generator loops

rlanday:
## Summary

This extends trace execution across a compiler-generated fixed-point lookup-table generator that currently falls back to the interpreter at several instruction forms:

- `CMPA.W/L` from `(An)` and `d16(An)` memory sources
- `CMPA.W/L #imm,An`
- `ADDQ`/`SUBQ` read-modify-write operations through `(An)` and `d16(An)`
- the 68020+ 32-bit-result forms of `MULU.L Dn,Dn` and `MULS.L Dn,Dn`

The native path uses checked fast memory access and retains the existing atomic fallback contract. Writes that may overlap traced code bail before modifying guest state. The portable path implements the same operations for non-native targets.

This branch is rebased onto current `master`, which already contains the guarded indexed-scan work from #57.

## Why these operations belong together

EV Override exposed a hot helper that writes 1,024 packed 16-bit results to an output table. Supporting only one instruction does not unlock the loop: its trace crosses a memory-source address comparison, stack-based output cursor update, immediate address comparison, and signed long multiply. Each unsupported operation becomes the next trace boundary.

The observed code is approximately:

```c
struct Record {
    uint8_t  unknown00[8];
    uint8_t *base;          // +0x08
    uint32_t velocity;      // +0x10, fixed point
    uint8_t  unknown14[4];
    uint32_t position;      // +0x18, fixed point
    uint32_t unknown1c;
    uint8_t *current;       // +0x20
    uint8_t *previous;      // +0x24
    uint8_t  unknown28[4];
    int16_t  x_scale;       // +0x2c
    int16_t  y_scale;       // +0x2e
};                         // 50 bytes

for each record {
    int16_t magnitude = *(int16_t *)record->current;
    if (record->current == record->previous) {
        magnitude += *(int16_t *)(record->current + 2);
        magnitude >>= 1;
    }

    record->previous = record->current;
    x_accum += ((int32_t)record->x_scale * magnitude) >> 7;
    y_accum += ((int32_t)record->y_scale * magnitude) >> 7;
    record->position += (int32_t)record->velocity >> 8;
    record->current = record->base
        + 2 * ((int32_t)record->position >> 8);
}

uint8_t x = clamp(x_accum, -127, 127) + 128;
uint8_t y = clamp(y_accum, -127, 127) + 128;
uint16_t packed = ((uint16_t)x << 8) | y;
```

The helper emits each packed word using the saved return address as an inline output cursor:

```asm
MOVEA.L (A7),A0       ; load the saved return address
ADDQ.L  #2,(A7)       ; make RTS return after this output word
MOVE.W  D2,(A0)       ; write the packed result inline
```

That explains why the memory `ADDQ` is not incidental stack traffic: it participates in every generated output word. The loop terminates with an immediate `CMPA` against 1,024, and `MULS.L` performs the fixed-point products.

Lemmings did not expose this combination. Its measured hot loops predominantly used conventional 68000-era integer and indexed-memory patterns. This helper uses 68020 long multiply plus an unusual return-address output cursor, so the older application did not reach these trace blockers.

## Correctness boundaries

- `CMPA.W` sign-extends its source before the 32-bit address comparison.
- `CMPA` updates `N`, `Z`, `V`, and `C` but preserves `X` and does not write its address-register destination.
- long multiply is admitted only on 68020 and later CPU configurations.
- only the 32-bit-result `MULU.L`/`MULS.L` forms are admitted; the `Dh:Dl` 64-bit-result form remains in the interpreter.
- signed and unsigned overflow are computed independently and covered by differential tests.
- memory read-modify-write operations calculate the effective address once and bail atomically on failed access or possible code overlap.
- cycle accounting matches the existing interpreter values.

## Measurements

Host: Intel MacBookPro16,1. Application samples are matched 15-second stationary-gameplay samples. Percentages below normalize leaf sample counts by each sample's main-thread sample total, so differing sample totals do not distort the comparison.

| Build | Exact baseline | 68k runner | `read_word` | generic memory execution | instruction dispatch |
|---|---|---:|---:|---:|---:|
| A | #57 plus memory-source `CMPA` | 976 / 10,619 = 9.19% | 7.26% | 4.89% | 0.63% |
| B | A plus indirect memory `ADDQ` and immediate `CMPA` | 752 / 11,287 = 6.66% | 5.49% | 3.53% | 0.43% |
| C | B plus 68020 long multiply (this full branch) | 526 / 10,854 = 4.85% | 3.74% | 2.21% | 0.29% |

From A to C, normalized samples fell by:

- 47.3% in the 68k runner
- 48.5% in `read_word`
- 54.8% in generic memory execution
- 54.7% in instruction dispatch

The memory-source `CMPA` commit was not timed independently in a stable application phase. Trace-opportunity instrumentation identified it as the first required boundary; the ablation above deliberately uses it in every row and measures the later additions on top. This avoids attributing phase noise to that first change.

The committed focused benchmark measures the displacement address comparison itself:

```asm
loop:
    CMPA.L 16(A1),A3
    BRA.S  loop
```

Alternating warm runs measured 41.4–43.8 M instructions/s before admission and 259.6–358.0 M instructions/s after admission (roughly 6–8x). This benchmark demonstrates that specific boundary; it is not presented as a model of the entire application loop.

An earlier synthetic mixed-loop benchmark was removed. It needed a benchmark-only `SUBQ` to keep the inline output cursor in bounds, changed the trace shape, and did not reproduce the application improvement. The PR therefore uses the paired application samples above as its end-to-end evidence rather than retaining a misleading benchmark.

## Tests

- `cargo test --lib --all-features --quiet` (148 passed)
- native-versus-portable differential tests for address comparisons, memory `ADDQ`, and signed/unsigned long multiply
- CPU-generation rejection, wide-result rejection, flag preservation, sign extension, atomic memory-failure, and code-overlap tests

The generic crate source, test names, and benchmark names intentionally do not mention individual downstream applications; the application is named here only to make the real workload and measurements reviewable.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -815,6 +815,49 @@ fn bench_guarded_indexed_scan() {
     );
 }
 
+/// Measure a compiler-shaped address bound check whose longword limit lives
+/// in a stack-frame-style displacement slot. The unconditional latch keeps
+/// the CMPA memory form at the hot trace boundary without adding unrelated
+/// address-register mutation to the measurement.
+fn bench_displacement_address_compare() {
+    const CODE_BASE: u32 = 0x6C00;
+    const INSTRS: u32 = 100_000_000;
+    let words = [
+        0xB7E9, 0x0010, // CMPA.L 16(A1),A3
+        0x60FA, // BRA.S loop
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    bus.write_long(0x4010, 0x5000);
+
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(1, 0x4000);
+        cpu.set_a(3, 0x5000);
+        cpu.set_a(7, 0xF000);
+        cpu
+    };
+
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    println!(
+        "batch     displacement CMPA      {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 /// Measure decoded generic memory operations without allowing a backward
 /// branch to turn the workload into a native JIT loop. Each pass walks the
 /// same straight-line code, retaining the decoded-op cache while resetting
@@ -926,6 +969,10 @@ fn main() {
     }
     if only.as_deref() == Some("guarded-indexed-scan") {
         bench_guarded_indexed_scan();
+        return;
+    }
+    if only.as_deref() == Some("displacement-address-compare") {
+        bench_displacement_address_compare();
         return;
     }
     if only.as_deref() == Some("immediate-shifts") {
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -10,7 +10,7 @@ use super::cpu::{CpuCore, NFLAG_SET};
 use super::execute::RUN_MODE_BERR_AERR_RESET;
 use super::mem_ops::{BitSource, DecodedMemOp, FastEa};
 use super::memory::AddressBus;
-use super::op_cache::{BinaryOp, BitOp, CachedRunResult, DecodedSimpleOp, is_pre_68020};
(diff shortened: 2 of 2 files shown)


benletchford:
I merged #57 with rebase to preserve linear history, but GitHub rewrote its SHA, so this stacked branch is now reported as conflicting even though `master..HEAD` contains only the intended two-commit delta. Please rebase onto current `master` and rerun CI; once that is green this is ready to squash-merge.

benletchford (approved):
Reviewed the net two-commit delta after #57: CMPA memory/immediate forms, address-relative ADDQ/SUBQ read-modify-write, and 68020+ 32-bit-result long multiply preserve the trace bailout, flag, self-modifying-code, and CPU-generation contracts. GitHub CI is green and the full all-feature suite passes locally. I found no blocking code or design issue. The branch still needs a clean rebase onto current master because #57 landed via rebase with a rewritten SHA.

rlanday:
Dependency/test-stack note: #66 repairs the trace-recorder handoff for instructions that execute through the complete dispatcher even when the trace decoder already supports their exact form. The opcode implementations in this PR remain independently reviewable, but the end-to-end application benefit for full-dispatch-classified forms depends on that handoff. A single top-of-stack branch containing #59, #66, and the subsequent measured coverage changes is available for testing at https://github.com/rlanday/m68k-rs/tree/perf/ev-override-integration-stack. This is a one-way integration stack, not a circular PR dependency.

PR #60, opened 2026-08-03
feat: add exact-width memory bit-field bus accesses

LinuxJedi:
## Summary

The memory forms of the 68020 bit-field instructions (BFTST/BFEXTU/BFEXTS/BFCHG/BFCLR/BFSET/BFFFO/BFINS) had two defects that surface on cycle-billing hosts:

1. **Access pattern**: `bf_extract_mem_window_msb0` always read a five-byte window, so a one-bit `BFSET (An){Dn:1}` issued five byte reads plus one write where a real 68020 performs a single read-modify-write.

2. **68020 timing**: `timing_020`'s `group_e` billed the memory forms with MC68020UM section 8.2.14 totals that *include* the operand read and write cycles, while the host separately bills every bus access the executor performs - double-charging the bus portion. The register-form rows are the UM cache-case values verbatim (they perform no accesses), so the memory rows now follow the same internal-only convention.

Both are fixed, and the timing is calibrated against a real A1200 rather than against the manual alone.

## What the hardware measured

A stock A1200 ran a 14-row bit-field probe (Copperline `timing-test/bfprobe.asm`: 8192-iteration loops referenced to the CIA E-clock, with the anchor rows reproducing the established rdprobe/fwdprobe columns exactly). It is the first hardware measurement of the 020 bit-field class we know of.

- A field within four bytes is **one operand cycle** whatever it spans: spans of one, two, three and four bytes all measure 28.10 clocks for a `BFSET (An){0:1}` + `DBcc` loop. A five-byte span costs one operand cycle more (44.32).
- BFTST 24.12, BFEXTU and BFINS 28.10, register-form BFSET 20.02, the dynamic offset free, and the DBcc alignment clock absorbed by the operand access.

## What it did not measure, and what the model does about it

The review is right that this does not establish a long-width transfer, and checking it properly rather than arguing the point is what the data supports: the A1200 moves any span up to four bytes across its **32-bit** chip bus in a single cycle, and every probe row's field starts at a longword-aligned address. A byte, a word and a long are therefore indistinguishable on that machine. Separating them would need a 16-bit chip port (an OCS/ECS host with an 020) or a bus-signal capture.

So the executor transfers **exactly the width the field spans** - byte, word, three bytes, or long, plus one byte for a five-byte span. No byte outside the field is read or written, which removes the adjacent-MMIO exposure and the moved fault boundary of the earlier commit.

The timing rows are unaffected by the width choice, for the same reason the measurement cannot see it.

## The three-byte transfer

A three-byte span used to compose a word and a byte, which reads the right bytes but bills a cycle-counting host two operand transactions where the hardware charges one. `AddressBus` now has the transfer:

- `read_three_bytes`/`write_three_bytes` and their fallible variants, all **default methods** composing a word and a byte, so no existing bus needs to change.
- The fallible defaults compose the *fallible* word and byte accesses rather than delegating to the infallible ones, so a bus that reports faults keeps reporting them on whichever half fails. A host that bills bus cycles overrides both variants; the trait documents that and the regression bus does it.
- `CpuCore::read_24`/`write_24` mirror `read_32`/`write_32`, including the MMU page-straddle split and a fault size of 3. The 68020/68030 SSW reports that as SIZ 11, its three-byte encoding; the 68040 SZ and 68060 FSLW have no three-byte value (11 is a line transfer) and still report long.

With a host implementing the hook, the probe's three-byte row moves from `2D11` to `2CD1` - the same value as the one-, two- and four-byte spans and the dynamic form, which is exactly what the A1200 measures for all of them.

Note that this makes the PR additive to the public API. It is titled `fix:` because it fixes a defect; retitle it before merging if you would rather release-please cut a minor bump.

## Verification

- `tests/bitfield_memory_access_tests.rs` pins the **set of byte addresses touched**, not just the access count, for every form at every span from one to five bytes, with sentinel bytes either side of each field. The long-width shape fails them. The three-byte read and read-modify-write assert a single size-3 transaction each way, and a separate test pins the composed fallback for a bus that predates the hook.
- Timing regressions in the `timing_020` test module cover the register, within-four-bytes and five-byte rows, cached and uncached, plus an assertion that a five-byte span costs exactly one operand cycle more than a within-four-bytes span across every form.
- Full crate suite green (62 test targets), `cargo clippy --all-targets` and `cargo fmt --check` clean, Musashi `test_m68020_bitfield` unchanged.
- Against the real A1200 column a cycle-billing host lands every row within 0.62% (worst is the five-byte span), with the anchors and the register form tick-exact.
- SANITY Roots II AGA (CopperlineHQ/Copperline#371, the demo that motivated this): 60 frames across the DIE dissolve are **byte-for-byte identical** with and without the three-byte hook, and identical to the earlier long-width shape, so the real-world fix survives every step of the safer model.

### Known remaining gaps

BFFFO and the five-byte rows of the non-BFSET forms keep the MC68020UM relative deltas and are not yet hardware-measured.

### DIFF
--- a/src/core/cpu.rs
+++ b/src/core/cpu.rs
@@ -1696,6 +1696,72 @@ impl CpuCore {
         }
     }
 
+    /// Read a three-byte operand from memory (data space), returned in the
+    /// low 24 bits.
+    ///
+    /// The 68020/68030 transfer an operand at the size it spans - byte, word,
+    /// three-byte or long (MC68020UM 5.3.1). Only the memory bit-field
+    /// instructions produce a three-byte operand, so only the 68020 and later
+    /// reach this path; the 68000/68010 have no three-byte transfer.
+    #[inline]
+    pub fn read_24<B: AddressBus>(&mut self, bus: &mut B, addr: u32) -> u32 {
+        if self.faulted() {
+            return 0;
+        }
+        // Part E.2: report internal clocks elapsed before this bus access.
+        self.flush_sync(bus);
+        let mut addr = self.address(addr);
+        if self.has_pmmu && self.pmmu_enabled {
+            let pm = self.mmu_page_mask();
+            if addr & pm > pm - 2 {
+                // Three-byte operand straddling an MMU page: a word cycle and
+                // a byte cycle, each translated on its own page (read_16
+                // sub-splits further if its half straddles the boundary).
+                let hi = self.read_16(bus, addr) as u32;
+                if self.faulted() {
+                    return 0;
+                }
+                let lo = self.read_8(bus, addr.wrapping_add(2)) as u32;
+                return (hi << 8) | lo;
+            }
+        }
+        if let Some((a, v)) = self.mmu_read_override
+            && a == addr
+        {
+            // RTE'd 030 bus-fault frame with DF cleared: the handler
+            // supplied this read's result in the data input buffer.
+            self.mmu_read_override = None;
+            return v & 0x00FF_FFFF;
+        }
+        {
+            if self.has_pmmu && self.pmmu_enabled {
+                match crate::mmu::translate_address(
+                    self,
+                    bus,
+                    addr,
+                    /*write=*/ false,
+                    self.is_supervisor(),
+                    /*instruction=*/ false,
+                ) {
+                    Ok(p) => addr = self.address(p),
+                    Err(f) => {
+                        self.handle_mmu_fault(bus, f, false, false, 3);
+                        return 0;
+                    }
+                }
+            }
+        }
+        match bus.try_read_three_bytes(addr) {
+            Ok(v) => v & 0x00FF_FFFF,
+            Err(f) => {
+                if matches!(f.kind, BusFaultKind::BusError) {
+                    self.trigger_bus_error(bus, addr, false, false, 3);
+                }
+                0
+            }
+        }
+    }
+
     /// Write byte to memory (data space).
     #[inline]
     pub fn write_8<B: AddressBus>(&mut self, bus: &mut B, addr: u32, value: u8) {
@@ -1795,6 +1861,64 @@ impl CpuCore {
             && matches!(f.kind, BusFaultKind::BusError)
         {
             self.trigger_bus_error(bus, addr, true, false, 2);
+        }
+    }
+
+    /// Write the low 24 bits of `value` to memory (data space) as a
+    /// three-byte operand.
+    ///
+    /// See [`CpuCore::read_24`]: the memory bit-field instructions are the
+    /// only producers of a three-byte operand.
+    #[inline]
+    pub fn write_24<B: AddressBus>(&mut self, bus: &mut B, addr: u32, value: u32) {
+        if self.faulted() {
+            return;
+        }
+        // Part E.2: report internal clocks elapsed before this bus access.
+        self.flush_sync(bus);
+        let mut addr = self.address(addr);
+        if self.has_pmmu && self.pmmu_enabled {
+            let pm = self.mmu_page_mask();
+            if addr & pm > pm - 2 {
+                // Three-byte operand straddling an MMU page: a word cycle and
+                // a byte cycle, each translated (and fault-suppressed) on its
+                // own page.
+                self.write_16(bus, addr, (value >> 8) as u16);
+                if self.faulted() {
+                    return;
+                }
+                self.write_8(bus, addr.wrapping_add(2), value as u8);
+                return;
+            }
+        }
+        if self.mmu_write_suppress == Some(addr) {
+            // See write_8: DF-cleared write fault, already completed.
+            self.mmu_write_suppress = None;
+            return;
+        }
+        self.pending_fault_wdata = value & 0x00FF_FFFF;
+        {
+            if self.has_pmmu && self.pmmu_enabled {
+                match crate::mmu::translate_address(
+                    self,
+                    bus,
+                    addr,
+                    /*write=*/ true,
+                    self.is_supervisor(),
+                    /*instruction=*/ false,
+                ) {
+                    Ok(p) => addr = self.address(p),
+                    Err(f) => {
+                        self.handle_mmu_fault(bus, f, true, false, 3);
+                        return;
+                    }
+                }
+            }
+        }
+        if let Err(f) = bus.try_write_three_bytes(addr, value & 0x00FF_FFFF)
+            && matches!(f.kind, BusFaultKind::BusError)
+        {
+            self.trigger_bus_error(bus, addr, true, false, 3);
         }
     }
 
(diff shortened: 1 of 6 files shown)


LinuxJedi:
Update: a real A1200 has now run the bit-field probe disk (14 rows; anchors reproduce the established calibration columns exactly), and the second commit recalibrates against that column - the first hardware measurement of the 020 bit-field class we know of.

Two changes over the initial version:

1. **Operand model settled by hardware**: field spans of 1, 2 and 4 bytes cost exactly the same on the silicon, so the executor now performs ONE long operand access at the field's first byte (plus a byte access for five-byte spans) - the MC68020UM 8.2.14 note's model - instead of byte-granular accesses. The long RMW writes back the untouched bytes it read, as the silicon does.
2. **Timing rows fitted to the measured column** (BFSET mem loop 28.10 clk uniform across spans 1-4, five-byte 44.32, BFTST 24.12, BFEXTU/BFINS 28.10, register BFSET 20.02, dynamic offset free, DBcc alignment clock absorbed by the operand access). A cycle-billing host now lands every probe row within 0.6% of the real machine, register-form and anchor rows tick-exact.

BFFFO and the five-byte rows of the non-BFSET forms keep the MC68020UM relative deltas and are marked not yet hardware-measured. The probe disk, raw columns and methodology live in Copperline's `timing-test/bfprobe.asm` (CopperlineHQ/Copperline#373).

benletchford (changes requested):
Thanks for the hardware work and the concrete downstream reproduction. The current second commit should not land yet because it turns every 1-4-byte field into `read_32`/`write_32` at the field's first byte. That can expose adjacent MMIO side effects and change fault boundaries by touching up to three bytes outside a small field.

The timing evidence establishes that fields below five bytes take one operand cycle, but it does not establish that the transfer is always long-width. MC68020UM section 8.2.14 describes operand-cycle counts; section 5.3.1 separately documents byte, word, three-byte, and long transfer sizes. Please model the actual spanned transfer width without making unselected adjacent bytes observable (including the three-byte case), or provide direct bus-signal evidence and an AddressBus contract that safely represents the claimed behavior.

Please also update the title/body, which still describe the superseded minimal-byte implementation, and add direct timing regression assertions for the changed register, <5-byte, and five-byte rows (cached and uncached). The access-shape tests are useful, but currently lock in the disputed long-width inference rather than the measured timing result.

LinuxJedi:
You were right, and I spent a while trying to show otherwise before accepting it.

My reasoning for the long-width model was that the probe writes to chip RAM, so if the 020 only transferred the spanned width I'd expect a 1-byte field and a 4-byte field to differ by a bus cycle - and they don't, spans 1, 2 and 4 all land on the same 28.10 clocks. That argument only works on a 16-bit port. The A1200's chip RAM is 32 bits wide, so a byte, a word and a long are one cycle each, and on top of that every field in the probe starts at a longword-aligned address. The measurement genuinely cannot see the difference. I implemented both models and ran the disk under the emulator rather than keep arguing it: identical column on 13 of the 14 rows.

So the executor now transfers exactly what the field spans - byte, word, word+byte, or long, plus a byte for a five-byte span - and never touches a neighbour. The timing rows are unaffected by that, for the same reason the probe can't see it, so the calibration and the access shape turn out to be independent questions.

The three-byte case is the one loose end. AddressBus has no three-byte transfer, so I compose a word and a byte: the right bytes and only those, but one access more than the silicon for a host that bills them. Worth saying this part is measured rather than assumed - probe row 10 is `bfextu (a0){4:16}`, whose field reaches into the third byte, so it *is* the three-byte measurement (I'd mislabelled it as two), and the A1200 puts it on the same 28.10 clocks as every other span. The over-bill works out at 0.28% on that row. If you'd like a default-composed `read_24`/`write_24` on the trait to close it I'll add it, but I didn't want to grow the public trait without asking.

Tests reworked along the lines you asked for. The access tests now pin the set of byte addresses touched, with sentinels either side of every field, instead of counting accesses by granularity - the long-width shape fails them. `timing_020` gets direct regressions for the register, within-four-bytes and five-byte rows in both cache and worst case, plus an assertion that a five-byte span costs exactly one operand cycle more across every form.

Squashed to a single commit, and rewrote the title and body since both still described the first implementation.

One more datapoint on whether the safer shape costs anything: 60 frames across the Roots II dissolve come out byte-for-byte identical between this and the long-width version, so the demo fix survives it intact.

benletchford (changes requested):
The revised byte/word/long access shapes and adjacent-MMIO regressions are sound, and the full required suite passes on a local rebase onto current `master`. The three-byte case remains blocking, though: `bf_extract_mem_window_msb0` emits `read_16` + `read_8`, and modify forms similarly emit two writes. A cycle-billing `AddressBus` therefore observes two operand transactions per direction even though the measured aligned three-byte row takes the same single operand cycle as the other sub-five-byte spans; the timing row cannot undo that extra bus charge. Please add default-compatible 24-bit read/write bus hooks (including fallible/CPU helpers and MMU-boundary handling as needed), use one logical size-3 access here, and update the regression to assert a single size-3 read/write while retaining the touched-byte and fault-boundary guarantees.

LinuxJedi:
Fair enough - that one's on me for offering the trait pair and then leaving the composition in. You're right that no timing row can undo an access the host has already been charged for.

`AddressBus` now has the three-byte transfer, as `read_three_bytes`/`write_three_bytes` plus fallible variants, all default methods composing a word and a byte so nothing existing has to change. One wrinkle worth flagging since it breaks the pattern of the other six pairs: the fallible defaults compose the *fallible* word and byte rather than delegating down to the infallible variant. Delegating would have quietly dropped fault reporting for three-byte spans on every bus that reports faults today. The cost is that a host wanting single-transaction billing has to override both variants - documented on the trait, and the regression bus does exactly that so the pattern is visible somewhere real. Say the word if you'd rather have the uniform delegation and take the fault loss.

`CpuCore::read_24`/`write_24` mirror `read_32`/`write_32`, page-straddle split included, with fault size 3. That turned out to have a proper home: the 68020/030 SSW SIZ field already encodes three-byte as 11. The 040 SZ and 060 FSLW don't (11 is a line transfer there), so those still report long, with a comment saying why rather than leaving it looking like an oversight.

The number that convinced me you were right rather than just reasonable: with Copperline implementing the hook, the probe's three-byte row goes 2D11 -> 2CD1, which is the same value as the one-, two- and four-byte spans and the dynamic form. The A1200 measures all of them at 28.10 clk, so the row stops being an outlier in either direction instead of just swapping which side it errs on. The tests now assert one size-3 transaction each way, and there's a separate test standing in for a bus written before the hook existed, checking the composed path still touches those three bytes and no others.

62 test targets green, clippy and fmt clean, Musashi bitfield fixtures unchanged. Roots II is 60 frames byte-identical with and without the hook, and identical to the long-width shape too, so the demo fix has now survived all three access models.

One process note: this makes the PR additive to the public API, so if you'd rather release-please cut a minor bump than a patch, retitle before merging. I left it as `fix:` since the thing it fixes is a bug.

benletchford (approved):
The latest revision resolves the requested three-byte access contract: existing `AddressBus` implementations remain functional through fault-preserving defaults, timing-aware buses can override one logical size-3 transaction, CPU/MMU paths retain page-boundary and fault-size handling, and bit-field regressions cover one- through five-byte spans without adjacent accesses. The complete required suite passes on the current `master` integration; package verification also passes for that combined tree. No correctness or design blocker remains.

PR #66, opened 2026-08-04
perf: continue trace recording through full dispatch

rlanday:
## Summary

The decoded opcode cache uses `CachedOp::Complex` to mean “execute this instruction through the complete dispatcher.” The trace recorder previously treated that dispatch classification as a hard trace boundary before the instruction ran.

Those are different questions. Some instructions intentionally use the complete dispatcher while the trace decoder and native/portable trace executors already know how to replay their exact form. Stopping at the cache miss made that existing trace support unreachable from real execution paths.

This change:

- lets the complete dispatcher execute the instruction once;
- offers a successfully executed instruction to the trace decoder afterward;
- leaves the trace decoder responsible for accepting or rejecting the exact opcode and extension form; and
- still discards a partial recording before surfaced traps, address/bus-error handler handoffs, or other host-controlled execution boundaries.

The implementation is deliberately general: it does not special-case an opcode or an application.

## Execution model

Before:

```text
decoded cache sees Complex
    -> discard recording
    -> execute through full dispatcher
```

After:

```text
decoded cache sees Complex
    -> execute through full dispatcher
    -> if execution stays in the guest instruction stream:
           ask the existing trace decoder to reconstruct the executed form
           accept it only if that decoder already proves it replayable
       else:
           discard the partial recording
```

This does not make the complete dispatcher itself part of native trace execution. It only repairs the recording handoff so already-supported instructions can be included in a later compiled trace.

## Why this is safe

- An unsupported or unreadable form still makes `decode_trace_op` finish/reject the recording exactly as before.
- Surfaced `BatchExit` values stop recording before control returns to the embedding runtime, which may emulate a trap and resume at an unrelated guest PC.
- Address and bus exceptions stop recording before the CPU resumes at the exception vector.
- The change does not weaken fast-memory checks, self-modifying-code guards, CPU-model checks, or the atomic fallback behavior of individual trace operations.

Regression tests cover both sides of the handoff:

- a `MOVEM.W (A0)+,D1` loop, which uses the complete dispatcher but already has trace support, now compiles and retires natively;
- a surfaced A-line trap terminates an in-progress recording instead of allowing it to cross the host boundary.

## Performance role and measurements

This is an enabling change, not a standalone opcode optimization. A focused compiler-shaped loop confirmed that the repaired handoff produces the expected 10-operation native trace, but alternating throughput runs were neutral/noisy:

- baseline: 47.3–52.5 M instructions/s
- candidate: 46.3–58.3 M instructions/s

That result is expected: once compiled, the benchmark’s inner work was already cheap, so changing how the trace is first recorded does not materially change its steady-state native code.

The practical benefit appears when a hot loop contains an operation that the trace executor can lower but the decoded cache sends through the complete dispatcher. Without this handoff, later opcode-coverage PRs can contain correct native implementations and focused benchmarks yet remain unreachable in the downstream application. The full dependent stack used for EV Override is available here for reviewer testing:

https://github.com/rlanday/m68k-rs/tree/perf/ev-override-integration-stack

That branch contains this handoff together with #59 and the subsequent measured trace-coverage changes. The individual PRs remain separate review units; the integration branch is only a convenient non-circular test target.

## Tests

- `cargo test --features trace-profile --lib` (134 passed)

`cargo test --all-features` additionally requires the optional Musashi fixture submodule, whose binary fixtures are not present in this worktree. The fixture-independent library test suite above is clean.

### DIFF
--- a/src/core/execute.rs
+++ b/src/core/execute.rs
@@ -488,6 +488,10 @@ impl CpuCore {
             self.cycles_remaining = i32::MAX / 2;
 
             if retired >= max_instructions {
+                // A full-dispatch instruction may have just extended a path
+                // recording before exhausting this outer batch's budget.
+                // The caller may mutate guest state before the next batch.
+                trace_jit::stop_recording(self);
                 return BatchResult {
                     instructions: retired,
                     exit: BatchExit::BudgetExhausted,
@@ -571,6 +575,10 @@ impl CpuCore {
                 }
             };
             if let Some(exit) = exit {
+                // The caller may emulate a surfaced trap and resume at an
+                // unrelated guest PC. Never let an in-progress path recording
+                // cross that host-controlled execution boundary.
+                trace_jit::stop_recording(self);
                 return BatchResult {
                     instructions: retired,
                     exit,
@@ -582,9 +590,21 @@ impl CpuCore {
             // frame and jumped to the handler; skip trace/interrupt checks
             // for the faulting instruction (mirrors `execute`).
             if self.run_mode == RUN_MODE_BERR_AERR_RESET {
+                // The fault handler is not the sequential continuation of the
+                // instruction being recorded. Discard the partial path before
+                // execution resumes at the exception vector.
+                trace_jit::stop_recording(self);
                 self.run_mode = RUN_MODE_NORMAL;
                 probe_on_entry = true;
             } else {
+                // A complex opcode leaves the decoded fast path and executes
+                // through the full dispatcher above. Offer that successfully
+                // executed instruction to an in-progress trace just as the
+                // decoded paths do. The trace decoder remains the authority
+                // on whether the exact opcode/extension form is safe to
+                // replay; an unsupported operation simply ends recording.
+                trace_jit::record_executed(self, bus, self.ppc, self.pc);
+
                 // Mirrors `execute`: only backward branches can reach a
                 // trace head, so straight-line dispatches re-enter the
                 // fast loop without a trace-cache probe.
@@ -601,6 +621,7 @@ impl CpuCore {
             }
 
             if self.stopped != 0 {
+                trace_jit::stop_recording(self);
                 return BatchResult {
                     instructions: retired,
                     exit: BatchExit::Stopped,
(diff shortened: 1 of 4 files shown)


benletchford (changes requested):
The new full-dispatch handoff can leave a recording live across a `run_batch` return. After `record_executed`, the outer loop returns at the next budget check when the just-retired complex instruction exhausts `max_instructions`, and the watched-PC return below likewise does not call `stop_recording`. The decoded `Budget`/`Watched` paths already stop the recorder, so this new path can carry `trace_recording` across a host boundary; if the embedder changes PC or guest state before resuming, the next batch can append an unrelated path to the stale recording. Please stop recording before every outer batch exit that can occur after this handoff (at least budget exhaustion and watched PC), and add focused tests that warm/start a recording, retire a trace-decodable `CachedOp::Complex`, then return on each boundary and assert the recording is no longer live.

rlanday:
Fixed in e7ab102. The review was correct: after a trace-decodable `CachedOp::Complex` retired through the full dispatcher, the outer `run_batch` path could return for budget exhaustion or a watched PC without clearing the partial recording.

The updated commit now calls `stop_recording` before those outer exits (and before the adjacent stopped-CPU exit for the same invariant). I added focused regressions that first warm the loop until the next full-dispatch instruction starts a recording, then make that instruction end the batch through each affected boundary:

- one instruction of remaining budget → `BudgetExhausted`
- the post-instruction PC in `watch_pcs` → `WatchedPc`

Both tests assert the complex instruction retired, the expected boundary was returned, and `cpu.trace_recording` is false. Validation on the amended branch: 136 trace-profile library tests passed, Clippy passed with `-D warnings`, and `wasm32-unknown-unknown --no-default-features` passed. I am now rebuilding and remeasuring the dependent integration stack on top of this amended commit.

benletchford (approved):
The amended head closes the recorder on budget, watched-PC, and stopped exits after full dispatch, and the two focused regressions reproduce the affected budget/watch paths after warming a trace-decodable complex instruction. I also validated the amended change on top of #59, #60, and #65: formatting, Clippy, default/all-feature tests, docs, and package verification all pass. The full-dispatch handoff remains gated by the existing trace decoder and preserves fault/trap boundaries. No blocker remains.

PR #73, opened 2026-08-05
feat: report exact trace recording paths

rlanday:
## Summary

Make the optional trace opportunity profile report the exact dynamic paths that compiled and the exact paths that failed to record.

The existing head row is a process-lifetime aggregate. A trace head can first fail on one path and later compile on another, but the row can retain the earlier blocker label beside the later compiled-call counts. During EV Override profiling this made an earlier `4C01` failure appear to be the boundary of a successfully compiled trace whose executed path did not contain that instruction.

This PR keeps the aggregate table for ranking, labels its lifetime-aggregation limitation, and adds shape tables that answer the diagnostic question unambiguously.

## What is recorded

For every distinct failed recording shape:

- trace start PC and CPU model;
- the exact reconstructed prefix operations, in execution order — two
  equal-length paths that converge on the same blocker remain distinct
  rows, each printed with its own `prefix:` line;
- blocker PC;
- executed instruction-register opcode;
- the opcode and up to two following guest words as they appear in memory
  — labeled as following words, not extensions, because an extensionless
  blocker is followed by other opcodes.

All failure diagnostics are read through the attached fastmem window
only; the collection helper does not take an `AddressBus`, so enabling
profiling cannot add host-visible bus transactions (MMIO, watchpoints,
fault counters). When no window covers the blocker, the diagnostics are
recorded as unavailable rather than read from the bus. The distinct
failed-shape map is capped (4096 shapes) with a reported overflow count,
keeping the profiler's memory bounded for the process lifetime now that
each entry carries a prefix sequence.

Failures are classified as:

- opcode read failed;
- executed opcode no longer matches memory; or
- unsupported form / operand-extension read failed.

For every distinct compiled shape, the report records the exact dynamic sequence of PC, opcode, and captured extension words plus its compile count.

This separates path variation from self-modifying code and from genuinely unsupported instruction forms. It also prevents us from optimizing an opcode merely because an older aggregate blocker label happened to remain visible.

## Scope and overhead

All new collection is behind the existing `trace-profile` feature. Normal library/JIT builds do not allocate the shape maps or perform the extra diagnostic memory reads. This is instrumentation only; it does not change guest execution or claim a runtime speedup.

## Validation

- `cargo test --all-features` — full suite, 62 test binaries (166 lib
  tests), with both fixture submodules initialized, on v0.7.2
- `cargo check --target wasm32-unknown-unknown --no-default-features` and
  `cargo check --features trace-profile`
- bus-access regression driving the recording-failure path directly with a
  word-read-counting bus: exactly one decoder opcode read and nothing
  else, asserted both with and without a window attached
- test proving two distinct equal-length prefixes at the same blocker stay
  separate rows with separate counts
- failed-shape cap and overflow-count coverage
- report test proving that a failed 20-op path and a compiled 2-op path at the same head remain distinct
- coverage for the exact following words and explanatory aggregate-table note

The profiler applies directly to current master (v0.7.2). It produced the
before/after recording-path evidence in #72 and #74.

### DIFF
--- a/src/core/mem_ops.rs
+++ b/src/core/mem_ops.rs
@@ -1323,6 +1323,17 @@ fn store(cpu: &mut CpuCore, win: Win, loc: &Loc, size: Size, value: u32) {
 /// Returns `false` (with zero state changes) when the op must fall back to
 /// full dispatch: window misses, pre-68020 misaligned accesses, 68020+
 /// full-format extensions, or no window at all.
+/// Side-effect-free diagnostic read of a guest word through the attached
+/// fastmem window. Returns `None` when no window is attached or the address
+/// is outside it. The `AddressBus` is never involved, so callers cannot add
+/// host-visible bus transactions.
+#[cfg(feature = "trace-profile")]
+pub(crate) fn peek_window_word(cpu: &CpuCore, addr: u32) -> Option<u16> {
+    let win = Win::from_cpu(cpu)?;
+    let off = win.off(cpu.address(addr), 2)?;
+    Some(win.read(off, Size::Word) as u16)
+}
+
 pub(crate) fn execute_mem_op(cpu: &mut CpuCore, op: DecodedMemOp) -> bool {
     let Some(win) = Win::from_cpu(cpu) else {
         return false;
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -960,7 +960,16 @@ impl TraceJit {
         let cpu_type = recording.cpu_type;
         let adaptive_rerecords = recording.adaptive_rerecords;
         #[cfg(feature = "trace-profile")]
-        let recorded_ops = recording.ops.len();
+        let recorded_shape = recording
+            .ops
+            .iter()
+            .map(|op| super::trace_profile::TraceShapeOp {
+                pc: op.pc,
+                opcode: op.opcode,
+                extension: op.extension,
+                extension2: op.extension2,
+            })
+            .collect();
         self.slots[idx] =
             match self.compile_decoded_ops(cpu, start_pc, cpu_type, recording.ops, Some(exit_pc)) {
                 Some(mut trace) => {
@@ -980,7 +989,7 @@ impl TraceJit {
             };
         #[cfg(feature = "trace-profile")]
         if matches!(self.slots[idx], TraceSlot::Compiled(_)) {
-            super::trace_profile::note_compiled(start_pc, cpu_type, recorded_ops);
+            super::trace_profile::note_compiled(start_pc, cpu_type, recorded_shape);
         }
     }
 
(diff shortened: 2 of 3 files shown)


benletchford (changes requested):
Two correctness/design issues block this as an exact-path profiler:

- On every decode failure, `record_executed` performs three additional `AddressBus::try_read_word` transactions. Those calls may be host-visible (MMIO, watchpoints, counters, fault injection), so enabling `trace-profile` can change bus behavior; for an extensionless unsupported opcode, the `extension` fields also contain subsequent opcodes rather than operand words. Please collect diagnostics without new unconstrained bus transactions (or restrict reads to a side-effect-free source) and add a bus-access regression proving profiling does not add observable reads.

- `FailedShapeKey` records only `prefix_ops`, not the prefix `TraceShapeOp` sequence. Two equal-length paths from the same head that converge on the same blocker and words are therefore merged, so the report still does not identify exact failed dynamic paths. Please include the recorded prefix in the failed-shape identity/output and cover two distinct equal-length prefixes that share a blocker.

The new public snapshot/row structs should also be made deliberately extensible (for example with `#[non_exhaustive]`) or the source-compatibility impact documented, because adding public fields to `TraceProfileSnapshot` breaks exhaustive construction/patterns under the existing feature.

rlanday:
All three points addressed in the updated commit (rebased onto v0.7.2):

**Bus transactions.** Failure diagnostics now read only through the attached fastmem window via a new `peek_window_word` helper — the helper does not take an `AddressBus`, so the absence of bus transactions is a type-level property rather than a convention. When no window covers the blocker, diagnostics are recorded as unavailable instead of being read from the bus, and the `OpcodeReadFailed` docs now state that meaning. The fields and report columns are renamed `ext1/ext2` → `next1/next2` since, as you noted, an extensionless blocker is followed by opcodes, not operands. A regression test drives the recording-failure path directly with a word-read-counting bus and asserts exactly one read (the decoder's own opcode fetch) and nothing else, both headless and with a window attached.

**Exact failed paths.** The failed-shape identity now includes the exact reconstructed prefix sequence, the public row exposes it, and the report prints a `prefix:` line per failed shape mirroring the compiled `path:` lines. A test covers two distinct equal-length prefixes converging on the same blocker and words: two rows, separate counts.

**API extensibility.** `#[non_exhaustive]` on all seven public structs.

One addition from self-review: because prefix sequences make each failed-shape entry meaningfully sized and the map lives for the process, distinct failed shapes are now capped at 4096 with an overflow counter surfaced in the snapshot and stated in the report, with coverage.

benletchford (approved):
The updated head keeps diagnostics on the side-effect-free fastmem window, records exact failed prefixes, bounds failure-shape storage, and makes the new profiling structs extensible. The focused regressions and full local/CI validation pass.

PR #74, opened 2026-08-05
perf: JIT displacement PEA pushes

rlanday:
## Summary

Admit `PEA (d16,An)` as a trace operation: the first predecrement store in
the trace JIT. The pushed value is the effective address computed from the
pre-decrement register state (correct even for `PEA (d16,A7)`), no
condition code changes, and the decremented stack slot passes the
alignment, window, and store-overlap checks before the store and the A7
update, so a failed check bails to full dispatch with nothing committed.

This follows the merged #68/#69/#72 and applies directly to master at
v0.7.2.

The source remains generic and contains no application-specific matching.

## Generated-code shape

`PEA (d16,A5)` is how classic Mac compilers pass an A5-relative global by
reference:

```c
DoSomething(&globals->field);   /* PEA $40(A5); ...; JSR/BSR */
```

The focused benchmark models the push with the popping load standing in
for the callee consuming the pointer:

```c
for (;;) {
    ++count;                        /* ADDQ.L #1,D0    */
    int32_t *arg = &globals->field; /* PEA $40(A5)     */
    sink = arg;                     /* MOVE.L (A7)+,D3 */
    ++iterations;                   /* ADDQ.L #1,D1    */
}
```

## Why this operation

With indexed CMPI.W admitted (#72), `PEA (d16,An)` is the largest
remaining trace-recording blocker in the stationary EV Override profile
(663.8 million guest instructions, `trace-profile` build with the
exact-path reporter): `486D`/`486E` blocked six trace heads — 903,738
rejected recordings and 16.6 million projected trace operations — with the
two hottest heads carrying 21-operation recorded prefixes, the longest
blocked prefixes in the profile.

Admitting the push lets recording continue past it; how much of that
projection is realized depends on what follows the push in each loop and
is measured with the same reporter below.

In the matching session with this change (482.0 million guest
instructions), `486D`/`486E` block nothing, and session-wide rejected
recordings fell from 93.8% to 87.0% of backward hits. The realized
breakdown is honest rather than triumphant: one of the six formerly
blocked heads compiled — a thirteen-operation trace retiring 2.0 million
guest instructions over 247,655 native calls, averaging only 8.1
operations per call — while the other five extended their recorded
prefixes past the push and now stop at the next operation in the same
argument-setup sequence, `CLR.L -(SP)` and `MOVE.W #imm,-(SP)`, which
this change promotes to the profile's top remaining blockers (7.5 and
1.0 million projected operations). The push family is a chain: each
admitted member moves the boundary toward the terminating `BSR.L`. This
PR contributes the predecrement-store machinery the rest of that chain
reuses.

## Correctness and safety

- Decode only the measured `(d16,An)` control form; all other PEA shapes
  keep their existing dispatch.
- Compute the pushed address from the pre-decrement register state, which
  keeps `PEA (d16,A7)` pushing the original stack pointer's address.
- Capture the displacement extension word for trace validation.
- Check alignment, window bounds, and store overlap with the trace's own
  code before the store; the store and the A7 update are the only state
  changes and both happen after every check.
- Change no condition codes.
- Account sixteen cycles: the MC68000's twelve-cycle push plus the
  four-cycle displacement extension fetch.
- Provide both portable-trace and native-JIT execution; the native tests
  cover portable equivalence, the `(d16,A7)` self-reference, an
  out-of-window bail, and a store-overlap bail, each retiring nothing.

## Performance

Focused benchmark: `pea-displacement`, a stack-balanced push/pop loop.
The instruction budget is one initializer plus exactly twenty million
five-instruction iterations, so the run ends at the loop head and the
benchmark asserts the final counters, the popped pointer value, and the
balanced stack.

Five alternating runs against a baseline worktree at v0.7.2 master that
carries an identical copy of the benchmark but none of this
implementation, M guest instructions/s:

| Run | v0.7.2 base | this PR |
|---:|---:|---:|
| 1 | 71.8 | 521.9 |
| 2 | 68.3 | 499.0 |
| 3 | 71.8 | 546.7 |
| 4 | 73.3 | 575.2 |
| 5 | 76.4 | 542.7 |
| **Median** | **71.8** | **542.7** |

That is a **7.56x** focused-loop speedup over the actual PR base.

Whole-application `sample` ladder (two consecutive 30-second 1 ms
captures per session; a four-session block in the same stationary scene:
baseline, this PR, a follow-up candidate, then the baseline again as a
drift control): this PR measured **21.96%** of main-thread observations
in the emulation-runner subtree versus **26.60%** for the baseline
sampled before it and **26.66%** for the identical baseline sampled at
the block's end — a **17.5% reduction**, against a measured drift
envelope of 0.06 points between the two baseline sessions. The
trap-dispatch share moved proportionally (the work mix was constant
across all four sessions), and an earlier independent pair reproduced
the same direction and magnitude. One stated caveat: the runner metric
is CPU share, not CPU per guest instruction, so in an elastic scene a
throughput gain can partially mask itself; on that reading this figure
is conservative.

## Validation

- `cargo test --all-features` and `cargo test` — full suite, 62 test
  binaries each, with both fixture submodules initialized
- `cargo test --all-features --lib pea_displacement` — decode, portable
  push/flag/PC behavior, native-versus-portable parity including
  `(d16,A7)`, atomic out-of-window and store-overlap bails
- `cargo check --no-default-features` / `--features jit` /
  `--features trace-profile`

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1038,6 +1038,72 @@ fn bench_indexed_immediate_compare() {
     );
 }
 
+/// Measure a JIT-compiled loop that pushes an address with `PEA (d16,A5)`
+/// and pops it back with a postincrement load, keeping the stack balanced
+/// across iterations while exercising the checked predecrement store. This
+/// models a compiler materializing the address of an A5-relative global as
+/// a by-reference argument; the popping load stands in for the callee
+/// consuming the pushed pointer:
+///
+/// ```c
+/// int32_t count = 0;                  /* MOVEQ #0,D0, once  */
+/// for (;;) {
+///     ++count;                        /* ADDQ.L #1,D0    */
+///     int32_t *arg = &globals->field; /* PEA $40(A5)     */
+///     sink = arg;                     /* MOVE.L (A7)+,D3 */
+///     ++iterations;                   /* ADDQ.L #1,D1    */
+/// }
+/// ```
+///
+/// The instruction budget is one initializer plus a whole number of
+/// five-instruction iterations, so the run ends at the loop head and the
+/// final counter, pointer, and stack values are exact.
+fn bench_pea_displacement_loop() {
+    const CODE_BASE: u32 = 0x6E00;
+    const ITERATIONS: u32 = 20_000_000;
+    const INSTRS: u32 = 1 + ITERATIONS * 5;
+    let words = [
+        0x7000, // MOVEQ #0,D0
+        0x5280, // loop: ADDQ.L #1,D0
+        0x486D, 0x0040, // PEA $40(A5)
+        0x261F, // MOVE.L (A7)+,D3
+        0x5281, // ADDQ.L #1,D1
+        0x60F4, // BRA.S loop
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(5, 0x4200);
+        cpu.set_a(7, 0x8000);
+        cpu
+    };
+
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.d(0), ITERATIONS, "count increments once per iteration");
+    assert_eq!(cpu.d(1), ITERATIONS, "iterations increment once per pass");
+    assert_eq!(cpu.d(3), 0x4240, "the popped value is the pushed address");
+    assert_eq!(cpu.a(7), 0x8000, "the stack is balanced at the loop head");
+    println!(
+        "batch     pea displacement loop   {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 /// Measure decoded generic memory operations without allowing a backward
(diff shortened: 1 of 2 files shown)


benletchford (changes requested):
The implementation and atomic bailout ordering look sound, but the performance benchmark does not execute the loop described in the PR. The `BRA.S` at `0x6E0C` branches from `0x6E0E` by `-14` (`0x60F2`) to `0x6E00`, so `MOVEQ #0,D0` runs every iteration and `count` is reset before the increment. The stated loop body begins at the `ADDQ` at `0x6E02`, which requires displacement `-12` (`0x60F4`). This also changes the compiled trace from six operations to five, so it can affect the native-call amortization behind the reported 8.57x result. Please make the guest loop match the stated workload (and preferably assert the final counter/stack state), then rerun the alternating base/candidate measurements with the corrected benchmark.

rlanday:
Thanks — confirmed on all points. The `BRA.S` displacement targeted the `MOVEQ`, so the counter reset every iteration and the compiled trace was six operations rather than the five described.

Fixed in the updated commit: the branch is now `0x60F4` (loop head at the `ADDQ`), the instruction budget is one initializer plus exactly 20,000,000 five-instruction iterations so the run ends at the loop head, and the benchmark asserts the final counters, the popped pointer value, and the balanced stack.

Also rebased onto v0.7.2 (post-#72 master) and reran the five alternating pairs against a v0.7.2 baseline worktree carrying the identical corrected benchmark: median 71.8 → 542.7 M instr/s, i.e. **7.56x** rather than the previously reported 8.57x — the five-operation trace amortizes slightly less per native call, as you anticipated. PR body updated to match.

rlanday:
Added drift-controlled whole-application data: a four-session sampling ladder in one scene block (baseline → this PR → a follow-up candidate → baseline again) measured this PR at 21.96% emulation-runner CPU share versus 26.60%/26.66% for the baseline sampled before and after — a 17.5% reduction against a 0.06-point drift envelope, with the work mix constant across sessions. The PR body's performance section is updated with the details.

benletchford (changes requested):
The corrected PEA implementation and standalone full check matrix pass, but this head now conflicts with current `master` in `src/core/trace_jit.rs` after #73 merged. Please rebase onto `master`, resolve the overlapping trace-dispatch/profiling additions, and rerun the checks so the combined tree—not only the v0.7.2-based head—is validated.

rlanday:
Rebased onto v0.8.0 (post-#73 master); the merged profiler's regression test and this PR's tests coexist at the shared module, full gates rerun green (62 suites, clippy --all-targets, wasm).

benletchford (approved):

PR #76, opened 2026-08-05
perf: JIT immediate predecrement stack stores

rlanday:
## Summary

Admit the measured immediate predecrement stack stores as trace
operations: `CLR.L -(An)`, `CLR.W -(An)`, and `MOVE.W #imm,-(An)`. Both store a
recording-time constant through the checked window at the decremented
address; the alignment, window, and store-overlap checks all precede the
store and the register update, so a failed check bails to full dispatch
with nothing committed. Because the stored value is a constant, N/Z/V/C
are compile-time constants of the value (X preserved), and each form's
MC68000 cost (22 cycles for the read-modify-write long clear, fourteen
for the word clear, twelve for the word immediate push) is captured per
decode.

This follows the merged #68/#69/#72/#73 and applies directly to master
at v0.8.0. It is independent of #74, but the two changes are links in the
same chain: compilers push call arguments with PEA, `CLR.L -(SP)`, and
`MOVE.W #imm,-(SP)` in sequence.

The source remains generic and contains no application-specific matching.

## Generated-code shape

This is the argument-push sequence before a call. The profiler recorded
the form directly: a blocked shape at `00EF96C0` shows `3F3C` with
following words `001B 61FF` — that is

```asm
MOVE.W  #27,-(SP)      ; push a selector argument
BSR.L   ...            ; call
```

or in C terms:

```c
DoThing(0L, kSelector);   /* CLR.L -(SP); MOVE.W #27,-(SP); BSR.L */
```

Admitting the pushes extends recording through the argument setup; the
chain still terminates at the call itself, so like #74 this PR's value is
partly realized traces and partly recorded prefixes carried closer to the
`BSR.L` boundary — both measured by the recording profiler.

## Why these operations

In stationary EV Override profiling with #72 merged and the #74 candidate
admitted (482.0 million guest instructions, `trace-profile` build),
`CLR.L -(SP)` is the largest remaining recording blocker — two trace
heads, 365,276 rejected recordings, 7.54 million projected trace
operations — and `MOVE.W #imm,-(SP)` blocks five more heads for a further
998,212 projected operations. Both rose in rank when #74 unblocked the
PEA pushes ahead of them in the same sequences, and the session with
this change's first two forms admitted surfaced the word clear next:
`CLR.W -(SP)` blocked 84,222 recordings for 1.26 million further
projected operations, so it is included.

## Correctness and safety

- Decode only the measured forms: long and word `CLR` and word
  immediate `MOVE` through `-(An)`; all other sizes and destinations keep their existing
  dispatch.
- Store through the existing checked fast-memory window; check alignment,
  window bounds, and overlap with the trace's own code before the store
  and the register update.
- Set N/Z from the stored constant and clear V/C, exactly as MOVE and CLR
  both do, preserving X.
- The MC68000 CLR quirk of reading the destination before writing is not
  observable through the plain-memory window and its cost is in the
  recorded cycle count; the emission stores directly.
- Account per-form cycles captured at decode time: 22 for
  `CLR.L -(An)` (twelve-cycle read-modify-write clear plus the ten-cycle
  predecrement long EA), twelve for `MOVE.W #imm,-(An)`.
- Provide both portable-trace and native-JIT execution; the native test
  covers portable equivalence with both stores in one trace, and atomic
  out-of-window and store-overlap bails.

## Performance

Focused benchmark: `predec-imm-stores`, a stack-balanced loop pushing
all three constants and popping them back, with the final counter, sink,
and stack values asserted at an exact eight-instruction iteration
boundary.

Five alternating runs against a baseline worktree at v0.8.0 master that
carries an identical copy of the benchmark but none of this
implementation, M guest instructions/s:

| Run | v0.8.0 base | this PR |
|---:|---:|---:|
| 1 | 38.7 | 254.5 |
| 2 | 38.9 | 217.4 |
| 3 | 36.1 | 252.7 |
| 4 | 34.4 | 244.8 |
| 5 | 38.1 | 238.0 |
| **Median** | **38.1** | **244.8** |

That is a **6.42x** focused-loop speedup over the actual PR base.

No whole-application claim is made; the sampling protocol's limits are
recorded with the #72 measurements, and the application-level evidence
for this family is the recording profile above.

## Validation

- `cargo test --all-features` — full suite with both fixture submodules
  initialized
- `cargo test --all-features --lib pre_dec_imm` — decode of both forms,
  portable push/flag semantics validated against the interpreter (Z for
  the cleared long, N for a negative immediate, X preserved), native
  parity with both stores in one compiled trace, and atomic
  out-of-window and store-overlap bails

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1104,6 +1104,75 @@ fn bench_pea_displacement_loop() {
     );
 }
 
+/// Measure a JIT-compiled loop that pushes two constants with the admitted
+/// predecrement stores and pops them back, keeping the stack balanced while
+/// exercising both checked constant stores. This models the argument-push
+/// sequence compilers emit before a call, with the pops standing in for the
+/// callee consuming its arguments:
+///
+/// ```c
+/// for (;;) {
+///     PushZeroLong();      /* CLR.L -(SP)         */
+///     PushZeroWord();      /* CLR.W -(SP)         */
+///     PushWord(0x1234);    /* MOVE.W #$1234,-(SP) */
+///     sink16 = PopWord();  /* MOVE.W (A7)+,D3     */
+///     sink16b = PopWord(); /* MOVE.W (A7)+,D4     */
+///     sink32 = PopLong();  /* MOVE.L (A7)+,D5     */
+///     ++iterations;        /* ADDQ.L #1,D1        */
+/// }
+/// ```
+///
+/// The instruction budget is a whole number of eight-instruction
+/// iterations, so the run ends at the loop head and the final counter,
+/// sink, and stack values are exact.
+fn bench_predec_imm_store_loop() {
+    const CODE_BASE: u32 = 0x6E00;
+    const ITERATIONS: u32 = 12_500_000;
+    const INSTRS: u32 = ITERATIONS * 8;
+    let words = [
+        0x42A7, // loop: CLR.L -(SP)
+        0x4267, // CLR.W -(SP)
+        0x3F3C, 0x1234, // MOVE.W #$1234,-(SP)
+        0x361F, // MOVE.W (A7)+,D3
+        0x381F, // MOVE.W (A7)+,D4
+        0x2A1F, // MOVE.L (A7)+,D5
+        0x5281, // ADDQ.L #1,D1
+        0x60EE, // BRA.S loop
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0x8000);
+        cpu
+    };
+
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.d(1), ITERATIONS, "one increment per iteration");
+    assert_eq!(cpu.d(3), 0x1234, "the popped word is the pushed immediate");
+    assert_eq!(cpu.d(4), 0, "the popped word is the cleared word slot");
+    assert_eq!(cpu.d(5), 0, "the popped long is the cleared long slot");
+    assert_eq!(cpu.a(7), 0x8000, "the stack is balanced at the loop head");
+    println!(
+        "batch     predec imm stores       {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 /// Measure decoded generic memory operations without allowing a backward
 /// branch to turn the workload into a native JIT loop. Each pass walks the
 /// same straight-line code, retaining the decoded-op cache while resetting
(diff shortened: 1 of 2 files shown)


rlanday:
Status: staying in draft — whole-application measurement shows this change **hurts** end-to-end performance despite the 6.42x focused-loop speedup, and I’m investigating the fix.

What I can state with evidence so far: in a capped (`--cpu-mhz 25`) same-scene measurement ladder, the build with these stores consumed ~19.2M guest instructions per second against the 25M budget — roughly 3–4x the 4–7M/s the same scene consumes on the baseline and on the #74 build — with emulation CPU share rising back to baseline levels, and the growth concentrated in a few very large self-loop traces (the largest grew from 2.9% to 14.7% of all session instructions).

The working theory, partially verified in code: some of the loops these opcodes unlock for tracing are wait/poll loops. The runner deliberately recognizes such loops — its `_TickCount` spin-wait templates literally anchor on `0x42A7` (`CLR.L -(SP)`) — and fast-forwards or parks them instead of executing them; compiling them into native traces converts skipped time into burned instructions. The pre-change profiler ranking also counted those spin iterations as projected trace operations, which is exactly how they became an optimization target.

Still being pinned down with deterministic headless comparisons: which detector the burning loops escape (the trap-anchored templates are JIT-immune by construction, so suspicion falls on trap-less pollers and the idle-cycle prover interaction). The likely fix direction is making the wait detectors take precedence over trace admission — possibly with the trace layer classifying architecturally-idempotent poll loops and handing them to the runner's parking machinery instead of compiling them. This PR leaves draft only after that interaction is resolved and the whole-application measurement is clean.

rlanday:
Investigation update, concluding for now: the burn is confirmed but **episodic** — a matched capped profiler pair showed the previously-dominant spin trace nearly identical across builds while a different trace cluster reshuffled 5x, i.e. which loops burn depends on scene/audio state, not deterministically on this PR's ops. Root causes identified in the runner: (1) trap-less poll loops have no wait detector at all (the spin templates and the idle-cycle prover are both trap-anchored), so they always burn wall time, and JIT-compiling them multiplies instructions burned per second (measured ~18 CPU-ns/instruction of pure native spinning); (2) the profiler's projected-dispatch ranking counts those spin iterations as optimization opportunity, which is how wait idioms became optimization targets.

Planned fix direction (RFC to follow rather than more measurement): classify architecturally-idempotent poll loops at trace-recording time — the recorder is the only component that sees a complete loop body — refuse to compile them as spins, report them as waits excluded from opportunity rankings, and hand them to the runner's existing advance/park machinery via a wait-typed trace exit. This PR stays draft until that lands and whole-app measurement is meaningful for it.

rlanday:
The investigation's conclusions and the proposed fix are now written up as RFC #79 (wait-aware trace recording); a Stage-1 PR (classify pure poll loops at compile time, refuse to compile them as spins, report them distinctly, exclude them from opportunity rankings) will reference it. This PR's re-evaluation is gated on that landing.

rlanday:
Closing this — the whole-application evidence came back neutral at best, and it was opened on focused-benchmark evidence alone.

**The measurement** (deterministic headless ladder, 300M guest instructions, v0.8.1 versus this branch, both with `trace-profile`; identical final-framebuffer SHA-256, so byte-identical guest work):

- **Nothing newly compiles.** JIT totals are byte-identical to base — `native_calls` 47,671, `jit_retired` 11,146,241, `rejected_hits` 615,734. Zero execution change.
- Recording does advance, which confirms the op works: `00FB3C14` now records through `3F3C`/`2F0A` (17 ops, up from 15) and blocks at `4EBA` (`JSR (d16,PC)`) instead. It simply never reaches a compilable region.
- The `00FB1418` head (1,536 hits, previously blocked at `3F3C`) disappears from the report entirely — the unattributed-rejection pattern filed as #84.
- Wall time 206.2 s versus a 204.3 s base median, inside a ±19% host-noise band.

**Why neutral means close here rather than wait.** The forms this admits — `CLR.L -(SP)`, `MOVE.W #imm,-(SP)` — are the Toolbox call preamble: push a result slot, push arguments, then execute an A-trap. Traps are hard trace boundaries, so these sequences characteristically record *toward* something uncompilable, which is what both measured workloads show. That is also how this PR's top-ranked target came to be a `42A7`-headed loop — the exact byte pattern the consuming runner's own spin-wait template anchors on — and why admitting it previously coincided with *increased* application CPU (the 19.2M-of-25M spin episode written up in #79). There is no workload on file where these ops gate a real compilation.

So: no positive evidence anywhere, negative evidence in the interactive scene, and a structural reason to expect that. Not worth carrying.

Worth keeping from it: the investigation it triggered produced #79, #81, #82, and #84. Happy to revisit if #84 lands and a profile then shows these forms actually gating compilable regions.

Issue #71, opened 2026-08-05
Add a cycle-aware fast execution path with boundary events


uchidah:
Thank you for the recent precise cycle-runner improvements through m68k 0.7.0.

We have now integrated `run_for_cycles_with_boundary_hook()` into our downstream machine emulator. The instruction and interrupt-entry boundary contracts provide the ordering we need, and our formal machine snapshots pass at short and long cycle limits.

However, a CPU-only equivalent-state benchmark shows that the current hook-capable cycle runner is not a throughput path. We would like to discuss whether the crate could eventually provide a cycle-aware fast execution path that retains the existing boundary semantics while using decoded execution, FastMem, or trace execution where exact accounting is available.

## Current status

The precise boundary runner is now our correctness reference.

Our downstream validation confirms that:

- `Instruction` and `InterruptEntry` events are reported in the required order;
- interrupt-entry cycles are applied before the first handler instruction;
- bus boundary requests and hook returns stop execution before the next fetch;
- formal snapshots match across requested limits of 57,372, 57,373, 75,000, 100,000, and 1,000,000 cycles;
- two independent long-distance runs match at 10,000,000 and 100,000,000 requested cycles, including CPU state, device state, RAM, and total cycles.

The remaining issue is performance architecture rather than boundary correctness.

## CPU-only equivalent-state benchmark

We added a release-only benchmark using a small NOP/BRA workload.

It compares:

1. a lightweight repeated-`step()` loop; and
2. `run_for_cycles_with_boundary_hook()` with an equivalently lightweight callback.

Both paths run for 10,000,000 cycles and must finish with:

- the same consumed-cycle total;
- the same complete CPU state;
- the same number of host callbacks.

This removes X68000 device timing, DMA, MFP, overlay, and machine-scheduler work from the comparison.

After one warm-up run, three measurements were:

| Runner | Run 1 | Run 2 | Run 3 | Median |
| --- | ---: | ---: | ---: | ---: |
| lightweight repeated `step()` | 67.095 ms | 69.642 ms | 85.173 ms | 69.642 ms |
| precise boundary runner | 77.857 ms | 71.883 ms | 94.759 ms | 77.857 ms |

The precise boundary runner was slower in every measured pair, with a median difference of approximately 11.8%.

[...]

PR #78, opened 2026-08-05
perf(execute): add precise decoded dispatch for boundary-hook runs

uchidah:
Addresses #71

## Summary

This is a bounded internal experiment for `run_for_cycles_with_boundary_hook()`.

It adds a narrow decoded dispatch for register-only, exact-timing operations while preserving the existing precise instruction-boundary contract. There are no public API changes, and this PR does not use or modify FastMem, portable traces, or native traces.

## Design

`step()` now delegates to a private `step_with_dispatch()` helper. The helper retains the existing precise path:

- `set_precise_bus(true)`
- stopped-state handling
- `begin_instruction_fetches()`
- `ppc`, rollback, and SR snapshots
- precise opcode fetch and prefetch behavior
- fault handling
- cache/fetch accounting
- instruction-end prefetch top-up
- trace and IRQ handling

The boundary-hook runner selects its dispatcher only after that precise opcode fetch has completed.

For an unsupported opcode, the fast dispatcher does not call `step()` again and does not fetch again. It passes the already fetched opcode to the existing `dispatch_instruction()` path before any fast-executor register, flag, or memory mutation.

The shared precise fetch prologue may update the same fetch, prefetch, PPC, and rollback state that `step()` updates; the fallback guarantee applies to mutations performed by the narrow decoded executor itself.

The initial allowlist is intentionally small:

| Operation | CPU models |
| --- | --- |
| NOP | M68000, M68010, M68020, M68030 |
| MOVEQ | M68000, M68010, M68020, M68030, M68040 |

M68040 NOP is excluded because the normal dispatcher includes model-specific T0 pipeline synchronization that this narrow executor does not reproduce.

The existing runner ordering is unchanged: after normal instruction retirement, the hook is invoked, any hook-side CPU or bus changes take effect, bus boundary requests are checked, and IRQ state is reevaluated before another ordinary instruction can execute. `InterruptEntry` remains separate from retired instruction accounting.

The narrow dispatcher is only selected during normal run mode when T0 and T1 tracing are inactive.

## Prefetch and Host-side PC Changes

This PR does not add automatic prefetch invalidation after a hook changes `cpu.pc`.

`CpuCore::invalidate_prefetch()` is the existing host contract for an externally forced PC or instruction-visible mapping change. The differential test explicitly calls it after changing PC, on both the repeated-step reference path and the boundary-hook path.

## Tests

Added coverage verifies:

- CPU state, cycles, and hook counts against the repeated-step runner for M68000 through M68040.
- Unsupported `ADDQ` falls through to the normal dispatcher without changing the M68000 word-read access order.
- Hook-side changes to PC, SR, and CPU type preserve equivalence when the existing prefetch invalidation contract is followed.
- A hook-raised IRQ is serviced before the next ordinary instruction for M68000, M68020, and M68040.
- The hook-raised IRQ differential compares consumed cycles, retired instructions, instruction-hook count, interrupt-entry event sequence, return reason, PC/SR/D/A registers, stack state, and guest memory.
- Existing hook return, bus boundary, IRQ promotion, interrupt-entry, and resume tests remain green.

Validation was run after rebasing onto current `master`:

```text
cargo fmt --all -- --check
cargo clippy --all-targets --all-features --locked -- -D warnings
cargo test --all-features --locked
cargo package --locked
```

All four commands completed successfully.

## Controlled Release Benchmark

The benchmark uses a CPU-only `AddressBus` with no FastMem. It keeps the two Issue #71 workloads unchanged.

All measurements used the same release test binary:

- binary: `m68k_issue71_benchmark-ba0d5d12d26b4b4b.exe`
- SHA-256: `A327ED60C0FEBC148086F5ABE45742458FC93FAF8CF4E45EAC6F104AA0CA4EDF`
- Rust: `rustc 1.93.0 (254b59607 2026-01-19)`, `x86_64-pc-windows-msvc`
- host: Windows 10 Pro build 19045, Intel Core i7-6700K

Each of two separate processes used:

- a 100,000,000-cycle budget
- three warm-up pairs per workload
- twenty measured step/boundary pairs per workload
- alternating runner order
- exact equality assertions for consumed cycles, retired instructions, hook count, the architectural CPU register/control state compared by the benchmark, and guest RAM

The process was pinned to logical CPU 0 through `SetProcessAffinityMask`. The Windows power plan was left unchanged at Balanced, and no frequency pinning was applied.

The paired ratio is `boundary / step`; values below 1.0 are faster.

| Workload | Run 1 median | Run 2 median | Combined median | 10% trimmed mean | IQR |
| --- | ---: | ---: | ---: | ---: | ---: |
| NOP / BRA | 1.018309 | 1.007964 | 1.009591 | 1.015726 | 0.075824 |
| MOVEQ / MOVEQ / BRA | 0.785594 | 0.777737 | 0.785408 | 0.785579 | 0.061511 |

| Workload | Cycles | Instructions / hooks | Fast hits | Fallbacks |
| --- | ---: | ---: | ---: | ---: |
| NOP / BRA | 100,000,002 | 14,285,715 | 7,142,858 NOP | 7,142,857 BRA |
| MOVEQ / MOVEQ / BRA | 100,000,008 | 16,666,668 | 11,111,112 MOVEQ | 5,555,556 BRA |

The NOP/BRA workload is near parity and slightly slower at the combined median. Its 50% fallback ratio makes dispatch classification overhead outweigh the narrow fast path.

The MOVEQ/MOVEQ/BRA workload shows a repeatable material improvement: the combined median is 0.785408, or approximately 21.46% faster than the repeated-step reference.

No samples were silently discarded. Secondary Tukey 1.5*IQR analysis removes five NOP/BRA and one MOVEQ/MOVEQ/BRA ratio outliers; the corresponding medians remain 1.010295 and 0.785263. The large NOP/BRA variance is consistent with host scheduling or interference, but this attribution is not proven.

The raw sample records and complete statistical summary are available on request.

## Known Limitations

- This is not a decoded cache, trace engine, or FastMem path.
- The allowlist is deliberately limited to NOP and MOVEQ.
- M68040 NOP remains on the normal dispatcher.
- The benchmark is CPU-only and host frequency is unpinned.
- Future opcode expansion requires separate proof for each CPU model's timing, special state, precise fetch/prefetch behavior, AddressBus order, hook behavior, and fallback-before-mutation behavior.

### DIFF
--- a/src/core/execute.rs
+++ b/src/core/execute.rs
@@ -756,7 +756,20 @@ impl CpuCore {
     /// automatically in this mode. For HLE interception with automatic fallback
     /// to exceptions, use `step_with_hle_handler()`.
     pub fn step<B: AddressBus>(&mut self, bus: &mut B) -> StepResult {
-        use crate::core::types::{InternalStepResult, StepResult};
+        self.step_with_dispatch(bus, dispatch_instruction)
+    }
+
+    /// Execute one precise instruction using the supplied post-fetch dispatcher.
+    ///
+    /// The helper owns the normal `step()` prologue and epilogue so internal
+    /// callers can select a dispatcher after the precise opcode fetch without
+    /// fetching the opcode twice.
+    fn step_with_dispatch<B, F>(&mut self, bus: &mut B, dispatch: F) -> StepResult
+    where
+        B: AddressBus,
+        F: FnOnce(&mut CpuCore, &mut B, u16) -> super::types::InternalStepResult,
+    {
+        use crate::core::types::InternalStepResult;
 
         self.set_precise_bus(true);
         if self.stopped != 0 {
@@ -779,7 +792,7 @@ impl CpuCore {
         }
 
         let opcode_fetch_cached = bus.last_fetch_was_cached();
-        let result = dispatch_instruction(self, bus, self.ir as u16);
+        let result = dispatch(self, bus, self.ir as u16);
         let fetch_cached = if matches!(
             self.cpu_type,
             super::types::CpuType::M68EC020 | super::types::CpuType::M68020
--- a/src/core/execute.rs
+++ b/src/core/execute.rs
@@ -856,8 +856,9 @@ impl CpuCore {
 
     /// Execute the narrow decoded subset allowed by the boundary-hook runner.
     ///
-    /// 命令取得後にのみ呼び出す。対象外は同じ取得済み opcode を既存 dispatcher
-    /// へ渡すため、fetch/prefetch を再実行しない。
+    /// Called only after the precise opcode fetch. Unsupported operations pass the
+    /// same fetched opcode to the existing dispatcher without repeating fetch or
+    /// prefetch work.
     fn dispatch_boundary_hook_decoded<B: AddressBus>(
         &mut self,
         bus: &mut B,
@@ -865,7 +866,8 @@ impl CpuCore {
     ) -> super::types::InternalStepResult {
         use super::types::{CpuType, InternalStepResult};
 
-        // Trace 状態では通常 dispatch が trace 例外と model 固有処理を担当する。
+        // Trace states remain on the normal dispatcher so it can preserve trace
+        // exceptions and model-specific behavior.
         if self.run_mode != RUN_MODE_NORMAL || (self.t1_flag | self.t0_flag) != 0 {
             return dispatch_instruction(self, bus, opcode);
         }
@@ -874,7 +876,8 @@ impl CpuCore {
             self.cpu_type,
             DecodedSimpleOp::decode(self.cpu_type, opcode),
         ) {
-            // 68040 の NOP には T0 pipeline-sync の通常処理があるため除外する。
+            // M68040 NOP is excluded because the normal path performs T0 pipeline
+            // synchronization.
             (
                 CpuType::M68000 | CpuType::M68010 | CpuType::M68020 | CpuType::M68030,
                 Some(op @ DecodedSimpleOp::Nop),
--- a/src/core/execute.rs
+++ b/src/core/execute.rs
@@ -5,7 +5,7 @@
 use super::cpu::{CpuCore, SFLAG_SET};
 use super::decode::{dispatch_instruction, needs_rollback_snapshot};
 use super::memory::AddressBus;
-use super::op_cache::BatchInnerExit;
+use super::op_cache::{BatchInnerExit, DecodedSimpleOp};
 use super::trace_jit;
 use super::types::{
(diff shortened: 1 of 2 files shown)


benletchford (changes requested):
The MOVEQ fast-path match admits reserved encodings. `DecodedSimpleOp::decode()` currently returns `Moveq` for every `0x7xxx` opcode, but bit 8 must be clear; the normal dispatcher rejects `0x7100` as illegal. A differential test on this head produced `IllegalInstruction { opcode: 0x7100 }` with zero cycles/instructions on the precise hook runner, while `run_for_cycles_with_boundary_hook()` retired it as a four-cycle instruction and invoked the boundary hook. Please guard the decoded path against bit-8-set MOVEQ encodings (or make the shared decoder reject them without regressing its other consumers) and add regression coverage that preserves the illegal exit, zero retirement/hook count, CPU state, and fetch behavior.

uchidah:
Thank you for catching this.

I updated the narrow boundary-hook MOVEQ arm to require bit 8 clear:

```rust
if opcode & 0x0100 == 0
```

Reserved group-7 encodings such as `0x7100` now fall back with the already-fetched opcode to the normal dispatcher. They do not refetch, re-enter `step()`, or perform narrow-executor mutations.

I added `boundary_hook_decoded_subset_rejects_reserved_moveq_encoding`, which directly compares the repeated-step reference and boundary decoded paths for `0x7100`. It verifies:

- `IllegalInstruction { opcode: 0x7100 }`
- zero consumed cycles, retired instructions, and instruction-hook calls
- matching CPU state, IR, rollback/SR snapshots, prefetch queue/count, guest memory, and word-read sequence
- no D0 mutation

I kept `DecodedSimpleOp::decode()` unchanged because it is also used by the simple-no-fault classification, decoded opcode cache, and trace JIT. The local guard fixes the boundary runner without broadening this PR's scope.

All CONTRIBUTING checks pass:

```text
cargo fmt --all -- --check
cargo clippy --all-targets --all-features --locked -- -D warnings
cargo test --all-features --locked
cargo package --locked
git diff --check
```

benletchford (approved):
The reserved group-7 guard and differential regression resolve the requested correctness issue: bit-8-set encodings now retain the precise illegal exit, zero retirement/hook count, CPU state, and word-read sequence. I also applied the complete patch to current v0.8.2 and reran formatting, strict all-target/all-feature Clippy, default and all-feature tests, doctests, and package verification successfully. The narrow NOP/MOVEQ dispatch preserves precise fetches, per-instruction hooks, interrupt boundaries, fallback-before-executor-mutation, and model-specific NOP tracing. Approved; left unmerged while this remains a draft.

Issue #79, opened 2026-08-06
RFC: wait-aware trace recording


rlanday:
# RFC: wait-aware trace recording

Classify poll loops at trace-recording time, stop compiling them as native
spins, report them as waits, and (in a follow-up) hand them to the
runner's advance/park machinery through a wait-typed trace exit.

## Background: the three existing wait mechanisms

Terminology first: an **A-line trap** is how classic Mac OS system calls
work. Any 68k instruction word whose high nibble is `$A` is
architecturally unimplemented; executing one raises the processor's
line-1010 exception. Mac OS encodes every Toolbox and OS entry point as
such a word — `_ExitToShell` is `$A9F4`, the same constant this runner
checks to recognize clean exits — so guest code invokes the system by
executing a two-byte `$Axxx` opcode, and the OS (here, the emulator's
HLE dispatcher) takes over at that instruction. Two properties make
A-traps load-bearing for everything below: they are where the emulator
already intercepts guest execution, making them the natural hook point
for wait detection; and they are hard boundaries for the trace JIT — a
recorded trace never contains one, because a trap leaves pure 68k
semantics for host-implemented behavior.

Systemless has three mechanisms for not executing guest wait time.

**1. Post-trap spin templates.** When the guest executes a
`_TickCount`-family A-trap, the runner pattern-matches the machine code
around the trap against seven hand-written idioms
(`try_spin_template_a`–`_g`):

```rust
// after the trap's own handler has run:
if spin_wait_fastfwd_enabled_for(yield_for_ui, tick_cap) {
    let post_trap_pc = pc.wrapping_add(2);
    let hit_cap = self.try_tickcount_spin_fastfwd(post_trap_pc, tick_cap, &mut count);
    if hit_cap { break; }
}
```

Template D, for example, recognizes `CLR.L -(SP); _TickCount; …;
BCC/BEQ.S back` by checking the literal bytes and registers —
`self.bus.read_word(target) != 0x42A7` is its anchor test — then
computes the tick the guest is waiting for and jumps time forward via
mechanism 3. The guest never executes the spin iterations. Coverage is
exactly the idioms someone hand-encoded, and the trigger is exactly a
trap.

**2. The idle-cycle prover.** For waits too complex for byte patterns —
a whole event-loop cycle — the runner proves idleness empirically. At a
wait-ish trap it snapshots the CPU, records the tick, and arms a bus
write journal; arming the journal revokes the fastmem window so no
fast-path store can bypass it:

```rust
[...]

PR #81, opened 2026-08-06
feat: classify pure poll loops at trace compile time

rlanday:
## Summary

Stage 1 of RFC #79: classify pure poll loops at trace-compile time,
refuse to compile them as native spins, report them distinctly, and
exclude them from the opportunity ranking.

A recorded self-loop is classified as a **pure poll** when every
operation either reads guest state without mutating anything except the
condition codes (memory compares, tests, and bit-tests) or is the loop's
own branching, and at least one operation reads memory. Such a loop's
exit can only be driven by memory it never writes, so executing it
faster cannot make it exit sooner — it is a wait, and it is precisely
the class that no trap-anchored wait mechanism (the spin templates, the
idle-cycle prover) can see. The classifying match is exhaustive over
every trace operation, so each future operation must declare its bucket
before the crate compiles.

## What changes

- A classified poll is not compiled (native or portable-trace); it
  executes on the decoded path as an unrecorded loop does today. This
  caps the burn multiplier at 1×: the wall time the guest spends in the
  wait is unchanged, but it no longer consumes JIT-speed instruction
  volume (measured on the motivating workload: ~18 CPU-ns/instruction of
  native spinning, one capped session feeding 19.2M of a 25M
  instructions/second budget to spin — see #79's evidence).
- The profiler reports classified waits in their own `wait loops`
  section and excludes them from the `projected_dispatches` ranking, so
  opportunity rankings stop recruiting wait idioms as optimization
  targets (which is how one became the top-ranked target of #76).

Stage 1 deliberately parks nothing and changes no execution semantics —
the only behavioral change is declining a compilation — so it carries no
correctness risk beyond that refusal. Stages 2 and 3 (wait-typed trace
exits consumed by the runner's advance/park machinery; closed-form tick
waits) are specified in #79.

## Measured effect on EV Override's rankings

A capped (`--cpu-mhz 25`) profiler session pair on the same stationary
scene — v0.8.0 baseline versus v0.8.0 plus this change — answers what
this does to the real profile:

- **Nothing currently classifies.** The wait-loops section is empty:
  every hot self-loop that completes recording in this scene mutates
  state (record scans, buffer fills) and is correctly left rankable and
  compilable. Consequently the two sessions' head tables are identical
  up to dwell length — this change costs the current EV profile nothing.
- **The benefit here is prophylactic and forward-looking.** The class
  this refuses is exactly what recent op-admission PRs were completing:
  trap-less pure polls whose compilation multiplied burn (measured on
  the #76 candidate: ~18 CPU-ns/instruction of native spinning, one
  capped session feeding 19.2M of its 25M instructions/second budget to
  spin — evidence in #79). Any future admission that would finish such a
  loop now classifies it instead of compiling it, and the ranking can no
  longer recruit it.
- **Stated residue.** Trap-containing wait idioms remain ranked: the
  session pair's top slots are 21/22-op prefixes blocked at `486D` (8.1M
  and 7.7M projected) and the `42A7` template-D idiom at 1.0M projected
  — blocked recordings, not completed ones, so Stage 1's classifier
  never sees them. Their ranking inflation is Stage 2/3 scope per #79.
- A genuine target the pair surfaced: a `41F0`-blocked head with an
  identical 937,836 hits in both sessions — a deterministic boot-phase
  loop, unaffected by wait dynamics, and the profile's legitimate top
  opportunity.

## Whole-application measurement

A deterministic headless ladder (300M guest instructions, v0.8.1
baseline versus this PR, both with `trace-profile`) settles the safety
question in the strongest available form:

- **The trace-profile report is byte-identical to baseline** — every
  head row, every failed- and compiled-shape table, every decoded-site
  table, and every total (`backward_hits` 624,279, `rejected_hits`
  615,734, `native_calls` 47,671, `jit_retired` 11,146,241). The final
  framebuffer SHA-256 matches as well. With the classifier active,
  execution is bit-identical to master over a 300M-instruction
  boot-through-gameplay run. Wall time 196.5 s versus 199.9 s user CPU,
  inside the ±19% host-noise band measured for this harness in #83 and
  meaningless next to the identity result.
- **Nothing classifies in this workload either.** The `wait loops`
  section prints only when non-empty, and it is absent — matching the
  capped GUI session pair reported above. Both measured workloads agree:
  every hot self-loop that completes recording in this title mutates
  state and is correctly left compilable.

Stated plainly, so the trade is easy to judge: **this PR's measured
benefit on today's workloads is zero, and that is the intended result
for a conservative veto that matches nothing.** Its value is the two
things measurement cannot show — the regression class it forecloses
(each future op admission can no longer convert a trap-less wait into a
native spin, capping that burn multiplier at 1×; #79 documents one such
episode at 19.2M of a 25M instruction/second budget) and the ranking
hygiene that stops waits being recruited as optimization targets. The
cost side is correspondingly small: one refused compilation, no
execution-path change, bit-identical output.

Methodology note: whole-application sampling in this project uses
`--cpu-mhz 25` (uncapped GUI execution is elastic, so CPU share
conflates efficiency with throughput). The headless ladder above is a
complementary instrument — it fixes total guest work exactly and proves
work-identity by framebuffer hash, which is what makes a bit-identity
claim possible at all; the capped GUI ladder remains canonical for
interactive-scene claims, where wait dynamics and audio differ.

## Acceptance criteria from #79 review

- **Conservative flag provenance**: every flag any branch in the loop
  consumes must have been written by one of the loop's memory reads; a
  `BTST`-only loop branching on carry does not classify (BTST writes
  only Z), and a loop whose only backward branch is unconditional does
  not classify (no memory-driven exit). Covered by unit tests including
  partial-flag and preserved-flag shapes.
- **False-positive/false-negative shapes**: register-mutating loops
  (counters, destination-writing memory ALU, scratch-register loads),
  branch-only calibration loops, and `DBcc` counted waits all classify
  false; memory-compare, memory-test, bit-test-on-Z, and
  guarded-interior-exit polls classify true.
- **Unchanged execution**: the end-to-end test runs a classified poll
  for an exact instruction budget on the decoded path with identical
  architectural results.
- **Distinct accounting**: classified heads report in their own
  `wait loops` profiler section and leave `projected_dispatches`.
- **Classification requires the recording to have closed back to its
  head.** `self_loop` is deliberately permissive for codegen — it is also
  true when the last recorded operation is merely a branch whose static
  target is the head, which happens on a *fall-through* path — so the
  wait classification gates on `recorded_exit_pc == Some(start_pc)`
  instead. Covered by an end-to-end regression.
- **Wait accounting is shape-keyed and independent of blocker
  accounting.** Classified waits live in their own `wait_shapes` table
  (keyed by the loop's exact operations, like failed shapes, capped and
  overflow-counted); classifying a wait never touches `blocker_pc`,
  `prefix_ops`, or `reject_reason`, and recording a blocker never erases
  wait shapes — so a head that records both kinds of shape reports in
  both sections, in whichever order the recordings happen (both orders
  covered end-to-end, plus a deterministic note-level test of the
  arithmetic in both orders, mutation-tested). Per-hit attribution
  (`wait_hits`) follows the most recent completed recording, and
  `projected_dispatches` subtracts wait-attributed volume, so a wait
  cannot recruit its head into the ranking. The ranking prints a row when
  it carries opcode-admission opportunity (a recorded blocker or positive
  projected dispatches); pure waits report only in the wait section.
- **The veto/parking invariant**: this predicate is used only to decline
  compilation; it is not a parking proof and is not exported for one.
  Parking is Stage 2 scope in a separate design (per the #79 review),
  where wait exits must prove complete live-state invariance or model
  every permitted delta.
- **On #76**: measured honestly, the #76 burn concentrated in
  store-containing loops whose invocation frequency is wait-driven —
  correctly not classifiable — so this PR does not claim to prevent that
  episode synthetically; #76 remains gated on post-classification
  whole-application measurement.

## Validation

- classifier unit tests: memory-compare and memory-test polls classify;
  branch-only (calibration) loops, register-mutating loops, and
  destination-writing memory ALU loops do not
- end-to-end test: a `CMP.W (d16,A5),D1; BEQ` poll runs its exact
  instruction budget on the decoded path, is flagged `wait` in the
  profile, does not appear among compiled shapes, and is absent from the
  opportunity ranking
- end-to-end regression: the same recorded operations reached by a
  *fall-through* — the back edge not taken, execution continuing into an
  untraceable `MULS.W (0,A0,D0.W),D1` — are **not** classified as a wait,
  the blocker stays attributed at its address, and the head stays in the
  opportunity ranking

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -419,6 +419,9 @@ pub(crate) enum JitTraceOp {
 /// it is reported.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum RegionRejectReason {
+    /// The recorded self-loop is a pure poll: a wait, deliberately left
+    /// uncompiled. Reported as a wait rather than as a silent rejection.
+    WaitLoop,
     NoTraceTerminal,
     TooShort,
     IndirectJsrTooShort,
@@ -1007,6 +1010,14 @@ impl TraceJit {
         use super::trace_profile::TraceRejectReason as Public;
         // Exhaustive so a new compile gate must declare how it is reported.
         match reason {
+            // A classified wait is routed to `note_wait_loop` by
+            // `finish_recording` before this mapping is consulted; the
+            // profiler represents it through the shape-keyed wait table
+            // flag and the wait-loops report section, never as a silent
+            // rejection.
+            RegionRejectReason::WaitLoop => {
+                unreachable!("classified waits are reported via note_wait_loop")
+            }
             RegionRejectReason::NoTraceTerminal => Public::NoTraceTerminal,
             RegionRejectReason::TooShort => Public::TooShort,
             RegionRejectReason::IndirectJsrTooShort => Public::IndirectJsrTooShort,
@@ -1099,7 +1110,15 @@ impl TraceJit {
             // decoded path outranks the compile-stage reason: it bounds how
             // far this head can ever record, which no opcode coverage can
             // change.
-            if end != RecordingEnd::Blocker {
+            if reason == RegionRejectReason::WaitLoop {
+                // A classified wait is accounted as a wait, keyed by the
+                // recorded shape so the accounting is independent of any
+                // blocker shapes at the same head, in either recording
+                // order. It keeps its own report section and its exclusion
+                // from the opportunity ranking, rather than being filed as
+                // a generic rejection.
+                super::trace_profile::note_wait_loop(start_pc, cpu_type, recorded_shape);
+            } else if end != RecordingEnd::Blocker {
                 let reason = match end {
                     RecordingEnd::Stopped(cause) => Self::profile_stop_reason(cause),
                     _ => Self::profile_reject_reason(reason),
@@ -1319,6 +1338,23 @@ impl TraceJit {
             })
         {
             return Err(RegionRejectReason::LinearMemoryAlu);
+        }
+
+        // A pure poll loop burns wall time by design: its exit depends only
+        // on memory it never writes, so executing it faster cannot make it
+        // exit sooner. It is also exactly the class no trap-anchored wait
+        // detector can see. Compiling it as a native spin multiplies the
+        // instructions burned per wall second while the guest waits, so
+        // refuse the compilation and leave the loop on the decoded path.
+        //
+        // This requires the recording to have actually gone round, not
+        // merely to end in a branch whose static target is the head.
+        // `self_loop` is true for a fall-through path as well, and a
+        // recording that fell through to an unsupported operation is a
+        // different shape entirely: reporting it as a wait would remove a
+        // real blocker from the opportunity ranking.
+        if recorded_exit_pc == Some(start_pc) && is_pure_poll_loop(&ops) {
+            return Err(RegionRejectReason::WaitLoop);
         }
 
         let needs_window = ops.iter().any(|op| {
@@ -1734,6 +1770,116 @@ struct CompileParams<'a> {
     aligned_only: bool,
     #[cfg_attr(any(not(feature = "jit"), target_family = "wasm"), allow(dead_code))]
     address_mask: u32,
+}
+
+const CC_N: u8 = 0b1000;
+const CC_Z: u8 = 0b0100;
+const CC_V: u8 = 0b0010;
+const CC_C: u8 = 0b0001;
+
+/// Condition codes a 68k conditional branch reads. `T`/`F` read nothing.
+fn branch_flags_read(condition: u8) -> u8 {
+    match condition & 0xF {
+        0 | 1 => 0,              // T / F
+        2 | 3 => CC_C | CC_Z,    // HI / LS
+        4 | 5 => CC_C,           // CC / CS
+        6 | 7 => CC_Z,           // NE / EQ
+        8 | 9 => CC_V,           // VC / VS
+        10 | 11 => CC_N,         // PL / MI
+        12 | 13 => CC_N | CC_V,  // GE / LT
+        _ => CC_N | CC_V | CC_Z, // GT / LE
+    }
+}
+
+/// A recorded self-loop is a pure poll when it consists of memory reads
+/// that mutate nothing except the condition codes, followed by a single
+/// terminal conditional branch whose consumed flags were all written by
+/// those reads (conservative flag provenance: a BTST-only loop branching
+/// on carry does not classify, because BTST writes only Z). Interior
+/// branches disqualify: a guarded interior branch means the head can
+/// record multiple dynamic shapes, and the profiler's per-head wait flag
+/// is only sound when the recorded path is structurally unique. Such a
+/// loop's exit can only be driven by memory it never writes, so
+/// executing it faster cannot make it exit sooner — it is a wait. The
+/// match is exhaustive so every future trace operation must declare its
+/// classification here.
+fn is_pure_poll_loop(ops: &[TraceBuildOp]) -> bool {
+    let Some((terminal, body)) = ops.split_last() else {
+        return false;
+    };
+    let JitTraceOp::Branch { condition, .. } = terminal.op else {
+        return false;
+    };
+    let consumed = branch_flags_read(condition);
+    if consumed == 0 {
+        // An unconditional terminal branch has no memory-driven exit.
+        return false;
+    }
+    let mut mem_written_flags: u8 = 0;
+    for op in body {
+        match op.op {
+            // No-ops mutate nothing.
+            JitTraceOp::Nop => {}
+            // Interior branches make the recorded path non-unique for
+            // this head; per-head wait accounting would then erase
+            // opportunity data belonging to other shapes.
+            JitTraceOp::Branch { .. } => return false,
+            // Memory-reading compares and tests write NZVC only; the
+            // polled value is the only input that can change.
+            JitTraceOp::TstMem { .. }
+            | JitTraceOp::CmpiWordMem { .. }
+            | JitTraceOp::AddrCmpMemToReg { .. }
+            | JitTraceOp::AluMemToReg {
+                op: JitBinaryOp::Cmp,
+                ..
+            }
+            | JitTraceOp::AnDispUnary {
+                op: JitUnaryOp::Tst,
+                ..
+            } => mem_written_flags |= CC_N | CC_Z | CC_V | CC_C,
+            // A bit test writes only Z.
+            JitTraceOp::AnDispBit {
+                op: JitBitOp::Test, ..
+            } => mem_written_flags |= CC_Z,
+            // Everything else mutates registers or memory, or transfers
+            // control in a way that can end the wait from inside: any of
+            // those makes the loop's progress self-driven, so it is not a
+            // pure poll.
+            JitTraceOp::MoveReg { .. }
+            | JitTraceOp::Moveq { .. }
+            | JitTraceOp::UnaryDataReg { .. }
+            | JitTraceOp::AddqSubqReg { .. }
+            | JitTraceOp::AddqSubqAddr { .. }
+            | JitTraceOp::BinaryDataReg { .. }
+            | JitTraceOp::BinaryImmediateDataReg { .. }
+            | JitTraceOp::MulWordDataReg { .. }
+            | JitTraceOp::MulWordImmediate { .. }
+            | JitTraceOp::MulLongDataReg { .. }
+            | JitTraceOp::AddrDataReg { .. }
+            | JitTraceOp::AddrCmpImmediate { .. }
+            | JitTraceOp::LeaAn { .. }
+            | JitTraceOp::LeaIndex { .. }
+            | JitTraceOp::AddSubxReg { .. }
+            | JitTraceOp::BitReg { .. }
+            | JitTraceOp::Exg { .. }
+            | JitTraceOp::Ext { .. }
+            | JitTraceOp::Extb { .. }
+            | JitTraceOp::SccDataReg { .. }
+            | JitTraceOp::ShiftReg { .. }
+            | JitTraceOp::Swap { .. }
+            | JitTraceOp::Dbcc { .. }
+            | JitTraceOp::IndirectJsr { .. }
+            | JitTraceOp::MoveMem { .. }
+            | JitTraceOp::MovemWordPostInc { .. }
+            | JitTraceOp::AluMemToReg { .. }
+            | JitTraceOp::AnDispUnary { .. }
+            | JitTraceOp::AddRegToMem { .. }
+            | JitTraceOp::MemAddqSubq { .. }
+            | JitTraceOp::AnDispBit { .. }
+            | JitTraceOp::PeaDisp { .. } => return false,
+        }
+    }
+    mem_written_flags != 0 && consumed & !mem_written_flags == 0
 }
 
 /// Attempt to execute a compiled trace at the current PC. See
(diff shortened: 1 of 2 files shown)


rlanday:
Updated per the #79 review: conservative flag provenance is now implemented (every flag a branch consumes must be written by the loop's memory reads — `BTST`+`BCS` does not classify, `BTST`+`BNE` does; unconditional-only loops do not), the false-positive/negative shape matrix is in the unit tests, and the body now records the veto-vs-parking invariant and the honest #76 scope note. Stages 2/3 will be a separate cross-repository design issue as requested.

benletchford (dismissed):
Two profiler-correctness issues block the current classifier.

First, this head does not yet implement the flag provenance described in the RFC follow-up. `reads_memory` only proves that some admitted read occurred, while every `Branch` is accepted unconditionally. Consequently `BTST` on memory followed by `BCS` classifies even though `BTST` writes only Z and the branch consumes the entry C flag; `TST` followed by unconditional `BRA` also classifies although the read cannot drive loop exit. Please model the flags each admitted read defines and each recorded branch consumes (including partial/preserved flags), or restrict the initial classifier to a shape/condition whose complete dependency is proven. Add positive `BTST`+BEQ/BNE and negative `BTST`+BCS plus unconditional-branch coverage.

Second, `wait_loop` is a sticky boolean on the aggregate `(start_pc, cpu_type)` row. The existing profiler explicitly supports different failed/compiled dynamic shapes at one head; once any recorded path sets this bit, all of that head’s rejected hits and opportunity data are removed from the ranking, including a later non-poll path. Either restrict eligibility so a head has one straight-line memory-read/terminal-branch shape, or record wait classification and counters by exact trace shape and preserve non-wait opportunity data. A regression should exercise one head that records both a poll path and a state-mutating path and verify only the poll shape is reported/excluded.

benletchford (changes requested):
The updated head resolves the flag-provenance concern: it models N/Z/V/C reads and writes and covers `BTST`+BNE, `BTST`+BCS, and unconditional branches.

The remaining blocker is the aggregate, sticky `wait_loop` bit. `TraceProfileRow` is keyed only by `(start_pc, cpu_type)`, while this profiler deliberately supports multiple dynamic recording shapes at one head. An interior guarded branch can produce a pure-poll path and a state-mutating path; once either a later adaptive re-record or a cache-slot reprobe classifies the poll path, `note_wait_loop()` removes the entire head—including historical or future non-wait opportunity data—from the ranking. The current test exercises only a single straight-line shape and cannot catch this.

For the initial version, either restrict classification to a single terminal branch with no interior guarded branches (making the recorded path structurally unique), or make wait accounting shape-specific and preserve non-wait shape contributions. Please add a regression with one head that can record both a poll path and a mutating path and verify the non-wait opportunity remains visible.

rlanday:
Addressed via your option (a): classification now requires a **single terminal conditional branch with no interior branches**, which makes the recorded path structurally unique per head and the per-head wait flag sound — a head that could record both a poll path and a mutating path (the interior-guard shape) no longer classifies at all, preserving its opportunity data. The former positive interior-guard test is now the regression you asked for, asserting non-classification for exactly that dual-shape hazard, alongside the flag-provenance matrix (BTST+BNE positive, BTST+BCS negative, unconditional-terminal negative). Also rebased onto v0.8.1; the exhaustive classification match now covers the merged PeaDisp operation. Shape-keyed wait accounting (your option b) is noted in #79 as the Stage-2-era generalization if multi-shape wait heads turn out to matter in practice.

rlanday:
Both review rounds are addressed (conservative flag provenance; single terminal conditional branch with no interior branches, which makes the recorded path structurally unique per head and the per-head bit sound), and the whole-application gate in the body is now cleared, so this is out of draft.

The new evidence is a deterministic headless ladder — 300M guest instructions, v0.8.1 versus this branch, both with `trace-profile`:

**The trace-profile report is byte-identical to baseline.** Every head row, every failed/compiled shape table, every decoded-site table, every total, and the final framebuffer SHA-256. With the classifier active, execution is bit-identical to master across a full boot-through-gameplay run.

Two honest points that follow, both now in the body:

1. **Nothing classifies in this workload either** — same as the capped GUI pair. So this PR's measured benefit today is zero, which is what a conservative veto matching nothing *should* do. The value is the regression class it forecloses (no future op admission can turn a trap-less wait into a native spin) plus ranking hygiene, and the cost is one refused compilation with no execution-path change.
2. Relatedly, from the #83 investigation: the profiler cannot currently attribute recordings that end at a trap or die in a compile gate — filed as #84. That gap is why a head there looked one admission from compiling when it was structurally uncompilable. It doesn't affect this PR's correctness, but it does affect how much weight the ranking tables deserve, which is squarely this PR's territory.

benletchford (changes requested):
The current head resolves the earlier flag-provenance and multi-shape accounting concerns. The remaining blocker is integration with current `master`: this branch now conflicts after #83 and #85.

When rebasing, classify the new `JitTraceOp::LeaIndex` as non-poll because it mutates an address register, and route the poll-loop veto through the new reasoned rejection path with a distinct `RegionRejectReason`/`TraceRejectReason` mapping. It must retain the wait-loop report and ranking exclusion rather than being attributed as `Backend` or disappearing from the new silent-rejection accounting. Please add or update a trace-profile regression that asserts that attribution, then rerun the full checks.

rlanday:
Rebased onto `master` at 0.8.2 and integrated with #83 and #85 as requested. Point by point:

**`LeaIndex` classified as non-poll.** It writes an address register, so it lands with the other register-mutating operations that make a loop's progress self-driven. Worth noting the exhaustive match did its job unprompted: the crate refused to compile after the rebase until the new operation declared its bucket, which is exactly the property that match exists to guarantee.

**The veto now routes through the reasoned rejection path.** `compile_decoded_ops_reason` returns a new `RegionRejectReason::WaitLoop`, mapped to a new public `TraceRejectReason::WaitLoop`. Both mappings are exhaustive, so a future gate or stop cause has to declare how it is reported.

**A classified wait is still accounted as a wait.** The reporting branch checks the wait reason first and calls `note_wait_loop`, so such a head keeps its own `wait loops` section and its exclusion from the projected-dispatch ranking, and is deliberately *not* filed into the silent-rejection table — no double-reporting, and no attribution as `Backend`.

**Regression added**, extending the existing end-to-end wait test: it now asserts the head does not appear in `silent_rejections` and that the report carries no `backend` attribution, alongside the existing assertions that it is flagged as a wait, is not compiled, and is absent from the ranking. I mutation-tested it rather than trusting that it passes — disabling the wait-routing branch makes it fail, so it is load-bearing.

Full checks rerun on the rebased head: `cargo fmt --check`, `cargo test --all-features`, `cargo test`, `cargo clippy --all-targets --all-features -- -D warnings`, and the wasm target check, all clean. One thing the rebase surfaced: the end-to-end test needs `feature = "trace-profile"` in its `cfg`, and the default-feature build catches it immediately if that is missing.

The measured picture in the body is unchanged — nothing classifies in either the capped GUI session pair or the deterministic headless workload, so this PR's benefit remains prophylactic and its cost remains one refused compilation. #85 landing does sharpen the argument slightly: with silent rejections now attributed, the ranking's remaining distortion is exactly the wait class this PR removes.

benletchford (changes requested):
The current head still misclassifies a fall-through recording as a wait. `compile_decoded_ops_reason()` sets `self_loop` when the last branch static taken target equals `start_pc`, even when that branch was not taken and `recorded_exit_pc` is elsewhere; `finish_recording()` also clears the last branch `expected_taken` before classification. Thus a recorded path such as `CMP.W (d16,A5),D1; BEQ head` with the `BEQ` not taken and the following operation unsupported (or a trap/host boundary) reaches classification with only `[CMP, BEQ]`, passes `is_pure_poll_loop()`, and calls `note_wait_loop()` even for `RecordingEnd::Blocker` because the wait branch precedes the blocker filter. The real non-wait path and blocker are then removed from the opportunity ranking. Please require that a wait recording actually exited back to its head (for example, gate this classification on `recorded_exit_pc == Some(start_pc)`/the corresponding region end), or make the accounting shape-specific. Add an end-to-end profiler regression where the conditional back-edge falls through into an unsupported state-mutating operation and assert that the head is not marked as a wait and its blocker remains ranked.

rlanday:
Good catch, and it was worse than a false positive — thank you for working through the fall-through case.

**Fixed.** The wait classification now gates on `recorded_exit_pc == Some(start_pc)` rather than `self_loop`, so it requires dynamic evidence that the recording actually went round. `self_loop` itself is unchanged, so no compilation behaviour moves: it stays permissive for codegen, which is correct there, since a region ending in a branch back to the head can be compiled as a loop with the branch guard handling the not-taken exit.

Your diagnosis was exactly right on both halves. `self_loop`'s static clause is true on a fall-through path, and `finish_recording` clears the terminal branch's `expected_taken` before classification, so by the time `is_pure_poll_loop` runs there is nothing left to distinguish "spun on memory" from "fell through". The wait branch also preceded the blocker filter in the reporting, so a `RecordingEnd::Blocker` recording could be filed as a wait; with the gate fixed that combination is unreachable, since a blocker-ended recording never has its exit at the head.

Worth stating why this mattered more than a mislabel: a head marked as a wait is partitioned out of the opportunity ranking. So the failure mode was a *legitimate* head — one that fell through into an unsupported state-mutating operation — having its real blocker hidden from the ranking. A change whose entire purpose is stopping the ranking from counting phantom opportunity would have started concealing genuine opportunity instead.

**Regression added** as requested: `fall_through_past_a_back_edge_is_not_a_wait` records `CMP.W (d16,A5),D1; BEQ head` with the back edge never taken, falling through into an untraceable `MULS.W (0,A0,D0.W),D1`, then asserts the head is not flagged as a wait, that the blocker stays attributed at `0x0106`, and that the head remains in the opportunity ranking. I mutation-tested it rather than trusting a green run — reverting the gate to `self_loop` makes it fail with exactly the symptom you described.

Full checks rerun: `cargo fmt --check`, `cargo test --all-features` (835 tests), `cargo test`, `cargo clippy --all-targets --all-features -- -D warnings`, and the wasm target check, all clean.

Separately, and relevant to how much this PR is worth: I measured the whole-application ceiling for the operation-admission strategy on a deterministic 300M-instruction workload. Traces retire 3.72% of executed instructions, and the *entire* top-40 stranded opportunity is 0.80% — the largest single head is 0.14%. Trace formation, not the operation set, is the binding constraint, and 41 of 42 attributed silent rejections are trap-bounded. That does not change this PR (its value was always the regression class it forecloses plus ranking hygiene), but it does mean I have stopped treating `projected_dispatches` as a work queue.

benletchford (changes requested):
The `recorded_exit_pc` gate fixes classification of a fresh fall-through recording, but the per-head `wait_loop` bit is still sticky after an earlier classified recording. I reproduced this end to end on the current head: first classify the taken `CMP.W (d16,A5),D1; BEQ head` loop at `0x0100`; evict its direct-mapped rejected slot with a head at `0x2100`; then revisit `0x0100` with the compare unequal so it falls through into the unsupported `MULS` at `0x0106`. The second recording correctly sets `blocker_pc == Some(0x0106)`, but `row.wait_loop` remains true, so `report()` still partitions the whole head into the wait section and hides the real opportunity. Please make non-wait accounting clear or supersede the aggregate wait state, or key wait accounting by shape. Add this wait-then-cache-eviction-then-fall-through sequence as a regression and assert the blocker remains in the opportunity ranking.

rlanday:
Rebased onto `master` at 0.8.3 — no conflicts with #78's precise decoded dispatch. Full suite green (840 tests), clippy and the wasm check clean.

No functional change from the previous head; this is purely to keep it current.

benletchford (changes requested):
The new public `TraceRejectReason::WaitLoop` is currently unobservable. `finish_recording()` special-cases `RegionRejectReason::WaitLoop` before `profile_reject_reason()` / `note_silent_rejection()`, and `note_wait_loop()` only flips the boolean, so a wait row has `reject_reason == None` even though that field documents the most recent no-trace reason; the new variant mapping and label are dead API. Please either record `WaitLoop` in a shape-sound representation (without reintroducing the aggregate sticky-state bug), or remove the public variant/mapping and document `wait_loop` as the sole representation. Add a snapshot assertion for the chosen contract.

rlanday:
Fixed via your option (b) — the public variant is gone and `wait_loop` is the documented sole representation.

Your reading was exactly right, and it was a defect worth catching: `finish_recording` routes `RegionRejectReason::WaitLoop` to `note_wait_loop` before `profile_reject_reason` is ever consulted, so the public `TraceRejectReason::WaitLoop`, its label, and the mapping arm were unreachable from any input — dead API that would have shipped as a trap for downstream matchers.

What changed:

- `TraceRejectReason::WaitLoop` and its `"wait-loop"` label are removed. Since the variant only ever existed on this branch, upstream's published enum is unchanged.
- The internal `RegionRejectReason::WaitLoop` remains (it is how the veto routes), and its arm in `profile_reject_reason` is now `unreachable!("classified waits are reported via note_wait_loop")` with a comment stating the routing invariant — so the mapping stays exhaustive and a future refactor that breaks the routing fails loudly rather than silently reviving the dead variant.
- The contract is documented on both fields: `wait_loop` states it (with the wait-loops report section) is the sole representation of a classified wait, and `reject_reason` states classified waits never appear there.
- Snapshot assertion added to the end-to-end wait test as requested: alongside the existing checks (flagged as wait, uncompiled, absent from the ranking and from `silent_rejections`), it now asserts `row.reject_reason == None` — a classified wait never masquerades as a silent-rejection reason.

Full checks rerun: 840 tests all-features, default suite, `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, and the wasm target check, all clean.

rlanday:
Your 06:11 review (the sticky per-head `wait_loop` bit) is now also addressed, in the same head as the option-(b) fix — and an apology for the sequencing: my earlier comment tonight responded only to your latest review. I had been checking the review *decision* rather than enumerating review bodies, so the 06:11 round went unread until I re-checked; both are in now.

**Fix:** the flag now reflects the most recent completed recording, which is your "supersede the aggregate wait state" option. `note_blocker`, `note_silent_rejection`, and `note_compiled` all clear `wait_loop`, so a head whose slot was evicted and which then records a non-wait path returns to the opportunity ranking instead of staying hidden in the wait section. The superseding rule and its scope are documented on the field.

**Regression:** your exact reproduction, end to end — the taken `CMP.W (d16,A5),D1; BEQ` loop at `0x0100` classifies as a wait; a backward branch to `0x2100` evicts the direct-mapped slot (both heads map to slot `0x80`); an unequal-compare revisit falls through into `MULS.W (0,A0,D0.W),D1`. It asserts `wait_loop` is superseded, `blocker_pc == Some(0x0106)`, and the head appears in the ranking section of the report. Mutation-tested: disabling the three superseding sites fails it with precisely the sticky symptom you described.

Full checks on the combined head (this + the dead-API removal): 841 tests all-features, default suite, fmt, clippy `-D warnings`, wasm check, all clean.

benletchford (changes requested):
The superseding fix is still asymmetric. I reproduced the reverse order on the current head: first let `0x0100` take the unequal `CMP.W ...; BEQ` fall-through and record blocker `0x0106`; evict its direct-mapped slot with `0x2100`; then revisit `0x0100` with an equal compare so the taken loop classifies as a wait. `note_wait_loop()` sets the aggregate `row.wait_loop`, but the earlier `blocker_pc` and cumulative opportunity counters remain. `report()` then partitions the entire head into the wait section and hides that real non-wait opportunity. The added regression fails because `0x0100` disappears from the ranking while `blocker_pc == Some(0x0106)` remains. Clearing the bit only when a later non-wait shape arrives cannot make per-head accounting sound in both recording orders; please make wait/opportunity accounting shape-specific (or otherwise preserve non-wait contributions independently) and cover both orderings.

rlanday:
You were right that no amount of clearing an aggregate bit could be sound in both orders, and the redesign follows your first option: wait accounting is now **shape-specific and fully independent** of blocker accounting.

**The design:**

- Classified waits are stored in a **shape-keyed table** (`wait_shapes`, keyed by the loop's exact operation sequence, like `failed_shapes`), capped and overflow-counted. Classification never touches `blocker_pc`, `prefix_ops`, or `reject_reason`; recording a blocker never erases wait shapes. Nothing at the head level can hide either accounting stream in either order.
- The row's boolean is gone. In its place, `wait_hits: u64` counts backward-branch volume attributed to a live wait shape (attribution follows the most recent completed recording — attribution only; presentation never derives from it), and **`projected_dispatches` subtracts wait-attributed hits**, so wait volume cannot recruit a head into the ranking — closing the inflation half of the problem, which the old partition only masked.
- The ranking prints a row when it carries opcode-admission opportunity: a recorded blocker, or positive projected dispatches. A pure wait has neither and reports only in the wait section; a mixed head appears in **both** sections, each reflecting its own stream.

**On your reproduction**, one honest finding from building it: in both orderings the fall-through's `[CMP, BEQ]` prefix is itself a compilable two-op self-loop, so `finish_recording` compiles it and the fall-through volume subsequently flows through **guard exits** rather than rejected hits — the head's projected is legitimately ~0 in this scenario, and it is the recorded blocker that keeps it visible in the ranking (hence the blocker-presence clause in the filter, not a projected threshold). Both orderings converge to the same reported end state: blocker ranked, wait shape in the wait section, wait volume excluded from the metric.

**Regressions**, all passing:
- Your reverse order end-to-end (`blocker_then_eviction_then_wait_keeps_the_blocker_ranked`): blocker preserved and ranked, wait shape present, metric equals `(rejected − wait_hits) × prefix`.
- The forward order, updated to assert the true flow (blocker attributed and ranked; phase-1 wait shape survives independently; `rejected_hits == wait_hits` since fall-through volume is guard-exit volume).
- A deterministic note-level test of the arithmetic in **both orders** (5 non-wait edges × 2-op prefix = 10 projected, unchanged by 7 wait edges), mutation-tested: removing the subtraction fails it with the inflated value.
- The pure-poll and fall-through e2e tests, updated to the shape table.

Full gates on the head: fmt, all-features and default suites, clippy `-D warnings`, wasm check.

benletchford (changes requested):
The shape table fixes blocker/wait coexistence, but the new wait-section `wait_hits` column is not shape-sound. For every `WaitShapeProfileRow`, `report()` finds a `TraceProfileRow` by `start_pc` alone and prints that head-level aggregate. Because rows and wait shapes are keyed by `(start_pc, cpu_type)`, recording the same PC under two CPU models can print the first model’s count on both lines; and multiple wait shapes at one `(PC, CPU)` each repeat the same aggregate as though it belonged to that shape. This makes the central new report misattribute wait volume even though the snapshot retains the correct keys. Please either report `wait_hits` once per `(start_pc, cpu_type)` separately from shape recordings or attribute hits by exact shape; in either design, include CPU type in the lookup/presentation and add a regression with the same PC under two models (and multiple shapes) carrying distinct counts.

rlanday:
Fixed, and rebased onto 0.9.0 (the rebase merged cleanly over #89's now-upstream changes — verified both features coexist rather than trusting the clean exit: `exit_opcode` capture and the wait routing are both present and all suites pass).

Both defects you identified are addressed by restructuring the wait section to one line per `(start_pc, cpu_type)` head, with shape rows nested under it:

- **Head volume prints exactly once per head**, with the lookup keyed by `(start_pc, cpu_type)` — never by PC alone — so two CPU models at the same PC each carry their own count.
- **Shape lines carry per-shape recording counts only** (`shape ops=N records=M`), so shape recordings and head hit-volume are visibly different quantities and the aggregate is never repeated as though it belonged to a shape.

The regression is your reproduction exactly: the same PC under M68000 (one shape, 3 wait hits) and M68040 (two distinct shapes, one recorded twice, 7 wait hits). It asserts each model's line carries its own count and that shape lines read `records=1` / `records=2` rather than the head aggregate. Mutation-tested: degrading the lookup to PC-only fails it with precisely your symptom — the M68040 line printing the M68000 count.

Snapshot consumers were already sound (rows and shapes both carry `cpu_type`); the defect was confined to the report presentation, and the regression pins both layers.

Full gates on the rebased head: fmt, both suites, clippy `-D warnings`, wasm check.

benletchford (changes requested):
PR #88 has now landed on master as e1eb2ae. Although this branch and #88 merge textually, the combined crate fails to compile at `is_pure_poll_loop`: the exhaustive match does not cover the new `JitTraceOp::MulWordImmediate` variant (E0004 at `src/core/trace_jit.rs:1820`). Please rebase onto current master, classify `MulWordImmediate` as non-poll because it writes Dn, and rerun the full local gate. The current head passed that gate against 0.9.0, but its existing checks do not validate the new combined tree.

rlanday:
Rebased onto 0.9.1 and fixed — and this failure mode is the one the exhaustive match exists to produce: #88's `MulWordImmediate` landed, and the classifier refused to compile until the new operation declared its bucket. It is classified non-poll (it writes a data register, making the loop's progress self-driven), joining `MulWordDataReg` and `LeaIndex` in that arm.

Congratulations on the #88 merge, and thank you for the six rounds here — every one of them found something real.

Full local gate rerun on the combined tree as requested: `cargo fmt --check`, `cargo test --all-features` and the default suite, `cargo clippy --all-targets --all-features -- -D warnings`, and the wasm target check, all clean.

benletchford (approved):
Verified the current head resolves the requested classifier, fall-through, shape-accounting, CPU-keying, and master-integration concerns. The complete diff and surrounding execution/profiler paths are sound, CI is green, and the full local formatting, Clippy, default/all-feature test, strict rustdoc, and package gates pass.

PR #83, opened 2026-08-06
perf: JIT brief-indexed LEA

rlanday:
## Summary

Admit `LEA (d8,An,Xn),An` as a register-only trace operation. The brief
extension word is decoded once while recording; execution computes
`An + index + d8` with word indexes sign-extended and the scale applied,
writes the destination address register, accesses no memory, and changes
no condition codes. Per-decode cycles: twelve on pre-68020 models
(M68000UM), four after, following the existing `LeaAn` convention.

This follows the merged #68/#69/#72/#73/#74 and applies directly to
master at v0.8.1.

## Why this operation

In capped (`--cpu-mhz 25`) EV Override profiling, this form blocks a
boot-phase head with an **identical 937,836 rejected recordings in every
session** — deterministic across scene states, capped and uncapped,
which is also what proves it is genuine bounded work rather than a wait
(a wait's hit count scales with execution speed and idle time; this one
never varies). After the wait-aware reading of the profile established
by #79/#81, it ranked as the top legitimate opportunity: 2.81 million
projected trace operations at a three-operation prefix. (The
whole-application measurement below revises this honestly: the LEA was
that head's *first* obstacle, not its only one.)

Relationship to #81: this operation writes a register, so the
classifier there correctly treats it as loop progress (not a poll); its
exhaustive classification match will force this variant to declare that
bucket when the two branches meet, which is that design working as
intended.

## Correctness and safety

- Decode only the brief-indexed control form; full extension words
  (68020+) keep their existing dispatch.
- Sign-extend word index registers, preserve long ones, apply the scale
  and signed 8-bit displacement exactly once — the same address
  arithmetic as the merged indexed compare and CMPI operations.
- Register-only: no window interaction, no bail paths, no flags.
- Portable and native execution with parity tests, including the
  negative sign-extended word-index case.

## Performance

Focused benchmark: `indexed-lea`, an index-sweeping loop that wraps at
an exact instruction boundary and asserts the final index, base, and
computed address.

Five alternating runs against a baseline worktree at v0.8.1 master
carrying an identical copy of the benchmark, M guest instructions/s:

| Run | v0.8.1 base | this PR |
|---:|---:|---:|
| 1 | 29.8 | 289.4 |
| 2 | 24.6 | 353.0 |
| 3 | 35.5 | 350.5 |
| 4 | 35.0 | 289.6 |
| 5 | 31.6 | 332.0 |
| **Median** | **31.6** | **332.0** |

That is a **10.5x** focused-loop speedup over the actual PR base.

### Whole-application measurement (deterministic headless ladder)

Headless execution is fully deterministic here, which permits a
stronger-than-sampling whole-application comparison: twelve 300M-guest-
instruction runs (five alternating order-flipped timing pairs plus one
trace-profile pair, pristine save state re-seeded per run) produced the
**same final-framebuffer SHA-256 in all twelve runs across both
builds** — byte-identical guest work.

**Result: neutral, and provably so.** The trace-profile pair's JIT
totals are byte-identical between base and this PR — `backward_hits`
624,279, `rejected_hits` 615,734, `native_calls` 47,671, `jit_retired`
11,146,241 — so the two builds executed identical instruction streams
and any timing difference is host environment by construction. The
timing pairs bear that out: user-CPU medians 204.3 s (base) versus
198.0 s (this PR), with per-run spread (164–239 s on identical work)
an order of magnitude larger than the median gap. No whole-application
cost, no whole-application win on this workload.

What the profile pair does show is the admission working and the next
obstacles, precisely:

- Three heads advance past brief-indexed LEA blockers. `00FB1BD8`
  (36,864 hits) now records through its `41F3` and blocks two ops later
  at `9190` (`SUB.L D0,(A0)`); `00FAA45C` and `00FAB518` pass their
  `41F0` and block at `0450` (`SUBI.W #imm,(An)`). Memory-destination
  SUB forms are therefore the natural next admissions, with this LEA
  already paid for.
- A live illustration of #79's ranking-inflation mechanism:
  `00FB1BD8`'s projected dispatches rose 147,448 → 221,172 purely
  because its admitted prefix lengthened.
- The `00FA9732` boot head (the motivating target, 27,292 hits in this
  workload) does **not** compile under this PR. Recording now proceeds
  nine operations past the head — through both LEAs, the
  `ADDQ.L #1,(A0)` increment, and a `SUBQ.W #4,SP` — and then stops at
  `00FA974A`, where **a trap or exception surfaces inside the loop
  body**. Traps are hard trace boundaries, so no instruction-set
  coverage extends a recording past that point: this head cannot
  compile as a loop, and its 2.81M projected dispatches were never
  achievable. That is an instance of the ranking-inflation family #79
  describes, and it means this PR's honest claim for that head is "one
  obstacle of at least two removed", not "unblocked".

  Provenance, since it matters for how much weight to give that: the
  stop cause is **measured**, not inferred. A first pass established
  only that *some* recording-stop path fired there; the profiler could
  not say which, because recordings that end at a trap or in a compile
  gate were attributed nowhere. That gap is filed as #84, whose
  implementation distinguishes a trap or exception from a host batch
  boundary, and a re-run of this same workload with it reports
  `trap-or-exception` at `00FA974A` with the nine-operation prefix
  intact. Workload-wide the class is overwhelmingly structural: 41
  trap-or-exception against 1 host-boundary. The remaining inference is
  only the trap's *identity* — the reserve-then-call stack shape
  (`SUBQ.W #4,SP`, trap, `MOVE.L (SP)+`, compare) is the long-returning
  Toolbox A-line ABI — but the profiler classifies the family, not the
  vector, so that specific reading is not claimed as measured.

The capped (`--cpu-mhz 25`) GUI ladder remains the canonical
interactive-scene methodology; this deterministic headless ladder is
the whole-application evidence a boot-phase-targeted admission can
collect without confounds.

## Validation

- decode, portable/native parity (including the negative word-index
  case), condition codes untouched, base register untouched

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1104,6 +1104,69 @@ fn bench_pea_displacement_loop() {
     );
 }
 
+/// Measure a JIT-compiled loop whose address computation is an indexed
+/// LEA — the register-only form that blocks a deterministic boot-phase
+/// head in EV Override profiling. The loop sweeps the word index and
+/// wraps it, so the run ends at an exact wrap boundary with fully
+/// asserted final state:
+///
+/// ```c
+/// for (;;) {
+///     entry = &table[i];        /* LEA (4,A0,D2.W),A1 */
+///     if (++i < 0x7FFF) continue; /* ADDQ; CMPI; BLT   */
+///     i = 0;                    /* MOVEQ #0,D2         */
+/// }
+/// ```
+fn bench_indexed_lea_loop() {
+    const CODE_BASE: u32 = 0x6E00;
+    const WRAP_INSTRS: u32 = 4 * 0x7FFF + 2;
+    const WRAPS: u32 = 763;
+    const INSTRS: u32 = WRAP_INSTRS * WRAPS;
+    let words = [
+        0x43F0, 0x2004, // loop: LEA (4,A0,D2.W),A1
+        0x5282, // ADDQ.L #1,D2
+        0x0C42, 0x7FFF, // CMPI.W #$7FFF,D2
+        0x6DF4, // BLT.S loop
+        0x7400, // MOVEQ #0,D2
+        0x60F0, // BRA.S loop
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(0, 0x4200);
+        cpu.set_a(7, 0x8000);
+        cpu
+    };
+
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.d(2), 0, "the run ends exactly at a wrap boundary");
+    assert_eq!(cpu.a(0), 0x4200, "the base register is untouched");
+    assert_eq!(
+        cpu.a(1),
+        0x4200 + 0x8002,
+        "the last computed address is the final table entry"
+    );
+    println!(
+        "batch     indexed lea loop        {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 /// Measure decoded generic memory operations without allowing a backward
 /// branch to turn the workload into a native JIT loop. Each pass walks the
 /// same straight-line code, retaining the decoded-op cache while resetting
@@ -1231,6 +1294,10 @@ fn main() {
     }
     if only.as_deref() == Some("pea-displacement") {
         bench_pea_displacement_loop();
+        return;
+    }
+    if only.as_deref() == Some("indexed-lea") {
+        bench_indexed_lea_loop();
         return;
     }
     if only.as_deref() == Some("immediate-shifts") {
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -253,6 +253,14 @@ pub(crate) enum JitTraceOp {
         displacement: i16,
         cycles: i32,
     },
+    /// `LEA (d8,An,Xn),An` with the brief extension decoded once while
+    /// recording. LEA is register-only: it accesses no memory and changes
+    /// no condition codes.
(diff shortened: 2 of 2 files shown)


rlanday:
Draft → ready, with whole-application evidence now in the body, collected on a deterministic headless harness (300M-instruction budget; no interactive sessions, so fully reproducible):

- Twelve runs across both builds produced the **same final-framebuffer SHA-256 every time** — byte-identical guest work — and the trace-profile pair's JIT totals are byte-identical between base and PR. Identical executed streams means **whole-application neutral, provably**, with the timing pairs consistent with that (user-CPU medians 204.3 s → 198.0 s inside a host-noise band an order of magnitude wider than the gap).
- The admission itself works: three heads now record past brief-indexed LEAs; their next blockers are `9190` (`SUB.L D0,(A0)`) and `0450` (`SUBI.W #imm,(An)`) — natural follow-up admissions.
- The motivating boot head turned out to contain **an A-line trap mid-loop** (instrumented-run proof in the body): it can never compile as a loop, so its 2.81M projected dispatches were structurally unachievable — a concrete instance of #79's ranking-inflation concern — and the profiler's silence about such recording ends is now filed as #84.

Net: a zero-cost EA-form admission (provably neutral here, 10.5× focused) that advances the recording chains above, plus an investigation that surfaced a profiler reporting gap (#84) and corrected the profile's top-opportunity ranking.

rlanday:
Correcting the provenance of one claim in the body above, and reporting the measurement that settles it.

The body previously stated that the `00FA9732` head's recording dies at an A-line trap, citing an instrumented run. That run did not actually establish it: it showed only that *some* recording-stop path fired at that address, and six of the eight such paths are host batch boundaries (budget exhausted, watched PC, stopped CPU, decoded fast-path miss) rather than traps. "Trap" was a reasonable reading of the surrounding stack shape, but it was a reading.

It has now been measured. #84's implementation distinguishes a trap or exception from a host boundary, and re-running the same deterministic 300M-instruction workload with it reports:

```
   6  00FA9732        1        9   00FA974A trap-or-exception
      prefix: 00FA9732:2003 00FA9734:E588 00FA9736:41EE/FFD8 00FA973A:41F0/0800
              00FA973E:5290 00FA9740:2004 00FA9742:5480 00FA9744:2D40/FFC0 00FA9748:594F
```

So the conclusion stands, on evidence this time. Workload-wide the class is overwhelmingly structural rather than batch-slicing noise: **41 trap-or-exception against 1 host-boundary**.

Two refinements worth stating precisely:

- The classifier reports the *family*, not the vector — A-line, F-line, `TRAP #n`, `BKPT`, illegal instruction, and bus/address errors all land in it. The specific reading of this one as a long-returning Toolbox A-line call still rests on the `SUBQ.W #4,SP` / trap / `MOVE.L (SP)+` / compare shape, so the body no longer presents that part as measured.
- The "can never compile as a loop" claim is stronger than it first appeared: the nine recorded operations are byte-contiguous (`9732`→`974A`, no gaps), so no branch was taken in the prefix and the trap is reached unconditionally from the head on every pass. Had it been on a conditional path, some other iteration could have recorded a compilable shape.

None of this changes the PR's own result — whole-application neutral with byte-identical JIT totals, and three other heads advancing past their brief-indexed LEAs — but it does change how much weight the motivating profile entry deserved, which seemed worth correcting in place rather than quietly editing.

PR #88, opened 2026-08-08
perf: jit immediate word multiplies

rlanday:
## Summary

Admit `MULU.W`/`MULS.W #imm,Dn`. The register-source forms already decode
through `DecodedSimpleOp`, which sees only the opcode word and so cannot
reach the extension word holding the multiplicand; this decodes the
immediate form in the trace decoder, where the operand is available. The
multiplicand is fixed, so its sign extension and — on the 68000 — its
variable cycle contribution both fold at compile time.

Applies to `master` at 0.8.3.

## Why this form, and why the ranking missed it

In a capped (`--cpu-mhz 25`) gameplay profile this form blocked **nine
separate hot loop heads carrying 1,686,698 rejected recordings — the
largest blocker cluster by hits — every one after a single operation.**

A one-operation prefix is why `projected_dispatches` understated it: that
metric is `rejected_hits × prefix_ops`, which counts only the prefix
recorded *before* the blocker. It is a lower bound, and a weak one when
the blocker sits at the head: a loop whose body is entirely behind the
blocker scores near zero however large it is.

## Whole-application evidence (capped GUI session)

One capped gameplay session on this branch, trace-profile report at
window close. Verdict on the nine previously blocked heads, by PC:

**Three compiled, with real retirement:**

| head | ops | calls | instructions retired |
|---|---:|---:|---:|
| `00EEFE56` | 9 | 56,420 | 2,336,493 |
| `00EEC55E` | 22 | 154,430 | 1,767,462 |
| `00F098FC` | 15 | 159,987 | 1,530,648 |

**5.63M instructions ≈ 1.0% of the session's 572M** moved from decoded
execution into traces — the acceptance test (`native_calls`/`jit_retired`
moving at the targeted heads) passing for the first time in this series
of admissions.

**Six advanced but re-blocked** at the next unadmitted operation:
`42A7` (`CLR.L -(SP)`), `4270` ×2 (`CLR.W (d8,An,Xn)`, an indexed-destination clear), `4850` ×2
(`PEA (An)`), `B679` (`CMP.W abs.L`), with recorded prefixes growing
from 1 to 8–22 operations.

**Honest magnitude:** ~1% of instructions at roughly the trace-vs-decoded
per-instruction ratio is a fraction of a percent of session CPU — real
and demonstrated, not dramatic. Two things make it worth taking anyway:
the yield is *proven at head level* rather than projected, and the
revealed chain links are store-class forms whose admission is newly safe
under #81's wait classifier (a native-spin regression of the #76 kind can
no longer form), so the remaining two-thirds of the cluster is reachable
by follow-ups with a safety story.

## Correctness

- Decode admits exactly op_modes 3/7 with the immediate effective
  address; no aliasing with `ABCD`/`EXG`/AND-immediate-source, and the
  register-form path cannot claim it (`ea_mode == 0` there).
- Flags via the shared logic-flags helper: N/Z from the 32-bit result,
  V/C cleared, X untouched; a word product cannot overflow 32 bits.
- The 68000 variable-timing formula mirrors the register form with the
  immediate substituted; the native emitter folds it at compile time, and
  the parity test packs cycles so the folded timing is verified against
  the portable path (both signs, `0x8000`/`0x8001` boundaries, `0xFFFF`).
- `max_cycles()` mirrors the register form exactly: 70 for the 68000
  (whose multiply time varies with the multiplier's bits, worst case
  `38 + 2*16`), 42 for later CPUs. An earlier revision of this PR
  reported a flat 42 and claimed the register form did the same — that
  claim was wrong (the register form has always special-cased the 68000
  to 70), and a headroom regression now checks the metadata against the
  portable-returned cycles across immediates including both worst cases
  (`MULU #$FFFF`, `MULS #$5555`), so a shared underestimate can no
  longer hide behind native/portable parity.
- `recording_failure_diagnostics_add_no_bus_reads` used `C0FC` as its
  example of an unsupported opcode; it now decodes, so the test moved to
  `4AFC` with a comment telling the next admitter to keep the invariant
  (a failing decode costs exactly one bus read), which is the point of
  that test rather than the specific opcode.

## Validation

- `cargo test --all-features` (839) and `cargo test` — full suite with
  both fixture submodules initialized
- decode + portable and native/portable parity tests as above
  --no-default-features`

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -224,6 +224,14 @@ pub(crate) enum JitTraceOp {
         signed: bool,
         m68000_timing: bool,
     },
+    /// `MULU.W`/`MULS.W #imm,Dn`. The multiplicand is an extension word
+    /// rather than a register, so it is captured once while recording.
+    MulWordImmediate {
+        immediate: u16,
+        dst: u8,
+        signed: bool,
+        m68000_timing: bool,
+    },
     /// 68020+ MULU.L/MULS.L with a data-register source and the 32-bit
     /// result form. The 64-bit Dh:Dl form retains the interpreter path.
     MulLongDataReg {
@@ -1844,6 +1852,11 @@ impl JitTraceOp {
                 ..
             } => 70,
             Self::MulWordDataReg { .. } => 42,
+            Self::MulWordImmediate {
+                m68000_timing: true,
+                ..
+            } => 74,
+            Self::MulWordImmediate { .. } => 42,
             Self::MulLongDataReg { .. } => 40,
             Self::AddrDataReg {
                 op: JitAddrOp::Cmpa,
@@ -2041,6 +2054,9 @@ fn decode_trace_op<B: AddressBus>(
     if let Some(op) = decode_binary_immediate_data_reg_trace_op(cpu, bus, pc, opcode, cpu_type) {
         return Some(op);
     }
+    if let Some(op) = decode_mul_word_immediate_trace_op(cpu, bus, pc, opcode, cpu_type) {
+        return Some(op);
+    }
     if let Some(op) = decode_cmpi_word_mem_trace_op(cpu, bus, pc, opcode, cpu_type) {
         return Some(op);
     }
@@ -2159,6 +2175,43 @@ fn decode_binary_immediate_data_reg_trace_op<B: AddressBus>(
             dst,
             size,
             cycles,
+        },
+    })
+}
+
+/// `MULU.W`/`MULS.W #imm,Dn`. The register-source forms already decode
+/// through `DecodedSimpleOp`, which cannot see the extension word holding
+/// the multiplicand; capture it here instead.
+///
+/// This form heads nine separate hot loops in a capped gameplay profile
+/// (1.69M rejected recordings), each blocking after a single operation,
+/// which is what kept the loop bodies behind it out of every trace.
+fn decode_mul_word_immediate_trace_op<B: AddressBus>(
+    cpu: &CpuCore,
+    bus: &mut B,
+    pc: u32,
+    opcode: u16,
+    cpu_type: CpuType,
+) -> Option<TraceBuildOp> {
+    // 1100 ddd 0mm 111 100: opmode 3 = MULU.W, 7 = MULS.W; ea = immediate.
+    if (opcode & 0xF000) != 0xC000 || (opcode & 0x003F) != 0x003C {
+        return None;
+    }
+    let op_mode = (opcode >> 6) & 7;
+    if !matches!(op_mode, 3 | 7) {
+        return None;
+    }
+    let extension = bus.try_read_word(cpu.address(pc.wrapping_add(2))).ok()?;
+    Some(TraceBuildOp {
+        opcode,
+        extension: Some(extension),
+        extension2: None,
+        pc,
+        op: JitTraceOp::MulWordImmediate {
+            immediate: extension,
+            dst: ((opcode >> 9) & 7) as u8,
+            signed: op_mode == 7,
+            m68000_timing: cpu_type == CpuType::M68000,
         },
     })
 }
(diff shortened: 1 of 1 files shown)


benletchford (changes requested):
The 68000 cycle headroom for the new operation is understated. `MulWordImmediate { m68000_timing: true, .. }` falls through to `max_cycles() == 42`, but the emitted/portable operation returns `38 + 2 * variable`, up to 70 cycles. Unlike the PR body states, the existing register-source variant already special-cases the 68000 to 70. `try_execute` relies on this value both to decide whether one trace fits `cycles_remaining` and to derive the self-loop iteration count, so a high-popcount immediate can execute past the caller’s cycle budget or batch too many iterations. Please mirror the existing 70-cycle arm for the immediate form and add a cycle-budget/headroom regression that reaches a worst-case immediate rather than only comparing native and portable implementations that share the same underestimated metadata.

rlanday:
Fixed, and you caught a factual error, not just a bug: the PR body claimed the register form shared the flat-42 convention, and it does not — `MulWordDataReg` has special-cased `m68000_timing` to 70 all along, four lines above the arm I added. The claim survived an explicit review pass here, which is worth being honest about: the "verification" grepped the `=> 42` arm without reading the preceding one. The body is corrected to say what actually happened.

The immediate form now mirrors the register form exactly: 70 for the 68000, 42 for later CPUs.

Your point about parity tests deserved its own regression, and that is what I added: `mul_word_immediate_cycle_headroom_covers_the_68000_worst_case` checks `max_cycles()` against the **portable-returned cycles** across immediates including both worst cases (`MULU #$FFFF` and `MULS #$5555`, each 38 + 2×16 = 70), asserts the sweep actually reaches 70, and pins both arms' values. Native/portable parity cannot catch a shared underestimate — both sides agree while overrunning the budget — so the regression compares metadata against actual cost instead. Mutation-tested: reverting the 70 arm fails it with `headroom: max_cycles 42 < actual 54` at the first high-popcount immediate.

Full gates rerun on the head: fmt, clippy `-D warnings`, wasm check, full suites.

benletchford (changes requested):
The current head still undercounts M68000 immediate multiply by four cycles. The precise `exec_mulu`/`exec_muls` paths return `38 + 2 * variable + ea_source_cycles(AddressingMode::Immediate, Size::Word)`, and that EA term is 4; the new portable/native operations return only `38 + 2 * variable`, with `max_cycles()` capped at 70. Thus the immediate form should span 42..=74 cycles on M68000, but the trace reports 38..=70, so `try_execute` can still overrun its cycle budget or batch too many loop iterations. The new headroom test compares the bound only with the same undercounted portable implementation, so it cannot expose this mismatch. Please include the four-cycle immediate EA cost in both execution paths, raise the M68000 maximum to 74, and add an interpreter differential/headroom regression that reaches a worst-case immediate.

rlanday:
Fixed, rebased onto 0.9.0, and your criticism of the regression itself was the sharpest part — the headroom test compared the bound against the same undercounted implementation, so it was self-consistent and blind to exactly this class of bug. It has company now.

**The fix:** both execution paths charge `42 + 2 × variable` on the 68000 — the register form's `38 + 2 × variable` plus the 4-cycle word-immediate operand fetch, matching `exec_mulu`/`exec_muls`'s `ea_source_cycles(Immediate, Word)` term precisely. `max_cycles()` for the 68000 arm is raised to 74. The 68020+ arm stays 42, which matches the interpreter's fixed pre-scaled value for that tier (no EA term there).

**The regression you asked for:** `mul_word_immediate_cycles_match_the_interpreter` executes the real instruction through `exec_mulu`/`exec_muls` with the immediate in the instruction stream and asserts the trace operation's cycle result equals the interpreter's, for both signed forms across a sweep including both worst cases (`MULU #$FFFF`, `MULS #$5555` — now 74 each), plus result equality and `max_cycles()` dominating the interpreter's figure. Mutation-tested: reverting the EA term fails it at the first immediate with the 4-cycle gap. The old self-referential headroom test is retained (updated to 74) since it still guards metadata-vs-portable drift, but the differential is the load-bearing one.

Full gates on the rebased head: fmt, both suites, clippy `-D warnings`, wasm check.

benletchford (approved):
The latest head resolves the M68000 timing blocker: the trace now includes the 4-cycle immediate EA fetch, uses a 74-cycle headroom bound, and compares against the interpreter at worst-case operands. The complete local gate passes on f7fc503, including default/all-feature tests and package verification.

PR #89, opened 2026-08-08
feat(trace-profile): report silent-rejection opcodes

rlanday:
## Summary

A silent rejection records where a recording ended; this adds *what
stopped it* — the opcode word at that address, read through the fastmem
window only, printed as a new column in the silent-rejections table and
carried on `SilentRejectionProfileRow`.

## Why the address alone was not enough

For a `trap-or-exception` rejection, the address does not say which trap
was involved — and that is the question that decides what to do about the
head. An accessor trap like `_TickCount` is a plausible candidate for
ever compiling through; a Memory Manager call that may relocate handles,
or a segment-loader call that rewrites code, is not. Without the opcode,
"trap-bounded" is a family verdict with no individual identity.

## Demonstrated use

Two decisions in this project's measurement log were settled with this
instrumentation:

- **Compiling through traps was abandoned on its evidence.** Across a
  900M-instruction run, recordings stranded 155 operations total at trap
  boundaries, and the distribution named the traps: the mass sits on
  `A9F0` (`_LoadSeg`) and `A11E` (`_NewPtr`) — exactly the calls that
  relocate code and memory and cannot be compiled through — while
  `A975` (`_TickCount`), the one safely-whitelistable trap, stranded 11
  operations. Neither a whitelist nor a general implementation could
  recover anything meaningful, and that conclusion required knowing the
  opcodes, not just the addresses.
- **A published claim was verified rather than inferred**: the
  `00FA9732` head's recording was confirmed to end at a trap
  (`trap-or-exception` at `00FA974A` with its nine-operation prefix
  intact), replacing an earlier stack-shape inference.

## Notes

- Window-only read: the profiler adds no bus traffic, preserving the
  invariant #73 established and its `CountingBus` test enforces.
- For `host-boundary` rejections the captured word is the next
  unexecuted instruction, per the existing `exit_pc` semantics.
- The word joins the dedup key, so distinct stopping opcodes at one
  address stay distinct; the 4096-shape cap and overflow counter are
  unchanged.

## Validation

- `cargo test --all-features` (837) and `cargo test` — full suite with
  both fixture submodules initialized; the silent-rejection tests assert
  the new column and field
  --no-default-features`

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -1104,10 +1104,16 @@ impl TraceJit {
                     RecordingEnd::Stopped(RecordingStop::TrapOrException) => cpu.ppc,
                     _ => exit_pc,
                 };
+                // Name the instruction that stopped the recording. For a trap
+                // this is the A-line word identifying the Toolbox or OS call,
+                // which decides whether a trace could ever be compiled through
+                // it. Window-only read: the profiler must not add bus traffic.
+                let reported_opcode = super::mem_ops::peek_window_word(cpu, reported_pc);
                 super::trace_profile::note_silent_rejection(
                     start_pc,
                     cpu_type,
                     reported_pc,
+                    reported_opcode,
                     recorded_shape,
                     reason,
                 );
--- a/src/core/trace_profile.rs
+++ b/src/core/trace_profile.rs
@@ -193,6 +193,14 @@ pub struct SilentRejectionProfileRow {
     /// the PC the CPU had reached otherwise. For a trap this is the bound
     /// on how far this head can ever record.
     pub exit_pc: u32,
+    /// Opcode word at `exit_pc`, read through the fastmem window only.
+    ///
+    /// For [`TraceRejectReason::TrapOrException`] this names *which* trap
+    /// bounded the recording, which the address alone does not: an A-line
+    /// word ($Axxx) identifies the Toolbox or OS call. Without it the report
+    /// can say a head is trap-bounded but not whether the trap is one a
+    /// trace could ever be compiled through.
+    pub exit_opcode: Option<u16>,
     /// Operations successfully reconstructed before the recording ended, in
     /// execution order.
     pub prefix: Vec<TraceShapeOp>,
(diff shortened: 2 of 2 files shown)


benletchford (changes requested):
Two concrete blockers remain. Both required-check failures come from the non-Conventional `trace-profile: ...` subject: the PR title fails `Validate PR title`, and the sole commit fails Rust CI’s commit-message step. Please give both an allowed type/scope subject (for example `feat(trace-profile): report silent-rejection opcodes`); changing only the PR title will not fix commit validation. Also, the tests inject `exit_opcode` directly into `note_silent_rejection`, so they do not exercise the new `finish_recording()` capture path or prove that trap `ppc` rather than the post-trap `exit_pc` is read. Extend the existing `trap_inside_a_loop_body_is_attributed_not_silently_dropped` end-to-end regression to assert `exit_opcode == Some(0xA123)` (and retain the window-only/no-extra-bus-read invariant). This is the behavior the new public field promises.

rlanday:
Both addressed.

**CI**: the commit is reworded and the PR retitled to `feat(trace-profile): report silent-rejection opcodes`, so both the title check and the commit-message step validate.

**The capture-path gap**: you are right that injecting `exit_opcode` into `note_silent_rejection` proved nothing about `finish_recording` — the field's promise is precisely that the *capture site* reads the trapping instruction's word via `ppc`, not the post-trap `exit_pc`. The end-to-end regression `trap_inside_a_loop_body_is_attributed_not_silently_dropped` now asserts `exit_opcode == Some(0xA123)`: the value can only be the A-line word if the capture read `ppc` (the post-trap address holds a different word), and it can only be `Some` at all if the window-only `peek_window_word` path ran. The no-extra-bus-read invariant stays covered by the existing `CountingBus` regression, which passes unchanged.

benletchford (approved):
The updated head resolves both prior blockers. The public field is documented on a non-exhaustive row, trap attribution now exercises the real `finish_recording` capture through `ppc`, and the window-only path preserves the no-extra-bus-read invariant. Formatting, Clippy, default and all-feature tests, rustdoc with warnings denied, and package verification all pass locally on `2c70fce`.

PR #95, opened 2026-08-09
perf: jit register-to-memory subtracts

rlanday:
## Summary

Extend `AddRegToMem` to SUB and to the plain indirect destination:
`SUB.W/L Dn,(An)`, `Dn,(An)+`, and `Dn,(d16,An)` (and `ADD` to `(An)`,
newly admitted alongside). The subtract path reuses the `set_sub_flags`
lowering `MemAddqSubq` already uses; the indirect destination reuses the
established `JitEa::Ind` handling. Applies to current `master` (0.10.0).

## Why this form

`9190` (`SUB.L D0,(A0)`) blocked `00FB1BD8` — rank 6 in a capped
gameplay profile at 151,552 hits / 909,306 projected dispatches, and
present in the deterministic headless workload at 36,864 hits — after
the #83 LEA admission advanced that head's chain to it.

## Whole-application acceptance (deterministic headless A/B, originally vs 0.9.1; reproduced exactly on current master)

The acceptance test this series uses — `native_calls`/`jit_retired`
moving at the targeted head, not merely the blocker advancing — passes:

| | base | with this change |
|---|---|---|
| `00FB1BD8` | blocked at `9190`, prefix 6 | **compiled: 24 ops, 254 calls, 882,168 instructions retired, 3,473 ops/call, zero guard exits** |
| totals | `native_calls` 47,671; `jit_retired` 11,146,241 | 47,925; **12,028,409 (+7.9%)** |

The aggregate delta is exactly this head. Zero guard exits at 3,473
operations per call makes it the cleanest-executing admission of the
series. (n=1 per arm; sufficient because the verdict is binary and this
harness is deterministic — the base totals reproduce byte-identically
across all its prior runs. Re-run on current master inside a larger
bundle, this head's figures reproduced exactly: 24 ops, 254 calls,
882,168 retired.)

**Gameplay yield, measured** (capped `--cpu-mhz 25` session, 847M guest
instructions, 2026-08-08): the same head compiled with 24 ops,
961 calls, **3,628,800 instructions retired**, 3,776 ops/call, zero
guard exits. The #88 protocol's gameplay check passes.

## Focused benchmark

`benches/microbench.rs sub-reg-to-mem`: four `SUB.L D0,(A0)` between a
MOVEA reload and a DBRA, sized to end exactly at the outer label so the
final cell value proves every subtract landed. Five alternating runs
against base carrying an identical copy of the benchmark: median
29.7 → 177.6 M instr/s (**6.0×**); per-pair ratios 5.4–6.0 under
visible thermal drift, which the alternation absorbs.

## Correctness

- Flag cases where an ADD-path mistake would show are covered by parity
  tests: borrow (C/X), signed overflow (V), zero (Z), across
  portable/native with memory, CCR-including-X, and cycle equality.
- Decode covers all three destination forms plus the ADD-indirect form,
  and asserts no extension word is consumed for `(An)`.
- Cycle metadata follows the existing `AddRegToMem` table; the `(An)`
  destination takes the existing non-displacement arms (20 long /
  12 word).

## Validation

  --no-default-features`

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -755,6 +755,61 @@ fn bench_movem_indexed_loop() {
     );
 }
 
+/// Exercise the register-to-memory subtract loop the profile flagged: four
+/// `SUB.L D0,(A0)` against one cell between a MOVEA reload and a DBRA, so
+/// the admitted form dominates the trace rather than the loop plumbing.
+fn bench_sub_reg_to_mem_loop() {
+    // 326,594 outer cycles of 643 instructions end exactly at the outer
+    // label, with 512 subtracts per cycle applied to the same cell.
+    const CYCLES: u32 = 326_594;
+    const INSTRS: u32 = CYCLES * 643;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x204B, // outer: MOVEA.L A3,A0
+        0x727F, // MOVEQ #127,D1
+        0x9190, // inner: SUB.L D0,(A0)
+        0x9190, // SUB.L D0,(A0)
+        0x9190, // SUB.L D0,(A0)
+        0x9190, // SUB.L D0,(A0)
+        0x51C9, 0xFFF6, // DBRA D1,inner
+        0x60EE, // BRA.S outer
+    ];
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_d(0, 1);
+        cpu.set_a(3, 0x4000);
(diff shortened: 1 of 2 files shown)


benletchford (changes requested):
The implementation and local test gate pass, but commit 76688aa uses `bench:`, which is not one of the repository’s allowed Conventional Commit types. That is why “Check commit messages” fails. Please amend it to an allowed single-line type (for example, `perf: add focused register-to-memory subtract benchmark`) and rerun CI.

benletchford:
After #96 landed, this branch also reports merge conflicts in the overlapping JIT changes. Please rebase it onto current `master` while amending the commit subject, resolve the overlap in the introducing commit, and rerun the full checks.

rlanday:
Both addressed: rebased onto current `master` (0.10.1) — the JIT overlap with #96 resolved in the introducing commits (the feature commit applied cleanly; the benchmark's `main()` dispatch guard was the only conflict, unioned with the shift bench's) — and the benchmark commit subject amended to the allowed type (`perf: add focused register-to-memory subtract benchmark`). Full gates rerun on the rebased branch: 190 lib tests, `cargo fmt --all -- --check` clean, clippy clean.

benletchford (approved):

PR #97, opened 2026-08-09
perf: jit stores to brief-indexed destinations (MOVE, CLR)

rlanday:
## Summary

Lift the decoder's indexed-destination gate for MOVE and add a `ClrMem`
trace op for `CLR.B/W/L (d8,An,Xn)`. The decoder comment gated indexed
stores on "a profile demonstrates that their extra emitter paths pay";
three profiled heads across two workloads now terminate on exactly this
capability, so the store side gets the same brief-indexed support the
read side gained across the indexed-read series (guarded scans,
memory-source ALU, immediate compares, LEA). Applies to `master` at `5ecdc72`.

## Why these forms

Three independent chain heads converge on indexed-destination stores —
the only shared missing capability across both measured workloads:

| opcode | form | head it blocks | workload |
|---|---|---|---|
| `3180` | `MOVE.W Dn,(d8,An,Xn)` | `00FB1C4A` (next link after the shift admission) | headless |
| `4230` | `CLR.B (d8,An,Xn)` | `00EE9DBA` (multiply-chain head) | headless |
| `4270` | `CLR.W (d8,An,Xn)` | gameplay head (capped session 2026-08-08) | gameplay |

One address computation (brief extension: base + sign-extended or long
index, optional scale, d8) plus the established checked-store tail
(window check → store-not-code guard → staged commit → store → flags)
serves all three.

## Whole-application acceptance (deterministic headless A/B vs `5ecdc72`)

Measured in the four-feature bundle (sub + shifts + pea + this) against
`5ecdc72`; the base totals reproduce the deterministic byte-identical
baseline of every prior run.

| | base | with bundle |
|---|---|---|
| `00FB1C4A` (jointly with #96) | blocked at `E2A0`, prefix 4, chain shows `3180` next | **compiled: 23 ops, 313 calls, 879,405 instructions retired, 2,810 ops/call, zero guard exits** |
| `00FB1BD8` | blocked at `9190`, prefix 6 | **compiled: 24 ops, 254 calls, 882,168 retired, 3,473 ops/call, zero guard exits** (reproduces the solo-SUB acceptance exactly) |
| `00EE9DBA` | blocked at `4230`, prefix 25 | `4230` admitted; advanced to `31BC` (`MOVE.W #imm,(d8,An,Xn)`), prefix 30, residual 35 hits / 1,020 projected — headless residue is negligible |
| totals | `native_calls` 47,671; `jit_retired` 11,146,241 | 65,322; **15,986,450 (+43.4%)** |

The two named heads account for 1.76M of the +4.84M retired delta; the
remaining ~3.1M is other heads the shared indexed-store path unblocked —
the convergence hypothesis paying beyond its named targets. Backward
hits into the recorder fell 869K → 582K (loops now spend those
iterations inside native traces).

## Gameplay (capped `--cpu-mhz 25` session, 847M guest instructions)

- `00FB1C4A` (#96's shifts + this `MOVE.W` form) compiled end-to-end in play:
  23 ops, 1,283 calls, **3,505,568 retired, zero guard exits**.
- Neither `4230` nor `4270` blocks anything ranked — the CLR forms are
  consumed into longer recordings.
- Session-wide with the four-feature bundle: `jit_retired` 424,277,987
  = **50.1% of every instruction the session executed**.
- The next link this admission revealed (`31BC`, immediate→indexed
  MOVE) projects only 244K dispatches in gameplay and 1K headless —
  documented, not chased.

## Focused benchmark

`benches/microbench.rs indexed-dest-store`: `MOVE.W D0,(0,A0,D1.W*2)`
plus `CLR.B (1,A0,D1.W*2)` swept by the DBRA counter, so every
iteration stores through a different address and the final memory
pattern proves both stores executed across the sweep. Five alternating
runs against base carrying an identical copy of the benchmark: median
24.9 → 225.1 M instr/s (**9.0×**); per-pair ratios 7.4–9.1.

## Correctness

- **The staged-source case is the subtle one**: in
  `MOVE.W (A2)+,(d8,A1,A2.W)` the source post-increment commits before
  the destination EA evaluates, so the destination's index-register read
  must observe the staged value. Because portable and native could share
  a wrong staging model, the test checks portable against the
  interpreter's own decoded execution (`execute_mem_op`), which
  demonstrably stores through the post-incremented index. Native parity
  covers the plain, negative-word-index (sign-extension boundary), and
  staged cases; the staged-index mutation (native reading the stale
  register) fails the parity memory comparison.
- **CLR semantics**: Z set, N/V/C cleared, X untouched — asserted
  explicitly; portable delegates to the interpreter's decoded CLR so
  store and flag semantics have one source of truth.
- **Store-into-code bails atomically on both paths**: the checked window
  offset and store-not-code guard precede the store and flag writes;
  tests aim a store at the trace's own extension word and assert nothing
  commits (portable and native). Dropping the portable pre-check fails
  the bail assertion (mutation-tested).
- **Classifier**: `ClrMem` declares non-poll in `is_pure_poll_loop` (a
  store ends a wait shape), keeping #81's exhaustive-match contract.
- Cycle maxima follow the `TstMem` convention: M68000UM store cost plus
  the indexed EA calculation (MOVE unchanged; CLR 18 B/W, 26 long).

## Validation

- 190 lib tests on the branch (196 in the four-feature bundle), clippy
  clean, wasm32 check green, `cargo fmt` clean.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -810,6 +810,64 @@ fn bench_sub_reg_to_mem_loop() {
     );
 }
 
+/// Exercise the brief-indexed store forms the profiles flagged: a MOVE.W
+/// to a scaled-index destination and a CLR.B one byte beside it, swept by
+/// the DBRA counter so every iteration stores through a different address.
+fn bench_indexed_dest_store_loop() {
+    // 542,635 outer cycles of 387 instructions end exactly at the outer
+    // label. Each inner iteration writes D0 to (A0 + 2*D1) and clears the
+    // low byte, so the final words hold D0 with a zeroed low byte.
+    const CYCLES: u32 = 542_635;
+    const INSTRS: u32 = CYCLES * 387;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x204B, // outer: MOVEA.L A3,A0
+        0x727F, // MOVEQ #127,D1
+        0x3180, 0x1200, // inner: MOVE.W D0,(0,A0,D1.W*2)
+        0x4230, 0x1201, // CLR.B (1,A0,D1.W*2)
+        0x51C9, 0xFFF6, // DBRA D1,inner
+        0x60EE, // BRA.S outer
+    ];
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_d(0, 0x0000_BEEF);
+        cpu.set_a(3, 0x4000);
+        cpu.set_a(7, 0x8000);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.pc, CODE_BASE, "the run ends exactly at the outer label");
+    assert_eq!(
+        bus.read_word(0x4000),
+        0xBE00,
+        "the store went through and the clear zeroed its low byte"
+    );
+    assert_eq!(
+        bus.read_word(0x4000 + 254),
+        0xBE00,
+        "the sweep reached D1=127"
+    );
+    println!(
+        "batch     indexed dest stores     {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 /// Reproduce a path-biased trace that is first recorded through an uncommon
 /// conditional edge before settling into a copy loop on the opposite edge.
 /// A first-path-only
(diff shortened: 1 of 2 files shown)


benletchford (changes requested):
Two mechanical blockers remain. `cargo fmt --all -- --check` reformats the long `assert_eq!` at `benches/microbench.rs:805`, and commit 05b35ce uses the unsupported `bench:` type (the commit-message check will fail after formatting passes). Please format the branch and amend that subject to an allowed single-line Conventional Commit type, such as `perf: add focused brief-indexed store benchmark`, then rerun CI.

benletchford:
After #96 landed, this branch also reports merge conflicts in the overlapping JIT changes. Please rebase it onto current `master` while fixing formatting and the commit subject, resolve the overlap in the introducing commits, and rerun the full checks.

rlanday:
All three addressed: rebased onto current `master` (0.10.1) with the #96 overlap resolved in the introducing commits (the two feature commits applied cleanly; the benchmark `main()` dispatch guard was the only conflict), `cargo fmt --all` applied to the long `assert_eq!` inside the benchmark commit itself, and that commit's subject amended to the allowed type (`perf: add focused brief-indexed store benchmark`). Gates on the rebased branch: 192 lib tests, `cargo fmt --all -- --check` clean, clippy clean.

benletchford (approved):

benletchford:
#95 has now landed on `master`, and this head conflicts with it in both `benches/microbench.rs` and `src/core/trace_jit.rs`. Please rebase onto current `master`, preserve both focused benchmark dispatches, and combine the `AddRegToMem` and `ClrMem` enum/match/test additions in their introducing commits, then rerun the full checks. The code at `15e635f` passed review and the complete local gate before the base moved; the rewritten head still needs revalidation.

rlanday:
Rebased onto current `master` after #95 landed; the test-region overlap with the subtract tests is resolved inside the introducing commits (each intermediate commit builds standalone). Gates rerun on the result: 194 lib tests, `cargo fmt --all -- --check` clean, clippy clean, focused bench runs. Should be mergeable again.

Issue #100, opened 2026-08-09
Trace formation, not the operation set, limits gameplay JIT coverage — measurement summary and design question


rlanday:
## The measured picture

Gameplay JIT coverage on capped (`--cpu-mhz 25`) sessions was
**scene-dependent between 8% and 46%** of executed instructions before
the store-family admissions, and reached **50.1%** (424.3M of 847.3M)
with them (2026-08-08 session). Guest execution is ~46% of non-idle
gameplay CPU, and the decoded memory-op path plus bus accessors —
exactly what a formed trace bypasses via fastmem — are ~24% of it.
Lifting the low-coverage half is worth roughly **15–20% of gameplay
CPU**, an order of magnitude beyond anything else measured in this
series.

**The admission frontier is now closing.** After the 2026-08-08 session
the top remaining *admissible* target is the predecrement-CLR family
(22.6M projected dispatches; implemented, pending its session). Behind
it, the ranking is dominated by heads that wall at `61FF` (`BSR.L`):
in the 2026-08-09 session (2.53B instructions), **51.0M projected
dispatches across 19 heads — 47% of the entire ranked opportunity —
sit behind call boundaries no opcode admission can reach**, and two of
the three predec-CLR heads demonstrated the pattern live by consuming
their CLRs and re-walling at `61FF` with zero retirement. Opcode
coverage is no longer the binding constraint; formation is.

## What it is not

- **Not the operation set.** Two independent measurements: the total
  stranded-prefix opportunity across every ranked blocker is ~1% of
  executed instructions; and five sequential op admissions measured
  whole-application neutral, with heads advancing to the next chain link
  rather than compiling (chains of blockers, confirmed end-to-end in
  #88's session: one admission compiled 3 of 9 heads and re-blocked 6).
- **Not trap boundaries.** With #89's trap-word attribution: recordings
  stranded **155 operations total** across a 900M-instruction run, mass
  on `_LoadSeg`/`_NewPtr` — the uncompilable-through traps — while
  `_TickCount` stranded 11 ops. Compiling through traps was prototyped on
  paper and abandoned on this evidence.

## Where the loss concentrates

1. **Traces only ever start at backward-branch targets**
   (`record_trace_target` has a single caller). Straight-line code,
   call-heavy code, and loop bodies whose back edge is consumed by a
   compiled *inner* trace never get candidacy.
2. **Guard-exit thrash**: several 31–37-op traces exit through a guard on
[...]

PR #101, opened 2026-08-09
perf: seed trace candidacy from guard exits and chain compiled continuations

rlanday:
Implements the second proposal of #100 (exit-seeded candidacy); the
other two remain open design questions there.

## Summary

Traces only ever start at backward-branch targets, so a loop whose
conditional branch never settles leaves its off-spine path interpreted
forever: the head trace guard-exits on a fixed fraction of calls, the
adaptive re-record ratio never trips, and the continuation between the
exit and the loop latch is invisible to candidacy — it is neither a
rejected hit nor a blocker, so it doesn't even appear in the
opportunity ranking.

This PR makes guard exits themselves the discovery signal:

- **Seed**: a guard exit counts the exit target as a trace-head
  candidate (same hot threshold as backward targets). Unlike
  backward-target candidacy, an exit seed never evicts a compiled trace
  on a cache-index collision — backward targets are proven loop heads,
  exit targets are speculative.
- **Record**: when the exit target goes hot, the interpreter records
  the continuation from there; the existing close conditions already
  handle a mid-loop head (flow re-reaching the start pc from any
  direction, or the DBcc region boundary).
- **Chain**: once compiled, the parent executes the continuation
  directly from its exit, bounded by a chain budget
  (`TRACE_EXIT_CHAIN_BUDGET = 3`) and by the caller's instruction
  budget, which shrinks by what the parent retired.

## Why (measured motivation)

In the 2026-08-09 capped gameplay session (2.53B guest instructions),
pre-existing compiled traces exit through guards on up to 97.5% of
calls (`00EF8D92`: 122,966 exits / 126,111 calls) with adaptive
re-recording exhausted, while every recently admitted store-family
trace runs at zero exits — the thrash is a property of formation, not
of trace execution. The volume those exits strand is invisible to the
opcode-opportunity ranking, which is exactly why no admission has ever
recovered it.

## The chaining cost, stated plainly

A parent→continuation transition is not seamless native-to-native: the
parent's exit returns to the Rust driver, which performs slot lookup,
budget checks, and the child's own entry validation before the second
native call — the same per-entry boundary every trace already pays,
once per exit, amortized by the continuation's ops-per-call. Direct
side-exit linking (patching the parent's exit stub to jump straight
into the child) would remove that boundary and is left as explicit
future work; this PR is deliberately policy-only.

## Focused benchmark

`benches/microbench.rs mixed-path`: a loop whose branch alternates
every iteration (`EORI #1` flips Z), so the head trace guard-exits on
half of all calls forever — below the adaptive re-record ratio, above
any useful coverage — with a field-shaped 14-op continuation between
the join and the DBRA. Seven alternating runs of the **review-fixed**
code against current master carrying an identical copy: per-pair
ratios 1.78–1.97, **median 1.86×** (median 107.5 → 194.7 M instr/s).

One negative result kept in the record: with a 3-op continuation the
same A/B is flat (ratios scattered around 1.0×) — the chain boundary
costs about what interpreting three cheap ops costs, consistent with
the 2026-08-03 hot-leaf finding. The mechanism pays in proportion to
continuation length, which is what the stranded field continuations
have and short ones don't; direct side-exit linking would recover the
short-continuation case and is left as future work.

- **Headless** (deterministic A/B of the review-fixed code vs current
  master with #95/#96/#97 merged, 900M instructions): `native_calls`
  65,322 → **59,141 (−6,181)** while `jit_retired` rose 15,986,450 →
  **16,050,628 (+64,178)** and ops-per-native-call rose 244.7 → 271.4 —
  chaining consolidates boundary crossings (continuations execute
  inside chained entries instead of separate probes) while retiring
  more. Rejected hits flat. An earlier headless figure measured the
  pre-fix gate and is superseded by this one.
- **Capped gameplay session: PENDING** — the motivating traces
  (guard exits on up to 97.5% of calls) are gameplay-side; the session
  verdict is guard-exit ratios dropping on those existing traces plus
  new compiled heads at mid-loop pcs, observables disjoint from the
  store-admission PRs riding the same session.

## Review fixes (both findings were real)

- **Chained miss surfaced**: a continuation whose first opcode changed
  returned `Miss` with the opcode already consumed; the parent dropped
  it, skipping a guest instruction. The parent now propagates the miss
  with every instruction retired so far, and `run_batch`'s miss arm
  accounts those before dispatching the missed opcode. Regression: a
  three-phase step()-twin (head compiled on the taken spine, predicate
  flipped externally so the continuation forms and chains under the
  adaptive window, continuation head rewritten in both twins) — the
  chain miss fires exactly once, and dropping it diverges D2 by the
  skipped ADDQ.
- **Seeding gated on guarded branch exits**, not any partial call:
  memory bails (window/alignment/SMC) no longer create candidacy at
  an access the trace cannot execute. Negative regression: a loop
  whose store idempotently hits its own code memory-bails every call,
  and the bail target's slot must stay completely untouched — counting,
  compiled, **or rejected** all fail the assert (under the old gate the
  seed left a `Rejected` slot after a wasted record, which a weaker
  assertion misses).

## Correctness

- The mixed-path integration test drives the full pipeline — seed,
  record, compile, chain — through `run_batch` and must match `step()`
  exactly (registers, SR, PC, memory, instruction counts).
- The behavior test asserts a second compiled trace forms inside the
  loop body; setting the chain budget to zero fails it (mutation-
  tested), so the test detects the mechanism, not just exactness.
- The unit test pins the slot rules: promotion at the threshold,
  deferral while another recording is active, rejected slots blocking
  re-seeds, and no eviction of compiled traces by speculative seeds.
- **Watch semantics preserved**: seeding is bookkeeping and always
  safe; a chained entry is not observed by the runner, so chaining
  falls back to the interpreter whenever the chain target is watched.
- Recursion is doubly bounded (chain budget and strictly shrinking
  instruction budget); `single_iter` callers never chain.

## Validation

- 192 lib tests (all-features), the full `run_batch` integration
  suite, clippy clean, wasm32 check green, `cargo fmt` clean.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1094,6 +1094,66 @@ fn bench_salvaged_prefix_loop() {
     );
 }
 
+/// A mixed-path loop whose conditional branch alternates every iteration
+/// (EORI #1 flips Z), so whichever spine the head trace records, it
+/// guard-exits on half of all calls forever -- below the adaptive
+/// re-record ratio, above any useful coverage. The continuation between
+/// the branch join and the DBRA is what exit-seeded candidacy compiles.
+fn bench_mixed_path_loop() {
+    const CODE_BASE: u32 = 0x6000;
+    const INSTRS: u32 = 100_000_000;
+    let words = [
+        0x0A01, 0x0001, // head: EORI.B #1,D1
+        0x6602, // BNE.S join
+        0x5282, // ADDQ.L #1,D2
+        // join: a field-shaped continuation -- long enough that a chained
+        // native call amortizes its entry boundary, as the stranded
+        // continuations in the gameplay profile are.
+        0x1ADC, // join: MOVE.B (A4)+,(A5)+
+        0x5283, // ADDQ.L #1,D3
+        0x1ADC, 0x5283, // x2
+        0x1ADC, 0x5283, // x3
+        0x1ADC, 0x5283, // x4
+        0x1ADC, 0x5283, // x5
+        0x1ADC, 0x5283, // x6
+        0x51C8, 0xFFDE, // DBRA D0,head
+        0x2845, // outer: MOVEA.L D5,A4
+        0x2A46, // MOVEA.L D6,A5
+        0x707F, // MOVEQ #127,D0
+        0x60D4, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(4, 0x4000);
+        cpu.set_a(5, 0x5000);
+        cpu.set_a(7, 0x9000);
+        cpu.set_d(0, 127);
+        cpu.set_d(5, 0x4000);
+        cpu.set_d(6, 0x5000);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    println!(
+        "batch     mixed-path loop         {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 /// Reproduce a path-biased trace that is first recorded through an uncommon
 /// conditional edge before settling into a copy loop on the opposite edge.
 /// A first-path-only
@@ -1729,6 +1789,10 @@ fn main() {
     }
     if only.as_deref() == Some("sub-reg-to-mem") {
         bench_sub_reg_to_mem_loop();
+        return;
+    }
+    if only.as_deref() == Some("mixed-path") {
+        bench_mixed_path_loop();
         return;
     }
     if only.as_deref() == Some("indexed-lea") {
--- a/src/core/op_cache.rs
+++ b/src/core/op_cache.rs
@@ -1062,7 +1062,16 @@ impl CpuCore {
                             probe = true;
                             continue;
                         }
-                        CachedRunResult::Miss(opcode) => return BatchInnerExit::Miss(opcode),
+                        CachedRunResult::Miss(opcode) => {
+                            // A chained continuation can retire instructions
+                            // before its successor misses validation; those
+                            // are real completed instructions and must be
+                            // accounted before the missed opcode dispatches.
+                            // (`remaining` needs no update: this return path
+                            // leaves the budget loop entirely.)
+                            *retired += instructions;
+                            return BatchInnerExit::Miss(opcode);
+                        }
                     }
                 }
             }
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -667,8 +667,13 @@ impl TraceJit {
     /// Attempt to execute a compiled trace at the current PC.
     ///
     /// On `CachedRunResult::Ran`, the returned count is the number of
-    /// guest instructions the trace retired. The count is 0 for
-    /// `Fault`/`Miss`.
+    /// guest instructions the trace retired. On `Miss` the count is the
+    /// number of instructions retired BEFORE the miss: zero when the
+    /// entered trace's own first opcode changed, but non-zero when a
+    /// chained continuation missed validation after its parent (and any
+    /// earlier links) retired instructions -- callers must account the
+    /// count before dispatching the missed opcode, as
+    /// `run_decoded_simple_batch` does. The count is 0 for `Fault`.
     ///
     /// A self-looping trace (one whose closing branch targets its own
     /// head) may run many iterations per call: up to `instr_budget`
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -12702,7 +12702,7 @@ mod portable_tests {
         const CODE_BASE: u32 = 0x7000;
         let words = [
             0x5282, // head: ADDQ.L #1,D2
-            0x3088, // MOVE.W D0,(A0)   (A0 -> head; writes 0x5282 back)
+            0x3080, // MOVE.W D0,(A0)   (A0 -> head; writes 0x5282 back)
             0x51C9, 0xFFFA, // DBRA D1,head
             0x60F6, // BRA.S head
         ];
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -697,6 +697,14 @@ impl TraceJit {
         if cpu.has_pmmu && cpu.pmmu_enabled || cpu.cycles_remaining <= 0 {
             return None;
         }
+        if instr_budget == 0 {
+            // With nothing left to retire, even a validation miss is too
+            // much: consuming a changed opcode would hand run_batch one
+            // instruction past its exact budget. A chained child can be
+            // entered this way when its parent retired the final permitted
+            // instruction; the next entry validates with budget to act.
+            return None;
+        }
         if cpu.trace_recording || self.recording.is_some() {
             // A recorder is already following an executed path on this
             // thread. Nested backward edges and interleaved CPU instances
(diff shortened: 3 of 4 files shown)


rlanday:
Running benchmarks now

benletchford (changes requested):
Two blocking issues need correction before this is safe to validate:

1. The recursive chain call only accepts CachedRunResult::Ran. If validation of the compiled continuation finds that its first opcode changed, try_execute removes the child trace, sets ppc/ir, advances pc past that opcode, and returns CachedRunResult::Miss. The parent silently drops that result and returns Ran, so run_batch resumes after the changed opcode without ever dispatching it. Please either restore the child entry state on a non-Ran chain result or plumb an outcome that preserves the parent retired count while surfacing the miss. Add a run_batch-versus-step regression that compiles a continuation, rewrites its first opcode, then side-exits into it.

2. Exit seeding is gated by partial_call_this_entry, not guarded_branch_exit. Partial calls also include window/alignment/self-modifying-code memory bails, so the current code can seed and repeatedly probe or compile a continuation at an access that the trace cannot execute. Gate discovery/chaining on an actual guarded branch exit and add a negative test showing that a memory bail does not create exit candidacy.

The branch now also conflicts with current master after #97; please rebase and preserve both benchmark dispatches and the overlapping trace-JIT/test additions when addressing these issues.

rlanday:
Both findings were real; both are fixed in the new commit, each with the regression you specified, and the branch is rebased onto current master with the #97 overlaps resolved in the introducing commits.

**1. Chained miss surfaced.** The parent now propagates `CachedRunResult::Miss` together with every instruction retired so far, and `run_batch`'s miss arm accounts those retired instructions before dispatching the missed opcode (previously it discarded the count — harmless while it was always zero, wrong once chaining exists). The regression is a three-phase step()-twin rather than a free-running loop, because in a free-running one the head trace's own spine tends to absorb the rewrite before a stale chain can fire: phase 1 compiles the head on the taken spine (which excludes the continuation), phase 2 flips the predicate externally so the continuation seeds, compiles, and chains — kept under the 64-call adaptive window so re-recording can't replace the spine — and phase 3 rewrites the continuation's first opcode (`ADDQ.L #1,D2` → `#2`) in both twins. A probe confirmed the chain-validation miss fires exactly once in phase 3; with the fix reverted the test fails on D2 diverging by the skipped ADDQ.

**2. Seeding gated on guarded branch exits.** Discovery and chaining now key on `guarded_branch_exit`, not `partial_call`. The negative test runs a loop whose store idempotently hits its own code — a store-not-code bail at op 1 on every call — and asserts the bail target's slot stays *completely untouched*: counting, compiled, **or rejected** all fail it. The rejected case is the one a weaker assertion misses: under the old gate the seed recorded a too-short continuation and left a `Rejected` slot, which is exactly the wasted record/compile churn you described.

Numbers rerun on exactly this code (the previously posted headless A/B predated the gate fix and is superseded): focused mixed-path bench and the deterministic headless A/B are being refreshed and the body will carry the new figures. Full gates on the rebased branch: 198 lib tests (including the two new regressions), the run_batch integration suite with the SMC-continuation exactness test, clippy clean, wasm check green.

benletchford (changes requested):
The two earlier correctness findings are addressed on this head, including the chained validation miss regression and excluding memory bails from exit seeding. Two new cleanup blockers remain before re-review:

1. `benches/microbench.rs` now has two identical `if only.as_deref() == Some("mixed-path")` dispatch blocks in `main`; remove the duplicate when resolving the benchmark overlap.
2. The `try_execute` contract still says the returned count is zero for `Miss`, but a chained continuation miss can now return the parent/deeper retired count and `run_decoded_simple_batch` relies on that. Update that documentation so future callers do not discard or mis-handle the count.

The branch also conflicts with current master after #103; please rebase and preserve both the predecrement-CLR additions and this PR’s trace/chaining changes. The PR remains a draft with its gameplay acceptance still pending, so I have not run a merge validation suite.

rlanday:
All three addressed on the new head:

1. The duplicate `mixed-path` dispatch is gone — the resolution now asserts exactly one dispatch guard and one bench definition, so the class of error can't silently recur.
2. The `try_execute` contract documents the chained-miss count: on `Miss` the count is the instructions retired before the miss — zero for a first-opcode change on the entered trace, non-zero when a chained continuation misses after its parent retired — and names the caller obligation `run_decoded_simple_batch` implements.
3. Rebased onto current `master` (0.10.3): the predecrement-CLR cycle-table and test additions coexist with this PR's trace/chaining changes, resolved in the introducing commits.

Gates on the rebased head: 200 lib tests (all four regressions present), the run_batch integration suite including the SMC-continuation exactness test, fmt clean, clippy clean.

rlanday:
One self-caught correction pushed: the negative test's store opcode was `0x3088` (`MOVE.W A0,(A0)` — address-register source), so the loop stored its own address over its head opcode rather than performing the intended idempotent self-store; the test discriminated the gate for an accidental reason. Corrected to `0x3080` (`MOVE.W D0,(A0)`): the trace now compiles with the store intact and memory-bails at op 1 on every call exactly as documented, and the gate mutation fails it in both directions (re-verified). Full suite 200 green.

rlanday:
Gameplay acceptance for the draft, from a capped session (2.33B guest instructions) with this branch bundled: the motivating thrasher `00EF8D92` collapsed from 330,199 calls at 38.9 ops/call (prior session) to **10,068 calls at 625 ops/call** — call volume down 97%, 6.29M instructions retired through it — and its sibling `00EF8DB2` runs 2,977 ops/call at 26.6% exits. Continuation chaining consolidates the boundary crossings exactly as the design intends. Session-wide trace coverage was 54.5%. Observables are disjoint from the other branches in the bundle (their targets are named blocked heads; this branch's signature is exit-ratio collapse on existing traces).

benletchford (changes requested):
Two blockers remain on the current head:

1. Exit seeding can leave a recording active across a watched `run_batch` boundary. `note_trace_exit` runs before the chain target watch check; on `StartRecording` it installs `self.recording` and sets `cpu.trace_recording = true`, then `run_decoded_simple_batch` observes the watched PC and returns without stopping that recording. I reproduced this by compiling the taken spine, using a watch on the continuation for the first seed, single-stepping/redirecting back to the head, and taking the second side exit: `run_batch` returned `WatchedPc` with `cpu.trace_recording` still true. A debugger or runner may mutate state at that boundary, so the next execution can extend a recording whose intent predates the host boundary. Please gate promotion/recording on a watched exit target (while preserving the candidate for a later unwatched entry), or otherwise guarantee no recording survives the return, and add this regression.

2. This head is based on 0.10.3 and now conflicts with 0.10.5 in `benches/microbench.rs` and `src/core/trace_jit.rs`. Please rebase onto current `master` and preserve both this continuation logic/benchmark and the intervening immediate-memory-MOVE additions (`MoveImmMem`, its decode/emit/portable paths, match arms, tests, and benchmark dispatch).

The current head otherwise passes fmt, Clippy across all targets/features, default tests, all-feature tests, docs, and package verification locally.

rlanday:
From Claude Fable 5:

Both blockers are addressed on the new head (`b78aa4b`).

**1. Watched exit targets no longer seed recordings (or chain).** The watch check now runs *before* `note_trace_exit`, and the outcome is threaded in: a watched exit target still counts candidacy hits, but neither chains nor starts a recording, and the preserved candidate promotes on a later unwatched exit. Your reproduction shape is covered by `watched_exit_targets_never_seed_a_recording_across_the_boundary`, which asserts two layers: the unit contract (a hot candidate hit under a watch returns `None` with no recording installed, then promotes on the next unwatched exit), and the boundary invariant from your repro — 200 consecutive watched `run_batch` returns over the mixed-path rig, sweeping the compile/seed/promote transitions with the continuation pcs watched, asserting after every return that neither `cpu.trace_recording` nor recorder state survives. Reverting the gate makes the regression fail (mutation-checked).

**2. Rebased onto current `master` (0.10.5, `ccce7e7`).** The four commits were replayed preserving the intervening immediate-memory-MOVE additions in full — `MoveImmMem` decode/emit/portable paths, its match arms, tests, and benchmark dispatch all intact alongside this PR's continuation logic, tests, and `mixed-path` benchmark (single dispatch block).

Local gates on the new head: `cargo fmt --check`, `clippy --all-features --all-targets` clean, default-feature suite (including the Musashi/SST fixtures), all-features lib tests, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features`, and the run_batch integration tests all pass.

benletchford (commented):
The previously requested watched-exit and chained-miss fixes are present on b78aa4b, and an additional constrained-budget chained-miss probe did not expose an accounting regression. Current `master` now includes #108 and the 0.10.6 release, however, and this branch conflicts in the trace-JIT work. Please rebase onto current `master`, preserving the new `Link`/`Unlk` decode, native and portable execution paths, atomic-bail/cycle tests, and `link_unlk_frame_loop_matches_step`; then rerun the full repository checks on the integrated result.

rlanday:
From Claude Fable 5:

Rebased onto current `master` (0.10.6, `ae51410`) as requested, preserving #108's LINK/UNLK additions in full — decode, native and portable executors, dispatch arms, cycle charges, tests, and the `link-unlk` benchmark dispatch — alongside this PR's continuation logic, its tests (including the watched-exit regression), and the `mixed-path` benchmark (single dispatch block).

Local gates on the new head: `cargo fmt --check`, `clippy --all-features --all-targets` clean, default-feature suite including the Musashi/SST fixtures, all-features lib tests, and run_batch integration (including both LINK/UNLK and mixed-path step-parity tests) all pass.

benletchford (changes requested):
A zero-budget chained validation miss can still exceed `run_batch`'s exact instruction budget. The parent always calls the child with `instr_budget.saturating_sub(retired)`; when the parent has retired the final permitted instruction, that value is zero. The child validates code before the `instr_budget < ops_len` check, so a changed first opcode still returns `Miss` with PC advanced. `run_decoded_simple_batch` accounts the parent's retired count, returns the miss, and `run_batch` dispatches the changed opcode, reporting one instruction more than the requested budget.

Please prevent chaining from surfacing/consuming a miss when no instruction budget remains (or move the budget rejection ahead of validation without weakening SMC handling at executable entries), and add an exact-boundary regression: compile a continuation, rewrite its first opcode, take the side exit with a budget equal to the parent retired count, and assert both the returned count and PC remain at the budget boundary.

rlanday:
From Claude Opus 5:

Fixed as requested (`HEAD`), merged up to current master (0.10.7) so the salvage machinery from #109 runs alongside exit seeding.

**The fix.** Rather than special-casing the chain site, `try_execute` now rejects a zero instruction budget at entry, before any validation:

```rust
if instr_budget == 0 {
    return None;
}
```

That is the same boundary for every caller, and it is where the invariant belongs: with nothing left to retire, even a validation miss is too much, because consuming a changed opcode hands `run_batch` one instruction past its exact budget. Chained children are the only callers that can currently arrive with zero — the parent passes `instr_budget.saturating_sub(retired)` — but the guard does not depend on that remaining true.

**The regression** (`zero_budget_entry_never_consumes_a_validation_miss`) drives a compiled continuation exactly as the chain site does, on both sides of the boundary: it compiles a continuation, rewrites its first opcode, then enters with budget 0 and with budget 1. At 0 the call returns `None` with PC unmoved at the continuation head — nothing validated, nothing consumed. At 1 the miss still surfaces with the changed opcode consumed and PC advanced, which is the correct self-modification handling and fits the count. Mutation-checked: removing the guard makes the zero-budget case fail.

One note on how I arrived at the entry-level guard rather than a chain-site one: my first attempt at a two-level parent-plus-child regression could not actually produce a parent that guard-exits with retired exactly equal to the budget — the iteration allotment (`by_instrs.max(1)`) makes the parent run a whole extra lap — so that test would have asserted on a state the engine cannot reach. Driving the child directly reproduces the reported path faithfully and keeps the assertion honest.

Full suite green on the merged head: default suite including the Musashi/SST fixtures (814 tests), lib jit-only (186) and all-features (212), run_batch integration (35), `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build.

rlanday:
From Claude Opus 5:

Correcting my previous push: I brought master in with a **merge** commit, and CI runs `committed --no-merge-commit`, so `Check commit messages` failed on it. The branch is now a linear rebase onto 0.10.7 — six commits, no merge, every subject a valid Conventional Commit type (`perf`/`fix`/`docs`/`test`). The zero-budget fix and its regression are unchanged and sit at the tip.

The rebase itself needed care, so here is what was verified rather than assumed. Replaying these commits conflicts only at shared insertion anchors, and resolving those by keeping the upstream side silently dropped two of this branch's own edits — the corrected store encoding in the memory-bail test (`0x3088` → `0x3080`, the whole point of `test: correct the store encoding…`) and the third argument added to `note_trace_exit` call sites. Both survived the test suite, because a stale opcode in a bench-shaped test still runs and a signature change that never landed simply isn't exercised. So the final rebase preserves the branch side on conflicting hunks, and I audited it in both directions:

- **branch content**: every non-trivial line added by the original commits is present in the rebased tree — 0 missing across `trace_jit.rs`, `op_cache.rs`, `benches/microbench.rs`, `run_batch_tests.rs`;
- **master content**: every non-trivial line #109 added is likewise present — 0 missing. (That direction caught one real casualty: `bench_salvaged_prefix_loop` had been dropped while its dispatch arm survived. It is restored in the commit whose conflict dropped it.)

Full suite on the linearized head: default 814/0 (including the Musashi/SST fixtures), lib jit-only 186/0, all-features 212/0, run_batch integration 35/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean, `cargo fmt --check` clean.

benletchford (approved):
The current head addresses the requested chained-miss, watched-exit, zero-budget, integration, and documentation concerns. I audited the complete diff and reran the repository gates on 4474a64913f893f61744b7486c80f86f5e0a849a: formatting, strict all-target/all-feature Clippy, default and all-feature tests, warning-free all-feature docs, and package verification all pass.

PR #103, opened 2026-08-09
perf: JIT CLR to predecrement destinations

rlanday:
## Summary

Extend `ClrMem` (introduced in #97) to predecrement
destinations: `CLR.B/W/L -(An)`, no extension word. The predecrement
form adds one ordering obligation the indexed form does not have — the
register update — and it commits only after the window and
store-not-code guards pass, so a bail leaves the stack pointer
untouched.

## Why this form

In the 2026-08-08 capped (`--cpu-mhz 25`) gameplay session (847M guest
instructions), `42A7` (`CLR.L -(SP)`) blocks the #1, #2, and #6 heads
of the opcode-admission ranking — 22.6M projected dispatches combined,
9× anything else admissible. It is the opcode whose first admission
(#76) caused native wait-spins; that hazard is closed by the wait
classifier (#81 — a store ends a wait shape, so `ClrMem` classifies
non-poll) and the checked-store discipline.

## Focused benchmark

`benches/microbench.rs clr-predec`: `CLR.L -(SP)` + `CLR.W -(SP)` with
an ADDQ restoring SP each iteration, so the loop runs at a fixed stack
pointer; the cleared slots and restored SP prove the stores executed.
Five alternating runs against base carrying an identical copy: median
47.5 → 272.0 M instr/s (**5.7×**); per-pair ratios 5.3–6.4.

- **Headless** (deterministic A/B): as predicted, near-neutral — the
  only headless `42A7`/`4267` occurrences are two 1-hit heads that wall
  at a toolbox trap two links later. Documented so the headless
  neutrality is expected, not a surprise.
- **Capped gameplay session (2026-08-09, 2.53B guest instructions, branch bundled on the then-current master):
  the acceptance bar passes at the #1 target.** `00F1AA60` compiled:
  31 ops, 1,489,937 calls, **42,276,623 instructions retired**
  (1.7% of the whole session through this one trace), 28.4 ops/call,
  guard exits on 10.1% of calls. The other two heads consumed their
  CLRs and advanced to the `61FF` call boundary with zero retirement —
  consistent with the call-wall pattern this series has documented;
  their yield is unreachable by any opcode admission. Session-wide:
  jit_retired 1,086,012,018 = 43.0% of all instructions executed
  (scene-dependent; not comparable to other sessions' shares).

## Correctness

- Decode: `42A7` → `ClrMem { Long, PreDec(7) }` with no phantom
  extension word.
- Guard ordering: the decremented address is computed, both guards run
  against it, and only then are the register write, the store, and the
  flags committed. Bail tests on both paths assert SP untouched; the
  mutation committing the register before the guards fails the SP
  assertion.
- Byte-on-A7: `CLR.B -(A7)` steps by 2 to keep SP even, verified
  against the interpreter's decoded execution (the portable executor
  delegates to it, keeping one source of truth).
- Cycle maxima extend the Table 8-2 convention: 14 byte/word, 22 long
  for `-(An)`.

## Validation

- 192 lib tests on the branch, clippy clean, wasm check green.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -868,6 +868,58 @@ fn bench_indexed_dest_store_loop() {
     );
 }
 
+/// Exercise the predecrement CLR forms the gameplay profile flagged:
+/// CLR.L -(SP) and CLR.W -(SP) with an ADDQ restoring the stack pointer
+/// each iteration, so the loop runs indefinitely at a fixed SP.
+fn bench_clr_predec_loop() {
+    // 408,560 outer cycles of 514 instructions end exactly at the outer
+    // label with SP restored; the six bytes below SP end cleared.
+    const CYCLES: u32 = 408_560;
+    const INSTRS: u32 = CYCLES * 514;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x727F, // outer: MOVEQ #127,D1
+        0x42A7, // inner: CLR.L -(SP)
+        0x4267, // CLR.W -(SP)
+        0x5C8F, // ADDQ.L #6,SP
+        0x51C9, 0xFFF8, // DBRA D1,inner
+        0x60F2, // BRA.S outer
+    ];
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    for address in 0x3FFA..0x4000 {
+        bus.write_byte(address, 0xAA);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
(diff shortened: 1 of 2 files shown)


rlanday:
Second capped session (different scene, 756.7M instructions, this branch bundled on current master): the target head compiled again with 384,943 calls / 11,223,793 instructions retired / 7.1% guard exits — proportionally consistent with the 42.3M-retired result in the 3× longer session already in the body. The admission's yield reproduces across scenes.

benletchford (approved):
Reviewed the complete diff and validated the exact head locally: formatting, Clippy, documentation, default and all-feature tests, and package verification all pass. Predecrement state commits only after the memory and code-overlap guards, A7 byte stepping and CLR flags match the interpreter, and the cycle bounds match the 68000 timing model.

PR #104, opened 2026-08-09
perf: JIT immediate MOVE stores to memory destinations

rlanday:
## Summary

Admit `MOVE #imm,<memory>` as a `MoveImmMem` trace op: byte/word
immediates to `(An)`, `(An)+`, `-(An)`, `(d16,An)`, and `(d8,An,Xn)`,
and long immediates to the extension-less destinations. The bounding
constraint is stated in the decoder: `TraceBuildOp` carries two
extension words, so `MOVE.L #imm,(d16,An)` (immediate high/low plus
displacement) and the long indexed form stay on the decoded path. No
source registers, so none of `MoveMem`'s staging; a destination
register update (postinc/predec) commits only after the window and
store-not-code guards, so a bail leaves the register untouched — the
mutation committing SP before the guards fails the bail test's SP
assertion. The indexed arm reuses the brief-indexed address form the
store side gained in #97.

## Why these forms

In the 2026-08-09 capped gameplay session (2.53B guest instructions),
`3F3C` (`MOVE.W #imm,-(SP)`) blocked seven heads at 20.9M projected
dispatches combined — argument pushes ahead of the call sites the
recorder stops at, the second-largest opportunity behind the `61FF`
call walls. After admitting it, two of those heads advanced to `31BC`
(`MOVE.W #imm,(d8,An,Xn)`), which fits the extension budget and is
admitted by the second commit.

## Focused benchmark

`benches/microbench.rs move-imm-store`: a word push with an ADDQ
restoring SP plus an indexed immediate store swept by the reloaded
DBRA counter; the landed values and restored SP prove the stores
executed. Five alternating runs against base carrying an identical
copy: per-pair ratios 8.9–9.1, **median 9.0×** (median 51.0 → 458.3 M
instr/s).

Capped gameplay session (756.7M guest instructions, this branch
bundled): **zero heads remain blocked on any admitted form** — the
entire 20.9M-dispatch `3F3C` projection is consumed, with former heads
either compiling into the session's 44.5% whole-session trace coverage
or advancing to `31BC` (now also admitted) and, for two, to the `61FF`
call boundary. Headless is near-neutral by expectation and
measurement: these forms are argument-push shapes that headless's
dialog-weighted workload rarely executes.

## Correctness

- Decode covers every admitted form plus both budget rejections (the
  three-extension long forms must not decode as `MoveImmMem`).
- MOVE sets N/Z from the value, clears V/C, and preserves X — asserted
  with X set across all four native-parity cases (predec word, predec
  long, displacement with a negative value, indexed).
- Store-into-code bails atomically on both paths with SP untouched
  (mutation-tested, as above).

## Validation

- 196 lib tests, clippy clean, wasm32 check green, `cargo fmt` clean.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -920,6 +920,59 @@ fn bench_clr_predec_loop() {
     );
 }
 
+/// Exercise the immediate-store family the gameplay profile flagged: a
+/// word push (MOVE.W #imm,-(SP)) with an ADDQ restoring SP, and an
+/// indexed immediate store swept by the DBRA counter.
+fn bench_move_imm_store_loop() {
+    // Outer cycles of 514 instructions (128 four-op iterations plus the
+    // reload and restart) end exactly at the head with SP restored.
+    const CYCLES: u32 = 350_000;
+    const INSTRS: u32 = CYCLES * 514;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x3F3C, 0x1111, // head: MOVE.W #$1111,-(SP)
+        0x31BC, 0x0042, 0x1200, // MOVE.W #$42,(0,A0,D1.W*2)
+        0x548F, // ADDQ.L #2,SP
+        0x51C9, 0xFFF2, // DBRA D1,head
+        0x727F, // MOVEQ #127,D1 (reload: D1 is both counter and index)
+        0x60EC, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(0, 0x4000);
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(1, 0x7F);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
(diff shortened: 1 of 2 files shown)


benletchford (changes requested):
Three blockers before this can be reconsidered:

1. The branch now conflicts with current master after #103. Please rebase it and preserve both the new predecrement CLR support and this PR’s immediate-MOVE decode/execution/benchmark additions in the overlapping trace-JIT sections.
2. `cargo fmt --all -- --check` fails at the long `assert_eq!` in `bench_move_imm_store_loop`; run the formatter and rerun CI.
3. The correctness test compares native execution only with `execute_portable_move_imm_mem`, so both sides share the new addressing, flag, and `max_cycles` logic. Add an interpreter/`step()` differential (or direct interpreter-cycle comparison) covering the admitted predecrement word/long, displacement, and indexed forms, and assert the exact N/Z/V/C/X state and 68000 cycle totals. As written, a shared flag or cycle regression would pass.

On the current head, Clippy, docs, default/all-feature tests, and package verification pass locally; the items above are the remaining blockers.

rlanday:
All three addressed:

1. Rebased onto current `master` (0.10.3, with #103's predecrement CLR); the overlapping cycle-table and test sections carry both features, resolved in the introducing commits.
2. `cargo fmt --all -- --check` clean, the benchmark's long `assert_eq!` included.
3. New interpreter differential (`move_immediate_matches_the_interpreter_with_exact_68000_cycles`): the portable executor against `step()` on a 68000 for every admitted form — predec word/long, displacement (zero value so Z is exercised), and indexed — asserting identical registers, the stored bytes read from each twin's own memory at the interpreter-computed destination, full N/Z/V/C/X, and cycle-charge equality with the interpreter's true 68000 count, which promotes the move-table constants from citation to verification. One harness lesson the first draft caught: `step()` writes through the bus while the portable path writes through the attached window, so each twin now owns its memory and the comparison reads both at the destination.

Gates on the rebased head: 199 lib tests, fmt clean, clippy clean.

benletchford (approved):
The revised head resolves all three blockers: it is rebased with #103 preserved, formatting is clean, and the new interpreter differential independently verifies every admitted addressing form, stored bytes, complete NZVCX state, and exact 68000 cycles. I reviewed the full updated diff and ran the complete CONTRIBUTING.md suite locally on c8e972a (formatting, Clippy with all targets/features, docs with warnings denied, default and all-feature tests, and package verification); all pass. Remote title validation and Build & Test are also green.

PR #107, opened 2026-08-10
perf: record through BSR calls on a retry, with per-segment SMC intervals

rlanday:
## Why

Gameplay trace-opportunity profiling of EV Override (capped GUI sessions under the Systemless host, `trace-profile` feature — the profiler ranks recording heads by hits and names the opcode each failed recording blocked on) shows call instructions as the dominant recording blocker class: hot loops whose bodies call small leaf routines through `BSR`, and — after the frame-op walls fall — `BSR.L` far calls into ROM routines megabytes away. A recording that stops at the call leaves the whole loop interpreted.

## What

Two mechanisms, layered:

**1. Retry-gated record-through.** The first recording at a head still blocks at a `BSR` exactly as today, so the ungated recorder stays byte-identical to the ranked-blocker behavior the opportunity profile is built on. When a recording ends blocked on a `BSR` (0x61xx), the head's slot re-arms as a candidate carrying call-through permission; the retry records through the call as three op forms:

- `CallThrough { return_pc }`: a checked push of the constant return address (window-bounds, pre-68020 alignment, and self-modification guards all precede the store and the SP update).
- The callee's body, recorded inline as ordinary ops.
- `RtsReturn { expected_return }`: a checked load of the stacked return address, compared against the recorded constant **before** SP moves; a mismatch bails with the pop uncommitted, and the interpreter re-executes the RTS against whatever the guest really stacked.

Depth is one recorded call bracket at a time (a nested call blocks the recording at that op like any unsupported instruction; sequential brackets are admitted). BSR.B/W/L are all decoded; cycles match the interpreter (BSR 18, RTS 16).

**2. Per-segment self-modification intervals.** Trace stores are guarded against the trace's *own code bytes* so self-modifying code bails and re-forms, like the interpreter would observe. With one interval per trace, a far callee would inflate the guarded span across the whole caller–callee gap and false-bail every unrelated store between the regions — which is why an earlier revision capped the recorded span at 4KB and rejected far calls outright. `CompiledTrace` now carries **two tight intervals** — the caller's bytes and the callee's — every store checks both, the gap is deliberately unguarded, and the recorder caps each segment at 4KB independently. Ordinary traces have a zero-width callee interval and emit exactly the code they did before. The watched-PC pre-filter also checks both intervals, so a watchpoint inside a far callee keeps the trace cold exactly like one inside the caller's span.

## Measured effect

**Focused microbenches** (in-repo `benches/microbench.rs`, filters `call-through` / `far-call-through`): 100M instructions per run after a 5M warmup, deterministic `run_batch` workloads. Both arms built `--features jit` on the same idle machine; base is `origin/master` carrying identical copies of the benchmarks. Five alternating base/cand runs; the statistic is the median of pairwise ratios (arms drift together with machine state):

*Near call (BSR.W to a two-op leaf in a counted loop):*

| run | base | cand |
|-----|------|------|
| 1 | 76.7 | 344.9 |
| 2 | 74.0 | 346.1 |
| 3 | 76.3 | 344.6 |
| 4 | 76.6 | 345.3 |
| 5 | 75.4 | 344.0 |

Median pairwise **4.52×** (M instr/s; base blocks at the BSR and interprets).

*Far call (BSR.W at maximum forward reach, leaf 32KB away, with a store into the inter-region gap every iteration — the case a unified SMC interval false-bails):*

| run | base | cand |
|-----|------|------|
| 1 | 73.3 | 353.1 |
| 2 | 73.5 | 346.0 |
| 3 | 73.7 | 346.7 |
| 4 | 74.0 | 345.1 |
| 5 | 73.6 | 344.5 |

Median pairwise **4.70×** (base rejects the far call at admission).

**Whole-application evidence.** Headless 900M-instruction EV Override boot A/B vs master is flat (native_calls 68,043 → 68,043; jit_retired within 200 of 16.375M): this workload's far-call heads block on ops *inside* the ROM callees (LINK, MOVEM.L, LEA abs.L), so recording through the call moves each wall into the callee body rather than opening it alone. The gameplay profile shows the mechanism end to end: with frame-op admission alongside (separate PR), failed-shape prefixes record through `61FF` calls with 32-bit displacements up to +165KB into ROM callee bodies, and the capped-session retirement rises 43.7% → 46.65% for the bundle. This PR is the foundation of that chain; its standalone effect is the focused benches plus the admission of every call-shaped region the callee-side decoders can finish.

## Validation

- `retry_gated_call_through_records_and_compiles_the_leaf_inline` — the blocked-then-retried lifecycle compiles call + leaf + checked return + tail as one trace, with lockstep counters and a balanced or exactly-mid-call stack.
- `rts_return_mismatch_bails_after_the_push_commits` — the committed push stands (it is architecturally real), the mismatched RTS retires nothing, resume pc and flags asserted.
- `far_leaf_call_through_compiles_with_two_code_intervals` — the compiled far-call trace's caller and callee intervals asserted exactly.
- `store_between_caller_and_far_callee_retires_without_bailing` — the discriminator against a unified interval: a store into the gap must retire natively (native and portable). Verified to fail under a collapse-to-union mutation.
- `store_into_far_callee_code_bails_before_committing` — the second interval has teeth: nothing commits, resume at the store (native and portable). Verified to fail without the callee-interval check.
- `bsr_long_far_leaf_records_through_on_retry` — BSR.L's 32-bit-displacement decode path end to end (128KB leaf).
- Integration (`run_batch_tests`): `call_through_leaf_loop_matches_step` and `far_leaf_call_through_matches_step` — instruction-exact parity with single-step interpretation through the whole lifecycle, the far case at BSR.W's maximum reach.

**Composition note**: a prefix-salvage change (separate PR) compiles blocked recordings' prefixes. If both land, the merge should keep the retry ahead of salvage at a BSR blocker — a first blocked-at-call recording must reject so the slot re-arms with permission; the permitted retry's own blocked recording salvages normally.


benletchford (changes requested):
Blocking correctness/performance issue: the advertised 4 KiB per-segment cap is only checked when the BSR is intercepted. At that point the BSR has not been appended, no callee instruction has been recorded, and the post-return caller tail does not exist yet. Later ops are appended without another span check, and `compile_decoded_ops_reason` computes `code_start/code_end` and `callee_start/callee_end` without enforcing `CALL_THROUGH_MAX_SPAN`. A callee that branches more than 4 KiB before returning therefore compiles with an oversized callee interval and false-bails stores anywhere in that hole, contrary to the two-tight-interval design. Enforce the bound after the complete trace shape is known (or incrementally as ops are recorded) and add a regression with a far intra-callee branch plus a store in the resulting gap.

The branch is also currently conflicting with `master`, and Build & Test fails at `Check commit messages`: `Give call-through traces...` lacks a Conventional Commit type, while `test+bench:` and `bench:` are not allowed types in `committed.toml`. Rebase while preserving #108's LINK/UNLK additions and rewrite those commits to valid single-line Conventional Commit subjects before re-running the full checks.

rlanday:
From Claude Fable 5:

All three items are addressed on the new head (`12a2b7a`).

**1. The span cap is now enforced on the complete recorded shape.** You are right that the admission-time check ran before the BSR was appended, before any callee instruction existed, and before the post-return tail was recorded — a callee that branched widely before returning compiled an oversized callee interval and false-bailed stores in the hole. `compile_decoded_ops_reason` now enforces `CALL_THROUGH_MAX_SPAN` on both segments after computing the complete spans, with its own reject reason (`CallSpan`, reported as `call-span` in the profile so the opportunity ranking names it). The admission-time check remains as a cheap early-out; the compile-time check is authoritative. Regression: `wide_callee_rejects_when_the_complete_shape_exceeds_the_span_cap` — a callee with a `+0x1F00` interior recorded branch rejects with `CallSpan`, while an otherwise identical near-branching control compiles and its tight `callee_start/callee_end` are asserted exactly. Disabling the compile-time cap makes the regression fail (mutation-run).

**2. Rebased onto current `master` (0.10.6, `ae51410`)**, preserving #108's LINK/UNLK decode, native and portable executors, dispatch arms, cycles, and tests. The portable store predicate's move to the shared `CodeSpans` carrier is adopted by the LINK executor and the LINK/UNLK tests in the same commit that introduces it.

**3. Commit subjects rewritten to allowed Conventional Commit types**: `perf:` for the two feature commits, `test:` for the benchmark/regression commits.

Local gates on the new head: `cargo fmt --check`, `clippy --all-features --all-targets` clean, default-feature suite (including the Musashi/SST fixtures), all-features lib tests, run_batch integration (both step-parity call-through tests), and the benchmarks compile and run.

benletchford (changes requested):
There is a remaining CPU-model decode bug in the retry path. `decode_call_op` treats every BSR opcode whose low byte is `0xFF` as the 68020+ long-displacement form, without consulting `cpu_type`. On M68000/M68010/SCC68070, `0x61FF` is the short −1 form (as the repository’s `m68000_bcc_ff_displacement_is_short_minus_one` coverage already establishes). Once such a head has earned call permission, the recorder therefore reads words at `pc+2` and `pc+4` that are not operands of the executed instruction, records the wrong return PC, and can introduce guest-visible extra bus accesses before the odd-target address fault. Thread the CPU model into call decoding, use the long form only on 68020+, and add a counting/faulting-bus regression proving a permitted pre-68020 `0x61FF` performs no extension reads and cannot be recorded as BSR.L.

rlanday:
From Claude Fable 5:

Fixed as requested (`HEAD`), on top of a rebase onto current master (0.10.7, salvage preserved — its three unit cases, the step differential, and the `salvage-prefix` benchmark all run green alongside this branch's additions).

- `decode_call_op` now takes the recording's CPU type; the `0xFF` displacement byte decodes as BSR.L only on 68020+. On M68000/M68010/SCC68070 it falls to the short-form arm — return PC `pc+2`, no extension words.
- Regression `bsr_ff_call_decode_is_cpu_aware`: the pre-68020 half hands the decoder a bus that panics on any read, proving a permitted `0x61FF` performs no extension reads and records the short return PC; the 68020+ half uses a counting bus and pins the long form to exactly its two operand words with the `pc+6` return. Mutation-checked: reverting the CPU gate makes the panicking bus fire.
- Full suite green on the rebased head: lib (jit-only and all-features), default suite including the Musashi/SST fixtures, run_batch integration, clippy clean, portable build.

benletchford (changes requested):
Two blockers remain on the current head:

1. The new call-through push/pop paths are unsafe with a nonempty FastMem window shorter than four bytes. Native `emit_call_through` / `emit_rts_return` reach `checked_window_off`, which subtracts 4 from `fm_len` without first guarding `fm_len < 4`; the portable helpers repeat `cpu.fm_len - 4`. For lengths 1–3 the limit wraps, offset zero can pass, and the raw pointer access extends beyond the advertised window. Reject short windows before the subtraction on both paths (or harden the shared helper), and add push/pop regressions that assert a bail with SP and memory unchanged.

2. The composition with #109 does not implement the documented retry-before-salvage rule. On the first permissionless BSR blocker, `finish_recording(...Blocker)` can compile a >=8-op prefix through an earlier branch. The subsequent rearm only runs when the slot is `Rejected`, so that head never receives the permitted retry and never records through the call. Please add a regression with a salvageable interior branch before BSR, make the first recordable-call blocker rearm a permitted recording, and only allow a permitted attempt to salvage at a later blocker.

benletchford (commented):
Correction to blocker 1 in my review: I followed the FastMem value back through the public entry point, and `run_batch` only captures it when `fm.len >= 4` and the pointer is non-null (`src/core/execute.rs`). The `fm_*` scratch fields are crate-private and cleared after the batch, so the short-window arithmetic is protected by that invariant in reachable library use. I withdraw the short-window blocker and do not require those regressions here.

The retry-before-salvage blocker remains: the first permissionless call can still become a compiled salvaged prefix before the post-`finish_recording` rearm is considered, preventing the promised permitted retry.

rlanday:
From Claude Opus 5:

Fixed, and thank you — this one was real, and it explains something the gameplay census had been showing me without my understanding why.

**The fix.** `finish_recording` now takes a `call_retry_pending` flag, set exactly when a recording stops at a call the head does not yet have permission for. The prefix trim is skipped in that case, so the head rejects and the existing rearm installs the permitted `Counting` slot as designed. Once the head holds permission the flag is clear, so a later blocker — including one inside the callee — salvages exactly as #109 intends. On the stacked #114 the flag is computed from `is_recordable_call`, so it covers JSR blockers too.

**The regression** (`a_salvageable_prefix_before_a_call_still_yields_the_permitted_retry`) is built to your description: eleven admissible ops with a recorded interior branch, two non-terminal ops after it, then `BSR.W` to a leaf — so the blocked recording genuinely *is* salvageable. It asserts the head's finally-compiled trace contains a `CallThrough` op. Mutation-checked by removing the precedence: the test then fails with "the head must record through its call, not stop at a salvaged prefix (8 ops compiled)", which is precisely the eight-op salvaged prefix you predicted.

**Why this mattered more than it looked.** The most recent capped gameplay session has twenty-eight heads in the top-forty blocked table stopping at `61FF`, together about 43.2M projected instructions — 92% of all ranked mass. I had checked whether any of them were nested-call cases needing deeper call-through, and none were: every one dies at its *first* call, which was hard to explain given the permission machinery. Your blocker is a mechanism that produces exactly that signature. I am not claiming the field numbers yet — that needs a session on a build with this fix, which I will run — but the shapes match.

Full suite on the fixed head: default 814/0 (including the Musashi/SST fixtures), lib jit-only 189/0, all-features 215/0, run_batch integration 35/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

And noted on the FastMem point — thank you for chasing it back to the `run_batch` capture invariant and withdrawing it rather than leaving it to me to argue. I have not added those regressions.

benletchford (changes requested):
The new regression covers the salvage path, but the general retry precedence is still incomplete. If the first unpermitted BSR is the immediate target of the last recorded branch, `recording.ops.last().op.ends_trace()` is already true. `finish_recording_with_retry` skips the trim, then still clears that branch's `expected_taken`, compiles the branch-ended prefix successfully, and installs `Compiled`; the post-finish rearm still only runs for `Rejected`. That head therefore never gets its permitted retry, just as before, without entering the trim block at all.

Please make `call_retry_pending` force rejection/rearm before any prefix compilation—not merely suppress salvage—and add a regression where a recorded branch lands directly on BSR and the final compiled trace must contain `CallThrough`. The branch also conflicts with current master after #101/#116 and must be rebased with those paths preserved before validation.

rlanday:
From Claude Opus 5:

You were right — the salvage-only fix was incomplete, and your description of the remaining path was exact.

**Fixed.** A pending call retry now bypasses prefix compilation entirely rather than merely suppressing the trim: the region is rejected before `compile_decoded_ops_reason` runs, so the rearm's rejected-slot check fires and the head takes its permitted retry. The trim's own `call_retry_pending` clause is gone, since it is unreachable once the earlier return exists.

**Regression** (`a_branch_landing_on_a_call_still_yields_the_permitted_retry`): a recorded conditional branch whose target *is* the unpermitted `BSR`, so the region's last op is an ordinary terminal and it never enters the trim.

Two things I got wrong while writing that test, both worth stating because they changed the result:

1. My first version asserted on the *end* state after 60,000 instructions, and it passed with and without the fix — unrelated slot churn eventually replaces a call-less trace, exactly as you said. It now steps in 200-instruction batches and fails the moment a call-less trace is observed at the head, which is the invariant that actually matters.
2. My first shape used `BEQ` while `D1` is nonzero in steady state, so the branch fell *through* to a NOP and the recording ended on a non-terminal — reproducing the salvage case the sibling test already covers, not yours. Instrumenting `finish_recording` showed `len=9 last_term=false`; with `BNE` the branch is taken onto the call and the region ends on the terminal as intended.

Mutation-checked in the corrected form: restoring the salvage-only behaviour fails with "a call-less prefix was installed, starving the permitted retry", while the salvage sibling still passes.

**Rebased onto current master (0.10.9)**, preserving #101's exit seeding/chaining and #116's absolute CLR. That rebase needed real integration work rather than conflict resolution alone: exit seeding constructs `TraceRecording` and `TraceSlot::Counting` directly, and those constructors had to learn this branch's call fields. A seeded candidate now inherits its slot's permission, so a head rearmed after a call blocker keeps its permitted retry even when the recording that finally starts came from an exit seed rather than a backward branch — that commit is separate and titled accordingly.

I audited the result in both directions before pushing: every non-trivial line added by this branch's commits is present (0 missing), and every non-trivial line master added is present (the only three flagged are ones this branch deliberately changes — the `finish_recording` signature and two portable-trace calls converted to the `CodeSpans` carrier).

Full suite: default 819/0 including the Musashi/SST fixtures, lib jit-only 199/0, all-features 225/0, run_batch integration 38/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean.

benletchford (changes requested):
The current head still fails the required formatting gate. `cargo fmt --all -- --check` reports an overlong `execute_portable_trace` call in `src/core/trace_jit.rs` (around line 10533), and the GitHub Build & Test job consequently stopped before Clippy, documentation, both test matrices, packaging, and commit validation. Please run `cargo fmt --all`, push the formatting-only correction, and let the complete required workflow finish successfully.

rlanday:
From Claude Opus 5:

Fixed, and thank you for the precise pointer — the offending line was exactly where you said.

The cause: rebasing this branch restores tests that master added, and two of those call `execute_portable_trace` with the pre-`CodeSpans` `(code_start, code_end)` pair. Converting them to `CodeSpans::caller(...)` pushed the call past the line limit, and my last push captured that state without a final `cargo fmt --all`. Both calls are now converted and formatted, and I verified `cargo fmt --all -- --check` is clean on the pushed head rather than assuming it.

Also rebased onto current master (0.10.10) now that #115 has merged, with the two-way audit I have been using: every non-trivial line this branch's commits add is present (0 missing), and so is every line master added — the only entries that direction flags are the ones this branch deliberately changes (the `finish_recording` signature and the portable-trace calls converted to the carrier).

Full suite on the pushed head: default 821/0 including the Musashi/SST fixtures, lib jit-only 201/0, all-features 227/0, run_batch integration 39/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean, `cargo fmt --all -- --check` clean.

benletchford (approved):
The current head resolves the prior blockers: the retry is applied before prefix salvage, BSR 0x61ff remains an 8-bit -1 displacement before 68020, both caller and callee code intervals are bounded and protected, and the trace preserves checked stack-memory and exact cycle behavior. I also validated this exact SHA locally with formatting, Clippy for all targets/features, default and all-feature tests (including doctests), and cargo package.

PR #108, opened 2026-08-10
perf: admit LINK/UNLK frame ops into traces

rlanday:
Builds on the same admission machinery as #107 (call-through); its field
effect composes with that PR and with prefix salvage, as measured below.

## Why

Gameplay trace-opportunity profiling (EV Override, capped GUI session, `trace-profile` feature — the profiler ranks recording heads by hits and names the opcode each failed recording blocked on) shows the largest class of blocked trace mass is now ROM-prologue shapes: **9 of the top 15 blocked heads stop at `4E56` = LINK A6,#imm**, ~24M projected ops across them, and three heads block on LINK/UNLK as their *first* instruction. The same profile's most-interpreted memory op is `486E` (PEA d16(A6), 11.2M executions, 6.5% of decoded memory ops) — a form the JIT already supports, stranded inside those LINK-blocked regions.

## What

`LINK An,#d16` (0x4E50|reg) and `UNLK An` (0x4E58|reg) become recordable trace ops:

- **LINK**: one checked push (window-bounds, pre-68020 alignment, and self-modification guards all precede the store) of An's current value, then An = the decremented SP, then SP moves by the displacement — the interpreter's exact order, 16 cycles to match `exec_link`.
- **UNLK**: one checked load from An, then SP = An+4 and An = the loaded frame pointer, 12 cycles to match `exec_unlk`.
- Neither touches condition codes. A bail commits nothing (all checks precede all effects), so the interpreter re-executes the op from scratch.
- **A7 forms are never admitted**: LINK A7's pushed value is generation-dependent (the 68040 alone decrements before reading the source), and ROM prologues use A6. Excluding A7 keeps the trace semantics CPU-generation-independent.
- LINK.L (68020+, 0x4808) is not admitted — it does not appear in the profile.

## Measured effect

**Focused microbench** (in-repo `benches/microbench.rs`, filter `link-unlk`): a LINK / store-through-A6 / UNLK / DBRA loop, 100M instructions per run after a 5M-instruction warmup, single deterministic `run_batch` workload. Both arms built `--features jit` at the same optimization settings on the same idle machine; the base arm is `origin/master` carrying an identical copy of the benchmark. Five alternating base/cand runs; the statistic is the median of the five *pairwise* ratios (the arms drift together with machine state, so pairwise ratios are stabler than cross-run means):

| run | base (M instr/s) | cand (M instr/s) |
|-----|-----------------|------------------|
| 1 | 40.6 | 331.8 |
| 2 | 39.4 | 325.1 |
| 3 | 39.7 | 313.4 |
| 4 | 39.3 | 316.2 |
| 5 | 35.2 | 311.5 |

Median pairwise ratio **8.2×** (base median 39.4, cand median 316.2). Base rejects the head at op 0 (LINK unsupported) and interprets the whole loop.

**Whole-application headless A/B**: deterministic 900M-instruction EV Override boot workload under the Systemless host, `--features jit,trace-profile`, identical assets both arms; the counters are the JIT's own: `native_calls` (entries into compiled traces) and `jit_retired` (guest instructions retired inside them).

| counter | base (master) | cand (link-unlk) |
|---------|--------------|------------------|
| native_calls | 68,043 | 68,054 |
| jit_retired | 16,375,668 | 16,375,726 |
| rejected_hits | 524,768 | 524,755 |

Standalone headless movement is negligible (+11 native calls) because this
workload's frame-op heads block again on the next decoder gap in the same
ROM routines: the per-head profile shows the op-0 LINK/UNLK heads recording
their frame op and blocking exactly one op later (`00F07052`: op-0 `4E5E` →
`4E75` RTS at +2; `00F1B62A`: op-0 `4E56` → `203C` MOVE.L #imm at +4).

**Gameplay acceptance (composed).** In a capped EV Override GUI session of a
bundle carrying this PR together with call-through (#107) and prefix
salvage, against the prior session of the same bundle without them
(near-identical lengths, 636.7M vs 639.1M instructions): `4E56` disappeared
from the blocked-head table entirely (it had been the #1 blocker class, 9
of the top 15 heads, ~24M projected ops), session-wide native retirement
rose 43.72% → **46.65%**, and the hot loops fed by the formerly blocked
regions surged (e.g. head `00EEFFC0` 49.9M → 72.5M retired ops). The
attribution to this PR specifically is the disappearance of the `4E56`
blocker class; the retirement rise is shared with salvage.

## Validation

- `link_and_unlk_a7_forms_are_never_admitted` — decode admission, including the A7 negative and op lengths.
- `native_link_unlk_matches_portable` — native and portable executors agree on registers, memory, retired count, and cycle charge for the frame loop.
- `link_bails_atomically_on_window_and_own_span` — out-of-window and store-into-own-code bails each fire with nothing committed (registers, memory, and resume pc asserted), native and portable.
- `unlk_bails_atomically_on_misalignment` — pre-68020 odd frame pointer bails before either register moves, native and portable.
- `link_unlk_frame_loop_matches_step` (integration) — the frame loop matches single-step interpretation exactly through the whole record/compile/execute lifecycle, including instruction counts, registers, and memory.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -973,6 +973,56 @@ fn bench_move_imm_store_loop() {
     );
 }
 
+/// The ROM-prologue frame shape: LINK, a store through the frame pointer,
+/// UNLK, loop. Base cannot admit LINK (the head blocks at op 0) and
+/// interprets the whole loop; with admission the frame ops compile.
+fn bench_link_unlk_frame_loop() {
+    const INSTRS: u32 = 100_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x4E56, 0xFFF8, // head: LINK A6,#-8
+        0x5283, // ADDQ.L #1,D3
+        0x3D43, 0xFFFC, // MOVE.W D3,-4(A6)
+        0x4E5E, // UNLK A6
+        0x51C8, 0xFFF2, // DBRA D0,head
+        0x707F, // MOVEQ #127,D0
+        0x60EC, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(0, 0x7F);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(3) > 5_000, "the loop actually iterated");
+    // The budget may stop mid-iteration; the frame slot trails D3 by at
+    // most one.
+    assert!(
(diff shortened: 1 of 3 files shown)


benletchford (approved):
Validated LINK/UNLK decode, stack-access atomicity, register-update order, cycle charges, A7 exclusion, native/portable parity, and step-level regression coverage. The full repository formatting, Clippy, default/all-feature test, documentation, and package checks pass locally.

PR #109, opened 2026-08-10
perf: salvage a blocked recording's prefix through its last branch

rlanday:
## Why

Gameplay trace-opportunity profiling of EV Override (capped GUI sessions under the Systemless host, `trace-profile` feature) shows long recordings dying deep: regions record 14–26 admissible ops and then hit one instruction the decoder refuses (`LEA (abs).L`, `MOVEM.L`, wide multiplies, PC-relative jump tables), and the whole recording — good prefix included — is rejected. The head then re-probes forever. The failed-shape table's top entries are exactly these: multi-block prefixes crossing several recorded branches, terminated by a single inadmissible op.

## What

`finish_recording` already attempts compilation of whatever was recorded on every end; a recording stopped by an unsupported instruction fails only because its op list lacks a trace terminal. When such a recording's prefix already crossed a recorded branch, trim back to the last one and compile that instead:

- Ops are execution-ordered, so the recorded op *after* the trimmed terminal names the true continuation (whichever way the branch went); the guest re-enters full dispatch exactly where the trimmed tail began.
- The trimmed terminal branch is normalized to an ordinary final branch, exactly as when a recording stops at a branch destination today.
- The salvage bar is `SALVAGE_MIN_OPS = 8`, above `TRACE_MIN_OPS`: a salvaged trace exits to the interpreter at its terminal every pass, so short fragments would pay entry/exit costs without covering enough work.
- Recordings whose last op already is a terminal, blocked recordings with no recorded branch, and below-bar prefixes behave exactly as before.

## Measured effect

**Focused microbench** (in-repo `benches/microbench.rs`, filter `salvage-prefix`): a store-heavy loop — five MOVE.W stores interleaved with register counters, an interior conditional branch, two register tests past the branch (so the recording ends on a non-terminal), then the `LEA (abs).L` blocker and a DBRA tail — 100M instructions per run after a 5M warmup, deterministic `run_batch` workload. Both arms `--features jit` on the same idle machine; base is `origin/master` with an identical benchmark copy. Five alternating runs, median of pairwise ratios:

| run | base | cand |
|-----|------|------|
| 1 | 59.6 | 134.2 |
| 2 | 64.6 | 128.9 |
| 3 | 59.4 | 126.9 |
| 4 | 56.9 | 103.3 |
| 5 | 65.0 | 134.5 |

Median pairwise **2.07×** (M instr/s; base rejects the head — the recording's last op is not a terminal — and interprets the whole loop; the candidate natively retires the twelve-op salvaged prefix and interprets the five-op tail). A register-only nine-op variant of the shape measures ~1.0×: per-iteration trace entry/exit cost cancels the win on cheap ops, which is what the `SALVAGE_MIN_OPS` bar is for.

**Whole-application headless A/B**: deterministic 900M-instruction EV Override boot workload under the Systemless host (`--features jit,trace-profile`, identical assets both arms), counters from the JIT itself:

| counter | base (master) | cand (salvage) |
|---------|--------------|----------------|
| native_calls | 68,043 | **81,857** (+20.3%) |
| jit_retired | 16,375,668 | **16,495,136** (+119K) |
| rejected_hits | 524,768 | **510,921** (−13.8K) |
| avg ops per call | 240.67 | 201.51 |

Heads that re-probed and re-rejected forever now run compiled prefixes; the average ops per call drops because salvaged traces are short, while total retirement rises.

**Gameplay evidence (composed).** In a capped EV Override GUI session of a bundle carrying this PR with call-through (#107) and frame-op admission (#108), against the prior session of the same bundle without them (near-identical lengths): session-wide native retirement rose 43.72% → **46.65%**, unclassified memory-bail volume collapsed ~67K → ~13K, and heads that previously showed thousands of failed re-recordings now appear in the compiled table with trimmed shapes. The attribution to this PR specifically is the conversion of blocked-shape mass into short compiled traces; the retirement rise is shared with #108.

## Validation

- `blocked_recording_salvages_the_prefix_through_the_last_branch` — a blocked nine-op prefix compiles trimmed at its branch, the unsupported tail excluded from the recorded ops, lockstep counters across the salvaged-native + interpreted-tail iteration.
- `short_blocked_recordings_still_reject` — a five-op prefix whose last terminal sits at op four stays below the bar and rejects exactly as before.
- `blocked_recording_without_a_branch_still_rejects` — a branchless prefix has nothing to trim to.
- `salvaged_prefix_loop_matches_step` (integration) — the salvage shape matches single-step interpretation exactly through record/trim/compile/execute, including instruction counts, registers, and memory.

**Composition note**: if this lands alongside call-through (#107), the merge should keep the retry ahead of salvage at a BSR blocker — a first blocked-at-call recording must reject so the slot re-arms with call permission; the permitted retry's own blocked recording salvages normally. Without that ordering a salvaged prefix occupies the slot and the record-through retry never fires.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1051,7 +1051,8 @@ fn bench_salvaged_prefix_loop() {
         0x4E71, // NOP (skipped)
         0x4A42, // TST.W D2 -- past the branch: master has no terminal here
         0x4A42, // TST.W D2
-        0x41F9, 0x0000, 0x3000, // LEA $3000.L,A0 -- the blocker
+        0x4E57, 0x0000, // LINK A7,#0 -- refused by design (A7 exclusion)
+        0x4E5F, // UNLK A7
         0x51C8, 0xFFD2, // DBRA D0,head
         0x707F, // MOVEQ #127,D0
         0x60CC, // BRA.S head
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1028,9 +1028,10 @@ fn bench_link_unlk_frame_loop() {
 /// ROM regions the gameplay profile shows dying 14-26 ops deep. Base
 /// rejects the whole head and interprets; with salvage the twelve-op
 /// prefix through the branch compiles and only the tail stays
-/// interpreted. (A register-only nine-op variant measures ~1.0x -- the
-/// per-iteration trace entry/exit cost cancels the win on cheap ops; the
-/// salvage bar exists for exactly that reason.)
+/// interpreted. Two register tests sit between the branch and the
+/// blocker so the recording's last op is not a terminal: without salvage
+/// the whole head rejects. (A register-only nine-op variant measures
+/// ~1.0x -- trace entry/exit cost cancels the win on cheap ops.)
 fn bench_salvaged_prefix_loop() {
     const INSTRS: u32 = 100_000_000;
     const CODE_BASE: u32 = 0x6000;
@@ -1048,10 +1049,12 @@ fn bench_salvaged_prefix_loop() {
         0x4A41, // TST.W D1
         0x6602, // BNE.S +2 (taken until D1 wraps 16 bits)
         0x4E71, // NOP (skipped)
+        0x4A42, // TST.W D2 -- past the branch: master has no terminal here
+        0x4A42, // TST.W D2
         0x41F9, 0x0000, 0x3000, // LEA $3000.L,A0 -- the blocker
-        0x51C8, 0xFFD6, // DBRA D0,head
+        0x51C8, 0xFFD2, // DBRA D0,head
         0x707F, // MOVEQ #127,D0
-        0x60D0, // BRA.S head
+        0x60CC, // BRA.S head
     ];
     let mut bus = LinearMemoryBus::new(0x1_0000);
     for (index, word) in words.iter().enumerate() {
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1023,6 +1023,73 @@ fn bench_link_unlk_frame_loop() {
     );
 }
 
+/// The field shape salvage targets: a store-heavy body with one interior
+/// branch, then an instruction the decoder refuses (LEA (abs).L) -- the
+/// ROM regions the gameplay profile shows dying 14-26 ops deep. Base
+/// rejects the whole head and interprets; with salvage the twelve-op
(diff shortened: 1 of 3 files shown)


benletchford (changes requested):
This cannot be merged with the current history: Build & Test fails specifically at `Check commit messages`. The commits headed `bench: salvaged-prefix loop...` and `test+bench: make the salvage shapes...` use types not allowed by `committed.toml` (use one of the configured Conventional Commit types, such as `test:` for regression/benchmark support). The branch also conflicts with current `master` in the trace-JIT/benchmark integration. Please rebase, preserve #108's LINK/UNLK paths and tests, keep the call-through retry ahead of salvage if #107 is incorporated, rewrite the invalid subjects, and rerun the complete checks on the resolved result.

rlanday:
From Claude Fable 5:

Both items are addressed on the new head (`b55ed36`).

**1. Rebased onto current `master` (0.10.6, `ae51410`).** The rebase surfaced a real interaction worth noting: #108's LINK/UNLK admission made these tests' inadmissible tail admissible, so all three salvage shapes silently stopped blocking. The final commit swaps the blocker to `LEA (abs).L` — same six-byte footprint as the LINK+UNLK pair it replaces, so every branch displacement is unchanged — and the positive test again fails with the trim disabled (mutation-run).

**2. Commit subjects rewritten to allowed types**: the two flagged commits are now `test:`.

Local gates on the new head: `cargo fmt --check`, `clippy --all-features --all-targets` clean, default-feature suite including the Musashi/SST fixtures, all-features lib tests, run_batch integration (including both step-parity tests), and the benchmarks compile.

rlanday:
From Claude Fable 5:

One proactive commit (`HEAD`): the discriminating shapes' blocker is now the A7 LINK/UNLK pair instead of `LEA (abs).L`. Absolute LEA is on the queue after the latest gameplay census (it blocks twenty of the top-forty heads there), which would have silently defused these tests a third time the way #108's LINK admission did. The A7 forms are refused by documented design — the pushed value is generation-dependent — so no future decoder coverage can admit this blocker; the pair nets zero stack motion and matches the six-byte footprint it replaces, so every displacement is unchanged. All suites green on the new head.

benletchford (approved):
The prior history and rebase blockers are resolved. The blocker-only trim is confined to recording formation, preserves the recorded continuation, and has focused rejection/parity coverage. The full CONTRIBUTING validation suite passes on the current head.

PR #111, opened 2026-08-10
perf: admit long register-mask MOVEM push/pop into traces

rlanday:
## Why

Gameplay trace-opportunity profiling of EV Override (capped GUI sessions under the Systemless host, `trace-profile` feature) shows that once LINK/UNLK are admitted (#108), the top blocked class becomes the ROM caller-save idiom: **`48E7` (MOVEM.L regs,−(SP)) and `4CDF` (MOVEM.L (SP)+,regs) block 7 of the top 16 recording heads**, often deep inside callees reached through recorded calls, and they gate the profile's most-interpreted op behind them (`486E`, PEA d16(A6), 12.2M executions — a form the JIT already supports, stranded in MOVEM-blocked regions).

## What

The long register-mask MOVEM pair becomes recordable, for any base register:

- **`MOVEM.L <regs>,-(An)`**: one contiguous range check (window bounds, pre-68020 alignment, self-modification overlap) covers the whole transfer **before anything commits**, then the stores run unrolled at constant offsets in ascending-address order, then the base register updates. A bail commits nothing.
- **`MOVEM.L (An)+,<regs>`**: one bounds/alignment range check, then unrolled loads in ascending order, then the base update.
- The predecrement mask is bit-reversed (bit 0 = A7 … bit 15 = D0) with transfers running highest-register-first into descending addresses — which leaves ascending addresses holding ascending registers, so both directions walk one shared ascending-order helper. Native and portable use the same helpers and cannot diverge on order.
- **A base register inside its own list is never admitted**: the predec store's value is generation-dependent (the 68040 decrements before reading the source), and the postinc load is overwritten by the final address. Word-size forms and other EAs fall back to the interpreter.
- Cycles are the MC68000 formula (store 8+8n, load 12+8n); `(An)+`/`-(An)` carry zero per-mode timing overhead on every CPU path, so one charge is exact everywhere.

## Measured effect

**Focused microbench** (in-repo `benches/microbench.rs`, filter `movem-frame`): the caller-save idiom — MOVEM.L push of D2/D3/A2, a three-op body, MOVEM.L pop, DBRA — 100M instructions per run after a 5M warmup, deterministic `run_batch` workload. Both arms `--features jit` on the same idle machine; base is current `master` (0.10.6) with an identical benchmark copy. Five alternating runs, median of pairwise ratios:

| run | base | cand |
|-----|------|------|
| 1 | 20.3 | 222.1 |
| 2 | 20.9 | 229.5 |
| 3 | 22.7 | 250.6 |
| 4 | 24.8 | 267.9 |
| 5 | 25.3 | 274.8 |

Median pairwise **10.9×** (M instr/s) — the largest focused ratio in this optimization line, because interpreted MOVEM pays per-register dispatch and bus traffic that the compiled transfer collapses into raw window stores behind a single range check. (Both arms trend upward together across the runs with machine state; the pairwise-ratio statistic is used for exactly that reason.)

**Whole-application headless A/B**: deterministic 900M-instruction EV Override boot workload under the Systemless host (`--features jit,trace-profile`, identical assets both arms), counters from the JIT itself:

| counter | base (0.10.6) | cand (movem) |
|---------|--------------|--------------|
| native_calls | 68,054 | 68,055 |
| jit_retired | 16,375,726 | 16,375,652 |
| rejected_hits | 524,755 | 524,753 |

Standalone headless movement is negligible, the same shape #108 showed: this workload's MOVEM heads sit inside regions that still block on other decoder gaps (`4850` PEA (An), wide multiplies, PC-relative jump tables), so admission moves each wall deeper rather than opening it alone. The gameplay profile is what names this op class as the top blocker — 7 of the top 16 heads — and the field payoff composes with #107/#109 exactly as #108's did (that composition measured +2.9pp session-wide retirement when LINK/salvage landed together in a bundled session). A bundled gameplay session with this branch is the acceptance follow-up.

## Validation

- `movem_long_decodes_and_excludes_base_in_list` — decode admissions with exact cycle charges, plus the negatives: base-in-list (both directions), word-size forms, other EAs, and the empty mask.
- `native_movem_long_matches_portable` — a push+pop round-trip loop: native and portable agree on registers, the exact memory image (ascending addresses hold ascending registers, asserted byte-for-byte), retired count, and cycle charge; the stack balances and the saved registers survive the round trip.
- `movem_long_bails_atomically` — out-of-window, store-into-own-code, and pre-68020 odd-base bails each fire with nothing committed, native and portable.
- `movem_long_roundtrip_loop_matches_step` (integration) — the idiom matches single-step interpretation exactly through the whole record/compile/execute lifecycle.
- The ascending-order helper is mutation-checked: dropping the predec mask's bit reversal fails the parity test's memory-image asserts.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1867,6 +1867,62 @@ fn bench_tst_abs_loop() {
     );
 }
 
+/// The ROM caller-save idiom the gameplay profile names as the top
+/// blocked class after LINK admission: MOVEM.L push, a short body,
+/// MOVEM.L pop, loop. Base cannot admit the register-mask forms and
+/// interprets the whole loop.
+fn bench_movem_frame_loop() {
+    const INSTRS: u32 = 100_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x48E7, 0x3020, // head: MOVEM.L D2/D3/A2,-(SP)
+        0x5284, // ADDQ.L #1,D4
+        0x5285, // ADDQ.L #1,D5
+        0x3684, // MOVE.W D4,(A3)
+        0x4CDF, 0x040C, // MOVEM.L (SP)+,D2/D3/A2
+        0x51C8, 0xFFF0, // DBRA D0,head
+        0x707F, // MOVEQ #127,D0
+        0x60EA, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(3, 0x4000);
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(0, 0x7F);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(4) > 5_000, "the loop actually iterated");
+    assert!(
+        cpu.d(4).abs_diff(cpu.d(5)) <= 1,
+        "counters advance in lockstep"
+    );
+    assert!(
+        cpu.a(7) == 0x8000 || cpu.a(7) == 0x7FF4,
+        "stack balanced or exactly mid-bracket: {:#06x}",
+        cpu.a(7)
+    );
+    println!(
+        "batch     movem frame loop        {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 fn main() {
     println!("m68k microbench");
     let only = std::env::args().nth(1);
@@ -1948,6 +2004,10 @@ fn main() {
     }
     if only.as_deref() == Some("tst-abs") {
         bench_tst_abs_loop();
+        return;
+    }
+    if only.as_deref() == Some("movem") {
+        bench_movem_frame_loop();
         return;
     }
     if only.as_deref() == Some("clr-abs") {
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -16143,4 +16143,93 @@ mod portable_tests {
         assert_eq!(portable.a(7), native.a(7));
         assert_eq!(pmem, mem, "memory effects agree");
     }
+
+    /// The MOVEM range guard has to honour the *second* code interval, not
+    /// just the caller's. A call-through trace carries a far callee, and a
+    /// predecrement frame push whose range straddles the callee's own bytes
+    /// must bail with nothing committed -- exactly as a single-word store
+    /// aimed there does. A caller-only range check would let it through and
+    /// the trace would overwrite the code it is about to return into.
+    #[cfg(all(feature = "jit", not(target_family = "wasm")))]
+    #[test]
+    fn movem_predec_into_far_callee_code_bails_before_committing() {
+        let mut ops = far_call_store_loop_ops();
+        // Replace the plain store with a MOVEM.L push through A0. Two
+        // registers, so the range is [A0-8, A0).
+        ops[2] = TraceBuildOp {
+            opcode: 0x48E0,
+            extension: Some(0x0300),
+            extension2: None,
+            pc: 0x0104,
+            op: JitTraceOp::MovemLongPredec {
+                base: 0,
+                mask: 0x0300,
+                cycles: 24,
+            },
+        };
+        let spans = CodeSpans {
+            code_start: 0x0100,
+            code_end: 0x0108,
+            callee_start: 0x8100,
+            callee_end: 0x8102,
+        };
+
+        // A0 = 0x8108 puts the range at [0x8100, 0x8108): it covers the
+        // callee's bytes and misses the caller's entirely.
+        let prepare = |mem: &mut [u8]| {
+            let mut c = cpu();
+            c.set_cpu_type(CpuType::M68040);
+            c.set_a(7, 0x0800);
+            c.set_a(0, 0x8108);
+            c.set_d(0, 0xDEAD_BEEF);
+            c.set_d(1, 0xFEED_FACE);
+            attach_window(&mut c, mem);
+            c
+        };
+
+        let mut mem = vec![0u8; 0x10000];
+        let mut native = prepare(&mut mem);
+        let mut jit = TraceJit::new();
+        let compiled = jit
+            .compile_decoded_ops(&native, 0x0100, CpuType::M68040, ops.clone(), Some(0x0100))
+            .expect("far call/MOVEM trace should compile");
+        let packed = unsafe { compiled.call_native(&mut native, 1) };
+        assert_eq!(
+            (packed >> 32) as u32,
+            2,
+            "the call and return retired; the callee-code MOVEM did not"
+        );
+        assert_eq!(
+            &mem[0x8100..0x8108],
+            &[0u8; 8],
+            "nothing from the bailed transfer committed"
+        );
+        assert_eq!(native.a(0), 0x8108, "the base register did not move");
+        assert_eq!(native.pc, 0x0104, "resume at the MOVEM for full dispatch");
+
+        let mut pmem = vec![0u8; 0x10000];
+        let mut portable = prepare(&mut pmem);
+        let ppacked = execute_portable_trace(&mut portable, &ops, spans);
+        assert_eq!(
+            (ppacked >> 32) as u32,
+            2,
+            "portable bails on the same transfer"
+        );
+        assert_eq!(&pmem[0x8100..0x8108], &[0u8; 8]);
+        assert_eq!(portable.a(0), 0x8108);
+
+        // The discriminator: moved clear of both intervals, the very same
+        // transfer must retire. Without this the test would also pass if
+        // MOVEM simply never compiled inside a call-through trace.
+        let mut gapmem = vec![0u8; 0x10000];
+        let mut gap = prepare(&mut gapmem);
+        gap.set_a(0, 0x4000);
+        let gpacked = execute_portable_trace(&mut gap, &ops, spans);
+        assert_eq!(
+            (gpacked >> 32) as u32,
+            4,
+            "a transfer clear of both intervals retires"
+        );
+        assert_eq!(gap.a(0), 0x3FF8, "the base register moved for the push");
+    }
 }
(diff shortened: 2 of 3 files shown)


benletchford (changes requested):
The new whole-transfer range guard is not safe for a short nonempty FastMem window. Both portable paths compare `off` with `cpu.fm_len.wrapping_sub(total)`, and `checked_window_off_range` emits `fm_len - bytes` without first guarding `fm_len < bytes`. For example, a 4-byte window and a 12-byte three-register transfer wrap the limit to a large unsigned value, accept offset zero, and then perform raw-pointer accesses beyond the window. The existing MOVEM.W helper explicitly rejects `bytes > fm_len`; please give both portable and native paths the same precondition and add short-window regressions for load and store bails with no committed state. Also remove the duplicate consecutive `decode_movem_long_trace_op` call in `decode_trace_op`; rejected MOVEM.L shapes currently re-read the extension word unnecessarily.

rlanday:
From Claude Opus 5:

Full suite on the rebased head: default 813/0 (including the Musashi/SST fixtures), lib jit-only 184/0, all-features 210/0, run_batch integration 34/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean. The short-window guard and its regression from my previous push are preserved on top of the rebase.

rlanday:
From Claude Opus 5:

| run | base | cand |
|-----|------|------|
| 1 | 50.4 | 329.5 |
| 2 | 47.8 | 312.4 |
| 3 | 46.6 | 304.7 |
| 4 | 46.0 | 303.8 |
| 5 | 45.8 | 294.1 |

benletchford (commented):
Current `master` now includes #101 and #116, and this head conflicts in the trace-JIT integration. Please rebase while preserving continuation exit seeding/chaining and the absolute-CLR decode, native/portable execution, cycle tests, run-batch regression, and benchmark alongside the MOVEM-long work; then rerun the full repository checks.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.9), preserving #101's continuation exit seeding/chaining and #116's absolute-CLR decode, native and portable execution, cycle coverage, run-batch regression, and benchmark alongside the MOVEM-long work and its short-window guard.

Full suite on the rebased head: default 818/0 (including the Musashi/SST fixtures), lib jit-only 193/0, all-features 219/0, run_batch integration 37/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.10) now that #115 has merged, preserving the AND/OR memory-source work alongside the MOVEM-long work.

benletchford (changes requested):
This head now conflicts with master after #107. Please rebase it and resolve the MOVEM changes on top of the new / trace ops and two-interval  store guard. In particular, keep the short-fast-memory-window underflow checks and ensure MOVEM predecrement stores reject overlap with either caller or callee code interval; then rerun the full required workflow.

benletchford (changes requested):
Correction to the identifiers omitted above: preserve the CallThrough and RtsReturn trace operations and apply MOVEM stores through the two-interval CodeSpans guard.

rlanday:
Rebased onto master at 0.10.12. Two commits: the MOVEM admission, and a separate fix for the point you raised.

**You were right that the range guard was incomplete, and it was incomplete on both paths.** `guard_store_range_not_code` tested only `[code_start, code_end)`, and the portable helper had the same single-interval check open-coded as `masked < code_end && masked + total > code_start`. A predecrement frame push whose range straddled a call-through callee's bytes was therefore emitted unguarded — it would have overwritten the code it was about to return into.

Both now apply the two-interval rule. The native side mirrors `guard_store_not_code` exactly, including the `callee_end > callee_start` test so ordinary traces with a zero-width callee interval emit no extra code. The portable side now takes the `CodeSpans` carrier and asks `spans.store_hits_code(masked, total)`, which already spans both intervals — so there is no MOVEM-specific copy of the rule left to drift.

**The short-window underflow checks are kept**, unchanged and still ahead of the code check:

```rust
if total > cpu.fm_len || off > cpu.fm_len - total {
    return None;
}
```

That ordering matters — `fm_len - total` would wrap on a window shorter than the transfer and accept any offset, which is what `movem_long_bails_on_a_window_shorter_than_the_transfer` pins.

New coverage: `movem_predec_into_far_callee_code_bails_before_committing` puts a two-register push at `[0x8100, 0x8108)`, straddling the callee while missing the caller entirely, and asserts native and portable both bail with no byte written and the base register unmoved. It carries its own discriminator — the same transfer moved clear of both intervals must retire all four ops, so the test cannot pass merely because MOVEM never compiles inside a call-through trace. Mutation-checked twice: it fails with the native `callee_end > callee_start` branch disabled, and fails again with the portable guard reverted to the caller-only comparison.

Gate: fmt clean, `clippy --all-targets --all-features -D warnings` clean, lib jit-only 210/0, default 825/0 including the Musashi and SingleStep fixtures, all-features 909/0.

benletchford (changes requested):
The MOVEM implementation and the short-window/two-interval fixes review cleanly against 0.10.12, but this head is now conflicting with master after #117 and cannot be validated or merged as-is. Rebase it onto current master while preserving #117’s absolute-TST decode, exact cycle table entries, native checked-load and portable execution paths, all three focused unit tests, `mem_trace_tst_absolute_matches_step`, and the `tst-abs` benchmark alongside MOVEM. Keep the existing MOVEM short-window atomic-bail and caller/callee overlap regressions intact, then rerun the full required workflow on the integrated head.

rlanday:
From Claude Opus 5:

Whole-app numbers for this PR, measured on matched arms.

**Instrument.** EV Override 1.0.1 running on `benletchford/systemless`,
driven headless by a scripted input replay: input events are scheduled by
**retired instruction count**, not wall time, so the same script drives every
run to the same point regardless of how fast the build is. The script boots the
app, dismisses the shareware dialog, creates a pilot, and enters the ship,
reaching live flight at roughly 150M instructions.

**Metric.** `jit_retired / total_instructions` — the share of guest instructions
retired by compiled traces rather than the interpreter — read from the
so the denominator is identical by construction and the shares are directly
comparable.

**Arms.** Three binaries, identical profiler, distinct hashes verified:

- `baseline` — master at 0.10.12 (the base this branch was measured on; master has since moved to 0.10.13 with #117)
- `111` — baseline + this PR's two commits
- `stack` — `111` + an unrelated experimental change, included as a control

Each arm was run **8 times**. Every run starts from a pristine copy of the save
directory, because the emulator writes save files on exit and without the reset
the arms drift apart.

**Noise.** The harness is repeatable but not bit-deterministic: the game seeds
default pilot names, and apparently in-game spawns, from a clock source. That is
what the repeats are for — the within-arm spread is the noise floor, and a delta
has to clear it to mean anything.

**Results.**

| arm | n | mean native share | sd | min | max |
|---|---|---|---|---|---|
| baseline | 8 | 45.738% | 0.112 | 45.610 | 45.944 |
| **this PR** | 8 | **46.378%** | 0.416 | 45.540 | 46.798 |

Welch's t-test, 95% CI: **+0.640pp, ±0.299, t = +4.20** — the effect clears the
noise floor.

**Scope and caveats.** This is one application, one scripted route through it,
on one host. The metric is native share, not wall-clock time; it says more guest
work is being retired by compiled traces, which is the thing this PR is meant to
do, and does not by itself quantify a frame-rate change. All 24 runs across the
three arms were verified to have reached gameplay, so no run is excluded — the
one low outlier in the third arm is genuine in-game variance and was left in.

Happy to publish the harness or the raw per-run reports if useful.

rlanday:
From Claude Fable 5:

Rebased onto current master (0.10.14) per your last review; `HEAD` is the integrated head.

**#117's content is preserved, verified at content level rather than by eyeball.** Against master: zero deleted opcode-literal lines and zero deleted definitions in the diff; a two-way function-presence audit over `trace_jit.rs`, `microbench.rs`, and `run_batch_tests.rs` shows nothing from master missing and no duplicate definitions. Specifically present and passing: the absolute-TST decode with its exact cycle entries, native checked-load and portable paths, all three focused TST unit tests, `mem_trace_tst_absolute_matches_step`, and the `tst-abs` benchmark with its dispatch arm. During resolution `bench_movem_frame_loop` was re-inserted as a whole definition cut from this branch's own blob (git had interleaved it with `bench_tst_abs_loop`), and its `movem` dispatch arm is wired.

**MOVEM regressions intact:** the short-window atomic-bail and caller/callee-overlap tests are present and pass on the integrated head.

**Full required workflow on `HEAD`:** rustfmt clean; `clippy --all-targets --all-features -D warnings` clean; lib (`--features jit`) 214 passed / 0 failed; default suite 828 / 0; all-features 914 / 0; `cargo doc` clean. The differential suites also exercise this head against 0.10.14's #126 interpreter fix (MOVEM discarded read after long mem-to-reg transfers) with no divergence.

rlanday:
Fresh evidence for this PR from a trace-profile census taken today (EV Override under Systemless, boot → title crawl → flight, current masters plus the pending startup PRs on the Systemless side):

MOVEM.L push/pop (`48E7`/`4CDF`) is the recorded blocker at heads worth ~203 K projected dispatches in that one session — 12 of the top 45 uncompiled heads. It and PEA/CMPI (#112) are the top-of-list admission blockers in today's game, not just in the profiles from when this PR was written.

— Claude Fable 5

rlanday:
From Claude Fable 5, working with @rlanday — whole-application CPU evidence for the pending perf PRs (#111, #112, #128, #131, #133), measured together.

## Arms

Two Systemless binaries, byte-identical on the Systemless side (v0.16.0, commit 7fa0789), differing **only** in the `m68k` crate:

- **base** — m68k `master` (fd2f2a3)
- **PRs** — m68k `master` + #111 + #112 + #128 + #131 + #133 merged (conflicts between #111/#112 resolved by chaining both decoders and keeping both function/test sets; the combined branch passes the full gates: fmt, clippy `-D warnings`, 925 all-features tests; every added `JitTraceOp` variant verified wired — decode, emit, execute, tests)

## Instrument and metric

One interactive EV Override session per arm (macOS, i9-9880H), launch → registration dialog → ~20 s title crawl → ~20 s flight → close, played back-to-back by the same operator. At window close Systemless prints total **guest** instructions retired; `/usr/bin/time -l` reports **host** CPU seconds, host instructions retired and host cycles. The headline metric is host work **per guest instruction**, which cancels session length and (to first order) phase mix — the raw 8 s sampling windows we tried first were dominated by moment-to-moment variance in what the game was doing.

## Result

| | base | PRs | Δ |
|---|---|---|---|
| guest instructions (session) | 216.7 M | 223.7 M | comparable sessions |
| host instructions / guest instruction | 414.4 | 383.8 | **−7.4 %** |
| host cycles / guest instruction | 211.3 | 194.8 | **−7.8 %** |
| guest instructions / host-CPU-second | 16.0 M | 17.7 M | **+10.2 %** |
| host CPU (user) | 13.51 s | 12.66 s | −6.3 % |

Boot alone (automated, three interleaved pairs, launch → 25 s): host instructions 37.6–37.8 G → 37.4 G (−1 %), cycles/CPU within noise — the boot is dominated by archive decode on the Systemless side, so the whole-session ratios above *understate* the gameplay-phase effect (the constant boot cost dilutes them).

## Honest scope

- One session pair for the headline ratios; the per-guest-instruction normalization is what makes a single pair meaningful on this machine, but it is n=1 — happy to run more pairs on request.
- This is the five PRs **combined**. Per-PR mechanisms are documented in each PR (line-level profile for #133; trace-profile census for #111/#112 — their forms are the top admission blockers in this exact game; recording-storm counters for #131; compile-share for #128).
- A trace-profile census of the same session puts the #111/#112 admission targets at ~2 % of total guest dispatch, #133 at ~4 % of the crawl's main thread, #128 at the JIT's compile share — the measured −7.8 % cycles/guest-instruction is consistent with those mechanisms adding up rather than with any single dramatic effect.

PR #112, opened 2026-08-10
perf: admit the profiled small forms (PEA abs, CMPI.W d16, MOVE #imm,Dn)

rlanday:
## Why

The latest capped-session trace-opportunity profile of EV Override (under the Systemless host, `trace-profile` feature) names three small decoder gaps directly, after the larger walls (#107/#108/#109/#111) are addressed:

- **`4878` PEA (xxx).W** blocks two of the top four recording heads — **10.7M projected ops** (7.5M + 3.2M).
- **`0C68` CMPI.W #imm,(d16,An)** blocks a 5.7M-projected head.
- **`203C` MOVE.L #imm,Dn** is the op the LINK-admission field test exposed one instruction later: ROM prologues stage trap selectors with full-width immediate loads, and the recording dies there.

Three commits, one per form, each self-contained with its tests.

## What

- **PEA (xxx).W / (xxx).L** (`perf: admit absolute-address PEA into traces`): both widths compile as a checked push of the decoded constant, sharing the existing PEA store discipline (window bounds, pre-68020 alignment, self-modification guard, all before the store and the A7 update). Cycles carry each width's extension-fetch cost (16/20). The (xxx).W form sign-extends.
- **CMPI.W #imm,(d16,An)** (`perf: admit the displacement form of memory CMPI.W`): joins the indexed form; the displacement EA skips the brief-extension arithmetic and charges 16 cycles — pinned by a step() differential on the 68000, not assumed.
- **MOVE.W/L #imm,Dn** (`perf: admit full-width immediate loads into data registers`): a pure register op — NZ from the sized value, VC cleared, X preserved, word writes merging into the low word. Cycles 8/12, also pinned by a step() differential.

## Measured effect

**Focused microbenches** (filters `pea-abs`, `cmpi-disp`, `move-imm-reg`): 100M instructions per run after a 5M warmup, deterministic `run_batch` workloads; both arms `--features jit` on the same idle machine; base is `origin/master` (0.10.6) with an identical benchmark copy; five alternating runs per form; the statistic is the median of pairwise ratios.

*`pea-abs`* (PEA (xxx).W + rebalance + counter):

| run | base | cand |
|-----|------|------|
| 1 | 55.4 | 306.4 |
| 2 | 56.6 | 300.1 |
| 3 | 56.8 | 309.5 |
| 4 | 57.3 | 330.3 |
| 5 | 60.2 | 327.6 |

Median pairwise **5.45×** (M instr/s).

*`cmpi-disp`* (CMPI.W of a displaced field + branch + counter):

| run | base | cand |
|-----|------|------|
| 1 | 59.3 | 393.8 |
| 2 | 61.5 | 394.4 |
| 3 | 58.0 | 403.2 |
| 4 | 58.3 | 408.6 |
| 5 | 60.1 | 390.3 |

Median pairwise **6.64×**.

*`move-imm-reg`* (both immediate-load widths + counter):

| run | base | cand |
|-----|------|------|
| 1 | 44.8 | 437.5 |
| 2 | 44.4 | 433.0 |
| 3 | 45.5 | 436.2 |
| 4 | 46.3 | 433.4 |
| 5 | 46.3 | 442.3 |

Median pairwise **9.59×**. In each case base rejects the head at the unadmitted form and interprets the whole loop.

**Whole-application headless A/B** (deterministic 900M-instruction EV Override boot workload, counters from the JIT):

| counter | base (0.10.6) | cand (small forms) |
|---------|--------------|---------------------|
| native_calls | 68,054 | 68,047 |
| jit_retired | 16,375,726 | 16,375,691 |
| rejected_hits | 524,755 | 524,762 |

Standalone headless movement is negligible, the same shape #108 and #111 showed: this workload's heads that touch these forms sit in regions still blocked on other decoder gaps, so each admission moves its wall deeper rather than opening it alone. The gameplay profile is what names these three forms — the two top-four `4878` heads, the `0C68` head, and the `203C` blocker exposed one op past LINK — and the field payoff composes with #107/#109/#111 the way #108's did (+2.9pp session retirement when LINK/salvage first landed together). A bundled gameplay session carrying this branch is the acceptance follow-up.

## Validation

- PEA abs: decode of both widths incl. sign extension and lengths; native/portable parity on the pushed constant, cycles, and memory; own-span push bails with nothing committed on both paths; a step-parity loop.
- CMPI disp: a step() differential for exact flags and cycle charges on both admitted forms; native/portable parity incl. flags and a far-read bail with flags untouched; a step-parity loop with a counter mutation (a bare compare/branch pair is a pure poll and stays refused).
- MOVE #imm,Dn: decode incl. the byte-form and address-register negatives; a step() differential for exact flags/cycles of both widths; native/portable parity incl. the word-merge and X preservation; a step-parity loop.
- Full suites green at the tip: lib tests (all-features and jit-only), default-feature suite including the Musashi/SST fixtures, run_batch integration, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1923,6 +1923,140 @@ fn bench_movem_frame_loop() {
     );
 }
 
+/// PEA of an absolute address, rebalanced -- the profile's second-widest
+/// blocked head class after MOVEM. Base blocks the head at the PEA.
+fn bench_pea_abs_loop() {
+    const INSTRS: u32 = 100_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x4878, 0x3000, // head: PEA ($3000).W
+        0x588F, // ADDQ.L #4,A7
+        0x5283, // ADDQ.L #1,D3
+        0x51C8, 0xFFF6, // DBRA D0,head
+        0x707F, // MOVEQ #127,D0
+        0x60F0, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(0, 0x7F);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(3) > 5_000, "the loop actually iterated");
+    assert_eq!(
+        bus.read_word(0x7FFE),
+        0x3000,
+        "the pushed constant lands below the rebalanced stack"
+    );
+    println!(
+        "batch     pea abs loop            {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+/// CMPI.W against a displaced structure field, with a counter mutation
+/// so the wait classifier admits the loop. Base blocks the head at the
+/// displacement form.
+fn bench_cmpi_disp_loop() {
+    const INSTRS: u32 = 100_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x0C68, 0x0042, 0x0010, // head: CMPI.W #$42,($10,A0)
+        0x6602, // BNE.S +2
+        0x4E71, // NOP
+        0x5283, // ADDQ.L #1,D3
+        0x51C8, 0xFFF2, // DBRA D0,head
+        0x707F, // MOVEQ #127,D0
+        0x60EC, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    bus.write_word_at(0x3010, 0x0042);
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(0, 0x3000);
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(0, 0x7F);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(3) > 5_000, "the loop actually iterated");
+    println!(
+        "batch     cmpi disp loop          {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+/// Full-width immediate loads (the `203C` trap-selector shapes), with a
+/// counter. Base blocks the head at the first load.
+fn bench_move_imm_reg_loop() {
+    const INSTRS: u32 = 100_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x323C, 0x8123, // head: MOVE.W #$8123,D1
+        0x2A3C, 0x0000, 0xA89F, // MOVE.L #$A89F,D5
+        0x5283, // ADDQ.L #1,D3
+        0x51C8, 0xFFF2, // DBRA D0,head
+        0x707F, // MOVEQ #127,D0
+        0x60EC, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(0, 0x7F);
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(3) > 5_000, "the loop actually iterated");
+    assert_eq!(cpu.d(5), 0x0000_A89F, "the long load lands");
+    println!(
+        "batch     move imm reg loop       {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 fn main() {
     println!("m68k microbench");
     let only = std::env::args().nth(1);
@@ -2008,6 +2142,18 @@ fn main() {
     }
     if only.as_deref() == Some("movem") {
         bench_movem_frame_loop();
+        return;
+    }
+    if only.as_deref() == Some("pea-abs") {
+        bench_pea_abs_loop();
+        return;
+    }
+    if only.as_deref() == Some("cmpi-disp") {
+        bench_cmpi_disp_loop();
+        return;
+    }
+    if only.as_deref() == Some("move-imm-reg") {
+        bench_move_imm_reg_loop();
         return;
     }
     if only.as_deref() == Some("clr-abs") {
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -204,6 +204,14 @@ pub(crate) enum JitTraceOp {
         reg: u8,
         data: u32,
     },
+    /// `MOVE.W/L #imm,Dn`: a full-width immediate load into a data
+    /// register. Pure register op -- NZ set from the sized value, VC
+    /// cleared, X preserved; word writes merge into the low word.
+    MoveImmReg {
+        reg: u8,
+        size: Size,
+        value: u32,
+    },
     UnaryDataReg {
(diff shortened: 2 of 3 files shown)


benletchford (changes requested):
The implementation and its focused semantics/cycle tests passed the full local CONTRIBUTING suite on the reviewed head, but this branch now conflicts with `master` after #109 in `trace_jit.rs`, the benchmark dispatch, and `run_batch_tests.rs`. Please rebase onto current master while preserving #109's `SALVAGE_MIN_OPS` blocker-only trim in `finish_recording`, its three salvage unit cases, `salvaged_prefix_loop_matches_step`, and the `salvage-prefix` benchmark alongside this PR's PEA/CMPI/MOVE additions, then rerun the complete checks on the integrated result.

rlanday:
From Claude Opus 5:

Full suite on the rebased head: default 817/0 (including the Musashi/SST fixtures), lib jit-only 186/0, all-features 212/0, run_batch integration 36/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean.

rlanday:
From Claude Opus 5:

| run | base | cand |
|-----|------|------|
| 1 | 50.4 | 329.5 |
| 2 | 47.8 | 312.4 |
| 3 | 46.6 | 304.7 |
| 4 | 46.0 | 303.8 |
| 5 | 45.8 | 294.1 |

benletchford (commented):
Current `master` now includes #101 and #116, and this head conflicts in the trace-JIT integration. Please rebase while preserving continuation exit seeding/chaining and the absolute-CLR decode, native/portable execution, cycle tests, run-batch regression, and benchmark alongside these small-form additions; then rerun the full repository checks.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.9), preserving #101's continuation exit seeding/chaining and #116's absolute-CLR decode, native and portable execution, cycle coverage, run-batch regression, and benchmark alongside the PEA-abs, CMPI-displacement and immediate-load small forms.

Full suite on the rebased head: default 822/0 (including the Musashi/SST fixtures), lib jit-only 195/0, all-features 221/0, run_batch integration 39/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.10) now that #115 has merged, preserving the AND/OR memory-source work alongside the PEA-abs, CMPI-displacement and immediate-load small forms.

benletchford (changes requested):
This head now conflicts with master after #107. Please rebase it while preserving the new call-through ops, per-segment trace validation, and both self-modification intervals. The PEA absolute store must use the resulting two-interval guard, and the CMPI/MOVE/PEA regression blocks need to be combined without dropping the new call-through coverage; then rerun the full required workflow.

rlanday:
Rebased onto master at 0.10.12 and squashed to a single commit covering all three forms.

**The PEA absolute store now uses the two-interval guard.** It shares `emit_pea_disp` with the displaced form, differing only in how the pushed address is materialised (a constant instead of `An + d16`), so it reaches the same `guard_store_not_code` — which tests the caller interval and, when the trace has one, the callee interval as well. The portable path likewise goes through `execute_portable_pea_disp`, whose guard is `spans.store_hits_code(masked, 4)` on the `CodeSpans` carrier. Neither path has a PEA-specific copy of the guard to fall out of date.

**Call-through coverage is intact.** Three conflicts landed exactly on the dispatch sites your review pointed at, and all three were resolved by keeping both sides rather than either:

- the native op dispatch keeps the `CallThrough` and `RtsReturn` arms and widens the PEA arm to `PeaDisp { .. } | PeaAbs { .. }`,
- the memory-write classifier lists `MoveImmReg` and `MoveImmMem` alongside `CallThrough` and `RtsReturn`,
- the portable dispatch keeps `execute_portable_call_through` / `execute_portable_rts_return` and routes both PEA forms to `execute_portable_pea_disp(cpu, op, spans)` — the two-interval signature, not the old `(code_start, code_end)` pair.

The CMPI/MOVE/PEA regressions are combined in one branch as you asked, and a two-way function-name audit against master over all three files confirms no master test was dropped in the process.

Gate: fmt clean, `clippy --all-targets --all-features -D warnings` clean, lib jit-only 211/0, default 829/0 including the Musashi and SingleStep fixtures, all-features 912/0.

rlanday:
New measurement that changes what this PR is worth, so I want to put it on the record rather than leave the description as-is.

I re-censused EV Override against merged master (0.10.12, so with call-through and durable permission in) to re-rank what to work on next. **`4878` — `PEA (xxx).W`, one of the forms this PR admits — is the single largest blocked class in the whole census.**

| rank | head | hits | prefix | projected instructions | blocker |
|------|------|------|--------|------------------------|---------|
| **1** | `00EE5D36` | 1,103,805 | 28 | **30,906,456** | `4878` |
| 12 | `00F1634A` | 120,258 | 10 | 1,202,570 | `4878` |

That is 32.1M projected from `4878` alone, against 12.9M for the next uncovered class. The rank-1 head is a single 28-op prefix hit 1.1M times in one session, declined every time.

The other two forms here also carry real mass in the same census: `0C6A` + `0C68` (displaced CMPI.W) total ~1.0M projected across ranks 22 and 25.

Method, so the number is checkable: capped play session on the stationary scene, 2,565,566,846 instructions, `trace-profile` build of `benletchford/systemless` against m68k `908ef70` with identical assets and save state to the previous session. "Projected instructions" is the profiler's own column — hits times recorded prefix length — so it is an upper bound on what admitting the form could convert, not a predicted speedup. I have not measured a whole-application delta for this branch on the merged base; the microbench numbers in the description are unchanged and remain the direct evidence.

I mention it because it bears on merge order more than on the code: the branch was written up as "the profiled small forms," which undersells it by a wide margin. Nothing about the diff changes — it is still rebased, gated green, and answering your two-interval and call-through points from the last round.

benletchford (changes requested):
The PEA/CMPI/MOVE implementation reviews cleanly against 0.10.12, but this head now conflicts with master after #117, so its passing checks do not validate the prospective integration. Rebase onto current master while preserving #117’s absolute-TST decode, exact cycle table entries, native checked-load and portable execution paths, all focused TST tests, `mem_trace_tst_absolute_matches_step`, and the `tst-abs` benchmark alongside these three forms. Retain PEA’s two-interval store guard and all current CMPI/MOVE/PEA regressions, then rerun the complete required workflow.

rlanday:
From Claude Opus 5:

Whole-app numbers for this PR, plus a merge-ordering observation that may be
worth having before you sequence these.

**Instrument.** EV Override 1.0.1 on `benletchford/systemless`, driven headless
by a scripted input replay: events are scheduled by **retired instruction
count**, not wall time, so the same script drives every run to the same point
regardless of how fast the build is. The script boots the app, dismisses the
shareware dialog, creates a pilot and enters the ship, reaching live flight at
roughly 150M instructions.

**Metric.** `jit_retired / total_instructions` — the share of guest instructions
retired by compiled traces rather than the interpreter — from the
so the denominator is identical by construction.

**Arms.** Separate binaries, identical profiler, distinct hashes verified.
`baseline` is master at 0.10.12, the base this branch was measured on (master
has since moved to 0.10.13 with #117). Every run starts from a pristine copy of
the save directory — the emulator writes save files on exit, and without the
reset the arms drift apart. Runs that did not report `Completed 400000000` are
excluded; one run was killed by the OS at exit and is not counted.

**Noise.** The harness is repeatable but not bit-deterministic: the game seeds
default pilot names, and apparently in-game spawns, from a clock source. That is
what the repeats are for.

| arm | n | mean native share | sd |
|---|---|---|---|
| baseline | 8 | 45.738% | 0.112 |
| **this PR** | 15 | **46.410%** | 0.733 |

Welch's t-test, 95% CI: **+0.673pp, ±0.379, t = +3.48** — clears the noise
floor. Most runs land in a tight 46.5–46.9 band; the sd is inflated by two low
outliers that are genuine in-game variance (both runs were verified to have
reached gameplay), and they are included rather than trimmed.

**Merge-ordering note.** In the blocker ranking, `61FF` (BSR.L) roughly
quadruples once this PR lands — ~0.97M projected dispatches on master, ~4.28M
with this branch — consistently across all runs, with no overlap between the two
distributions. Nothing regresses; the forms this PR admits simply let recordings
run further, and they then strand at the call layer instead of earlier. So after
this merges, the call layer becomes the largest uncovered blocker in the profile,
where before it was mid-table. Worth knowing if you are deciding what to look at
next.

For what it is worth I chased that call-layer mass and it is **nested** calls —
a call inside a call ending the region at the outer callee's boundary — not
anything about call spans. I measured a candidate fix for the sequential-call
case stacked on this branch: it removes its own mechanism completely and moves
whole-app share not at all, so I am not proposing it.

**Scope.** One application, one scripted route, one host. Native share is not
wall-clock time: it says more guest work is being retired by compiled traces,
which is what this PR is for, and does not by itself quantify a frame-rate
change. Happy to publish the harness or the raw per-run reports.

rlanday:
From Claude Fable 5:

Rebased onto current master (0.10.14) per your last review; `HEAD` is the integrated head.

Because git aligned this branch's test functions against #117's TST tests mid-function during a plain rebase, the head was instead rebuilt from master directly: the production diff applied as hunks, and every added test and benchmark re-inserted as a whole definition cut from this branch's own blob. Audited at content level afterwards: zero deleted opcode-literal lines and zero deleted definitions against master; a two-way function-presence audit over `trace_jit.rs`, `microbench.rs`, and `run_batch_tests.rs` shows nothing from master missing (specifically: the absolute-TST decode and cycle entries, native and portable paths, the focused TST tests, `mem_trace_tst_absolute_matches_step`, and the `tst-abs` benchmark) and none of this branch's additions dropped — all three form tests, all three `run_batch` differentials, all three benchmarks with their `pea-abs` / `cmpi-disp` / `move-imm-reg` dispatch arms, and PEA's two-interval store guard with its regressions.

**Full required workflow on `HEAD`:** rustfmt clean; `clippy --all-targets --all-features -D warnings` clean; lib (`--features jit`) 215 passed / 0 failed; default suite 832 / 0; all-features 917 / 0; `cargo doc` clean.

rlanday:
Fresh evidence for this PR from a trace-profile census taken today (EV Override under Systemless, boot → title crawl → flight, current masters plus the pending startup PRs on the Systemless side):

`PEA (xxx).W` (`4878`) is the single largest admission blocker in the game — 1.68 M projected dispatches across the top uncompiled heads in one session — and `CMPI.W #,(d16,An)` (`0C6A`) tops the failed-recordings list. Both are exactly the forms this PR admits. (The census also surfaced `PEA (An)` (`4850`, 870 K) as a natural follow-on in the same family.)

— Claude Fable 5

rlanday:

benletchford:
I’ve reviewed this as a generic JIT admission change and the implementation itself looks sensible. I’ve just merged #111, though, and this branch no longer merges cleanly with current `master`, so I don’t want to force it or treat the old green CI as validation of the conflicted integration.

Please rebase onto current `master`, preserving the MOVEM.L admission, short-window atomic-bail checks, caller/callee code-overlap guards and tests from #111 alongside the PEA abs / CMPI.W d16 / MOVE #imm,Dn work here, then rerun the full Rust CI. Once that integrated head is green I’m happy to re-review.

rlanday:
Rebased onto current `master` (db55299). #111's work is fully preserved alongside this PR's forms — the conflicts were same-anchor additions, resolved as a function-level union: both decode arms chained (`decode_movem_long_trace_op` + `decode_move_imm_reg_trace_op`), both sides' test sets intact (#111's MOVEM.L atomic-bail/window/decode tests and this PR's PEA abs / CMPI.W d16 / MOVE #imm regressions), `bench_movem_frame_loop` restored beside the three new benches, and #111's `movem_long_roundtrip_loop_matches_step` kept in `run_batch_tests`. Local checks on the rebased head: fmt clean, clippy clean, `--features jit` lib 220 pass, `run_batch_tests` 44 pass, all targets compile. — Claude Fable 5

benletchford (approved):
The requested integration work is now present on the current-master head: the MOVEM.L paths and safety guards are preserved alongside the PEA/CMPI/MOVE admissions, and the full CI is green. Approved.

PR #114, opened 2026-08-10
perf: record through constant-target JSRs and make call permission durable

rlanday:
**Stacked on #107** (record-through-BSR with per-segment SMC intervals); the diff over that branch is the two mechanisms here.

## Why

The latest capped-session census of EV Override (Systemless host, `trace-profile`) names two call-machinery gaps left after #107:

- **`4EBA` JSR d16(PC) is the single largest blocked head** — 7.2M projected ops — and `4EB9` JSR (xxx).L rides the same shape. Both targets are decode-time constants, so they are recordable with exactly the BSR bracket machinery.
- **Far-call heads whose permitted retry dies deep in the callee re-probe forever**: the profile shows a ~5:1 wasted-to-permitted recording-attempt cycle (e.g. 7,442 permissionless blocks against 1,479 permitted retries at one head, session after session), because the earned call-through permission lives only in the head's cache slot and every revival path recreates the slot from scratch.

## What

**1. Constant-target JSRs record through.** `JSR d16(PC)` and `JSR (xxx).L` decode into the same `CallThrough`/callee/`RtsReturn` bracket as BSR — the push is a constant return address, execution follows the jump. `CallThrough` now carries a per-form cycle charge (BSR and JSR d16(PC) 18, JSR (xxx).L 20), and the retry gate matches all recordable calls through one shared predicate.

**2. Earned permission survives slot churn.** A lossy direct-mapped side table (same index function and size as the trace cache) remembers heads whose recordings blocked at a recordable call; fresh candidacy — from backward-branch probes after a cache-collision stomp — consults it. A revived head therefore goes straight to its permitted recording instead of re-earning the bit through a doomed probe attempt. Losing a table entry to collision costs exactly the one blocked attempt that was previously the universal price, so the table can only reduce work. This also composes with prefix salvage (#109): under the retry-outranks-salvage rule, only permitted attempts may salvage, so heads that used to burn attempt after unsalvageable attempt now reach a salvageable permitted recording immediately.

## Measured effect

**Stack-delta headless A/B** — base is the #107 branch itself, so the numbers isolate exactly this diff (deterministic 900M-instruction EV Override boot workload, counters from the JIT):

| counter | base (#107) | cand (this branch) |
|---------|-------------|---------------------|
| native_calls | 68,054 | 68,126 (+72) |
| jit_retired | 16,375,538 | 16,376,306 (+768) |
| rejected_hits | 524,729 | 524,568 (−161) |

Small and positive: the boot workload carries a handful of constant-target JSR heads and modest permission savings. The named masses are gameplay-scene: in the following capped session with this branch bundled, the `4EBA` class emptied from the blocked-head table entirely (it had been the #1 single head), and the new walls' heads all show `attempts=2` — no re-probe churn — where the same session's pre-table revival path (exit seeding, whose candidacy arm predates this branch) still showed thousands of wasted attempts, isolating precisely the paths the table does and does not reach. The bundle composition note below closes that last path.

## Validation

- `jsr_constant_targets_record_through_on_retry` — both JSR forms drive the full blocked-then-retried lifecycle end to end, with branch displacements computed from the program layout, and compile with the call and checked return inline.
- `earned_call_permission_survives_a_slot_stomp` — the durable-permission contract: permission earned, slot stomped by a deliberately-colliding head, original head revived; fresh candidacy carries the bit, and a head that never earned it starts without it.

**Composition note**: exit seeding (#101) creates candidacy through its own path; when both land, its fresh-candidacy arm should also consult the permission table (a one-line composition, field-verified in the bundled session above).

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -666,10 +666,12 @@ pub(crate) struct TraceJit {
     /// Heads that have earned call-through permission: a recording here
     /// blocked at a recordable call, so future candidacy at the same pc
     /// starts with permission instead of re-earning it through a doomed
-    /// probe attempt. Direct-mapped and lossy: a collision drops a bit,
-    /// which only costs the one extra blocked attempt that was the
-    /// universal price before this table existed.
-    earned_call_permission: Vec<u32>,
+    /// probe attempt. Two-way per index: the gameplay profile shows hot
+    /// call-heads landing in the same slot (two of its worst cyclers
+    /// collide), and one spare way absorbs exactly that. Still lossy
+    /// beyond two -- losing an entry only costs the one extra blocked
+    /// attempt that was the universal price before the table existed.
+    earned_call_permission: Vec<[u32; 2]>,
 }
 
 impl fmt::Debug for TraceJit {
@@ -703,7 +705,7 @@ impl TraceJit {
             next_func: 0,
             slots: (0..TRACE_CACHE_SIZE).map(|_| TraceSlot::Empty).collect(),
             recording: None,
-            earned_call_permission: vec![u32::MAX; TRACE_CACHE_SIZE],
+            earned_call_permission: vec![[u32::MAX; 2]; TRACE_CACHE_SIZE],
         }
     }
 
@@ -1128,11 +1130,25 @@ impl TraceJit {
     }
 
     fn grant_call_permission(&mut self, pc: u32) {
-        self.earned_call_permission[trace_cache_index(pc)] = pc;
+        let ways = &mut self.earned_call_permission[trace_cache_index(pc)];
+        if ways[0] == pc || ways[1] == pc {
+            return;
+        }
+        if ways[0] == u32::MAX {
+            ways[0] = pc;
+        } else if ways[1] == u32::MAX {
+            ways[1] = pc;
+        } else {
+            // Both ways live: rotate so persistent pairs survive and a
+            // third colliding head still makes progress eventually.
+            ways[1] = ways[0];
+            ways[0] = pc;
+        }
     }
 
     fn has_call_permission(&self, pc: u32) -> bool {
-        self.earned_call_permission[trace_cache_index(pc)] == pc
+        let ways = &self.earned_call_permission[trace_cache_index(pc)];
+        ways[0] == pc || ways[1] == pc
     }
 
     fn record_trace_target(&mut self, pc: u32, cpu_type: CpuType) {
@@ -15071,5 +15087,11 @@ mod portable_tests {
                 ..
             } if *pc == COLLIDER
         ));
+        // Two colliding call-heads both keep their permission: the
+        // gameplay profile's two worst cyclers share one index, and the
+        // second table way exists for exactly that pair.
+        jit.grant_call_permission(COLLIDER);
+        assert!(jit.has_call_permission(HEAD));
+        assert!(jit.has_call_permission(COLLIDER));
     }
 }
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -1028,12 +1028,18 @@ impl TraceJit {
             }
             if rerecord_dominant_path {
                 let adaptive_rerecords = trace.adaptive_rerecords.saturating_add(1);
+                // A slot recreated from scratch must consult the durable
+                // grant, or a head that already earned permission pays a
+                // fresh permissionless blocker every time its slot is
+                // replaced -- which is exactly what the side table exists
+                // to prevent.
+                let allow_call_through = self.has_call_permission(pc);
                 self.slots[idx] = TraceSlot::Counting {
                     pc,
                     cpu_type,
                     hits: 0,
                     adaptive_rerecords,
-                    allow_call_through: false,
+                    allow_call_through,
                 };
                 cpu.trace_record_skip = [TRACE_PC_NONE; 4];
                 cpu.trace_probe_skip = [TRACE_PC_NONE; 4];
@@ -1197,6 +1203,10 @@ impl TraceJit {
         entry_watched: bool,
     ) -> ExitSeed {
         let idx = trace_cache_index(exit_pc);
+        // Read the durable grant before borrowing the slot mutably: a
+        // candidate created here must carry the permission its head
+        // already earned.
+        let permission = self.has_call_permission(exit_pc);
         match &mut self.slots[idx] {
             TraceSlot::Compiled(CompiledTrace {
                 pc: compiled_pc,
@@ -1250,7 +1260,7 @@ impl TraceJit {
                     cpu_type,
                     hits: 1,
                     adaptive_rerecords: 0,
-                    allow_call_through: false,
+                    allow_call_through: permission,
                 };
                 ExitSeed::None
             }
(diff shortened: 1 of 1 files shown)


rlanday:
From Claude Fable 5:

One field-driven addition (`HEAD`): the permission table gains a second way per index. The following capped session showed the two worst re-probe cyclers hashing to the same slot — each grant stomped the other's entry, so both heads kept paying the permissionless attempt. One spare way absorbs a colliding pair exactly (the stomp-survival test now covers the pair); beyond two the table stays lossy by design. All suites green on the new head.

rlanday:
From Claude Fable 5:

Field verification of the second way (`e778de5`) from the follow-up capped session, as promised — result is split by head, and honestly reported as such:

- **`00EFCC92` (the top re-probe cycler, one of the colliding pair): cured of the permissionless block.** It is out of the rejected-heads top-forty entirely — every attempt now records through its BSR.L. Its failure moved *into* the callee: the helper at `00F04BFC` contains a 128-iteration `ADDQ/CMPI/BLT` counting loop, and recording it exhausts the op budget before any trace terminal appears (rank-1 silent rejection, ~4.1K records, `no-trace-terminal`). That is a recording-cap/loop-shape problem the permission table was never going to fix — a different mechanism's work.
- **`00EE8C92` (its collision partner): mixed.** It now has three compiled traces, and silent recordings that pass the first `61FF` and reach a *second* one — but it still shows ~27.6K rejected hits with the first call as blocker. Its deep paths end at A-line traps with no trace terminal anywhere, so this head may be structurally untraceable regardless of permissions.

Net: the collision-stomp mechanism this commit targets is confirmed fixed (the head that was losing its grant no longer re-probes permissionless), and the residual churn is now attributable to two named non-permission shapes (callee-inner-loop budget exhaustion; terminal-free trap-bound paths). Whole-session ratios moved the right way (record share 55.1% → 58.7%) but the bundle also carried the CLR admission from #116, so no whole-session claim is made for this change alone.

benletchford (changes requested):
This stacked PR still contains #107's CPU-model decode bug: its retry path treats `BSR` low byte `0xFF` as a long displacement on M68000/M68010/SCC68070, where it is the short −1 form, so it can add non-operand bus reads and record the wrong return PC. The constant-target JSR increment cannot merge while that base correctness issue is present. Please incorporate #107's requested CPU-aware decode and counting/faulting-bus regression, then rebase the complete stack onto current master (including #109's salvage logic and coverage) and rerun all required checks.

rlanday:
From Claude Opus 5:

Restacked as requested (`HEAD`): this branch now sits on the corrected #107, which carries the CPU-aware BSR decode (`0xFF` is the long form only on 68020+) and its panicking/counting-bus regression, and the whole stack is rebased onto current master (0.10.7) with #109's salvage machinery preserved.

Verification before pushing:

- both of this branch's own commits replay with zero conflicts, and every non-trivial line they added is present in the restacked tree (whitespace-flattened comparison) — 0 missing;
- the base correctness fix you asked for is present in the stack (`0xFF if !is_pre_68020(cpu_type)`), as is this branch's two-way permission table;
- full suite on the restacked head: default 815/0 (including the Musashi/SST fixtures), lib jit-only 190/0, all-features 216/0, run_batch integration 35/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

The field evidence posted earlier for the second permission way still applies — it was measured on a bundle containing this exact change, and the restack does not alter it.

benletchford (changes requested):
This stacked head inherits two unresolved blockers from #107:

- The native and portable call-through push/pop paths can wrap their `fm_len - 4` bounds arithmetic for nonempty FastMem windows of length 1–3, allowing a four-byte raw access beyond the advertised window. Guard short windows before subtraction and cover both push and pop bails.
- The promised retry-before-salvage composition is not present. A first permissionless BSR/JSR blocker can be salvaged into `Compiled` before the rearm logic runs. Although this PR grants the durable permission bit, the compiled salvage remains installed, so the immediate permitted retry does not happen and the trace does not record through the call until unrelated slot churn. Add a salvageable-prefix-before-call regression and make the first call blocker install/rearm the permitted attempt before salvage is considered.

Please resolve these in the base behavior and rebase this stacked PR afterward.

benletchford (commented):
Correction to the FastMem portion of my review: the public `run_batch` entry point rejects windows shorter than four bytes before populating the crate-private `fm_*` scratch state, so the subtraction is protected by the capture invariant in reachable library use. I withdraw that blocker.

The retry-before-salvage issue remains and still blocks this stacked head: granting a durable permission after the first permissionless call does not replace the salvaged `Compiled` slot with the immediate permitted recording promised by the design.

rlanday:
From Claude Opus 5:

Both inherited blockers are resolved in the base and this head is rebased onto it.

- **Retry before salvage** is fixed in #107: `finish_recording` skips the prefix trim when a recording stops at a call the head lacks permission for, so the head rejects and the rearm installs the permitted recording. Because this branch computes the flag from `is_recordable_call`, the precedence covers its JSR blockers as well as BSR. The regression `a_salvageable_prefix_before_a_call_still_yields_the_permitted_retry` builds a genuinely salvageable prefix (interior branch, then non-terminal ops) before the call and asserts the compiled trace contains a `CallThrough`; removing the precedence makes it fail with the eight-op salvaged prefix you described.
- **FastMem short window**: withdrawn on your own follow-up, so nothing added here.

On the interaction you flagged specifically for this branch — that granting the durable permission bit does not replace an already-installed salvaged `Compiled` slot — that is exactly right, and the fix is upstream of the grant rather than in it: with the trim suppressed, no salvaged slot is installed at that blocker in the first place, so the permitted retry happens immediately instead of waiting on slot churn.

Full suite on the restacked head: default 815/0 (including the Musashi/SST fixtures), lib jit-only 191/0, all-features 217/0, run_batch integration 35/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean; #109's salvage machinery present and green throughout.

The two-way permission table's field evidence posted earlier was measured on a bundle predating this fix. Since the fix changes when heads earn permission at all, I plan to re-measure it on a session build that includes both, and will post the updated numbers rather than let the old ones stand for the new behavior.

benletchford (changes requested):
The inherited retry fix still misses a branch-ended prefix. When an unpermitted BSR or JSR is the immediate target of the last recorded branch, the last op already ends the trace. Skipping the salvage trim does not prevent `finish_recording_with_retry` from compiling that branch as an ordinary terminal and installing `Compiled`, so the rearm's `Rejected` check fails and durable permission never gets an immediate retry.

Please make a pending call retry bypass prefix compilation entirely, add a direct branch-to-recordable-JSR regression (in addition to the base BSR case), and restack/rebase the full PR onto current master after the base fix.

rlanday:
From Claude Opus 5:

Both points addressed, and the stack is rebased onto current master (0.10.9).

- **Retry bypasses prefix compilation** in the base (#107): the region is rejected before compilation whenever a call retry is pending, so a branch-ended prefix can no longer be installed ahead of the rearm. The trim-suppression clause is gone as unreachable.
- **Direct branch-to-recordable-JSR regression** added here (`a_branch_landing_on_a_jsr_still_yields_the_permitted_retry`), alongside the base's BSR cases, since this branch is what admits constant-target JSR as a recordable call. Like its siblings it inspects the head's slot as the run proceeds rather than at the end — an end-state assertion passes even when the defect is present, because slot churn eventually replaces the call-less trace. Mutation-checked: removing the bypass fails all three retry regressions, including the JSR one specifically with "starving the permitted JSR retry".

The restack is from this branch's true base rather than the older tip, so the stale duplicate of the first retry commit is gone and the history is two commits on top of the corrected #107.

Full suite on the restacked head: default 820/0 including the Musashi/SST fixtures, lib jit-only 202/0, all-features 228/0, run_batch integration 38/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean; #109's salvage machinery and #116's absolute CLR both present and green.

As noted before, the two-way permission table's field evidence predates this fix. Since the fix changes when heads earn permission at all, I will re-measure on a session build carrying both rather than let the old numbers stand.

benletchford (changes requested):
The durable-permission integration is incomplete on current master. `record_trace_target` consults `has_call_permission(pc)`, but the fresh-slot arm of `note_trace_exit` still creates `Counting { allow_call_through: false }`; the adaptive-rerecord arm in `try_execute` does the same. A head that earned permission and then revives through a guarded exit (or is adaptively rerecorded) therefore pays another permissionless blocker despite the side-table grant. This is the composition case the PR body anticipated for #101, which is now in master. Route every new/recreated `Counting` slot through the durable grant (while retaining the existing-slot permission behavior), and add a regression that grants a head, stomps its cache slot, revives it via `note_trace_exit`, and observes `allow_call_through: true`.

rlanday:
From Claude Opus 5:

Fixed exactly as described, and you were right that this is the composition case the body anticipated for exit seeding.

Both paths that build a `Counting` slot from scratch now consult the durable grant: the fresh-slot arm of `note_trace_exit` (the grant is read before the slot is borrowed mutably, so the candidate carries it) and the adaptive-rerecord arm of `try_execute`. The existing-slot behaviour is unchanged — a slot that already exists keeps whatever permission it holds.

The regression is the one you specified (`a_revived_head_carries_its_durable_call_permission`): grant a head, stomp its cache slot to `Empty` as a colliding head would, revive it through `note_trace_exit`, and observe `allow_call_through: true`. It also asserts the negative — a head with no grant is still revived *without* permission, so the test cannot pass by making permission unconditional. Mutation-checked: restoring `allow_call_through: false` on the revival path fails it.

Rebased onto current master (0.10.10) after #115 merged, with the two-way audit (branch content 0 missing; master content 0 missing apart from the lines this branch deliberately changes).

Full suite on the pushed head: default 823/0 including the Musashi/SST fixtures, lib jit-only 205/0, all-features 231/0, run_batch integration 39/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean, `cargo fmt --all -- --check` clean.

benletchford (changes requested):
Now that #107 is merged, this branch must be restacked onto current master so the PR contains only the constant-target JSR and durable-permission work, rather than a second copy of the BSR call-through implementation. Preserve the distinct JSR cycle counts and the durable grant lookup in both exit-seeded slot creation and adaptive re-recording, including the stomp/revival regression tests, then rerun the full required workflow.

rlanday:
From Claude Opus 5:

Restacked onto current master now that #107 is merged. The branch is four commits and contains only this PR's work — the constant-target JSR admission, the two-way permission table, the JSR retry regression, and the durable-grant fix — with the BSR call-through implementation now coming from master rather than a second copy. Master's squash meant patch-id could not drop my copies automatically, so the restack is explicitly `--onto` master from this branch's own first commit.

Preserved and verified on the restacked head:

- the distinct JSR cycle counts (18 for `BSR`/`JSR (d16,PC)`, 20 for `JSR (xxx).L`);
- the durable-grant lookup in **both** slot-creation paths — the exit-seeded fresh slot and the adaptive re-record — with the grant read before the slot is borrowed;
- the stomp/revival regression (`a_revived_head_carries_its_durable_call_permission`), including its negative case so it cannot pass by making permission unconditional, plus the two-way permission table's own stomp-survival test.

Full suite on the restacked head: default 823/0 including the Musashi/SST fixtures, lib jit-only 205/0, all-features 231/0, run_batch integration 39/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean, `cargo fmt --all -- --check` clean.

benletchford (approved):
The restacked head is now focused on the constant-target JSR and durable-permission delta. JSR d16(PC) and absolute-long return widths and cycle charges are correct; checked stack writes and RTS mismatch bails preserve atomicity; and normal, exit-seeded, and adaptive slot recreation all carry the two-way durable grant. The prospective merge tree passed formatting, all-target/all-feature Clippy, default and all-feature tests with doctests, and package verification; the live CI and title checks are also green.

PR #115, opened 2026-08-10
perf: admit AND and OR from memory into the register-ALU family

rlanday:
## Why

The latest capped-session census of EV Override (Systemless host, `trace-profile`) names `AND.W (d16,An),Dn` as the widest blocked class in the game — **twenty-three of the top forty recording heads** stop at `C270`-family opcodes. The memory-ALU family (`AluMemToReg`) already admits CMP/ADD/SUB from measured EAs; its decoder's op match simply never included And/Or, and the interpreter's fast path already decodes both.

## What

`AND <ea>,Dn` and `OR <ea>,Dn` join the family for its existing EA set — `(An)`, `(d16,An)`, brief-indexed. Both emit with logic flags (NZ from the sized result, VC cleared, X preserved) and follow the family's conservative cycle-maxima convention. EOR has no memory-to-register form on the 68k and is not involved.

## Measured effect

**Focused microbench** (filter `memory-and`): the census exemplar — AND.W of a displaced structure field plus a counter — 100M instructions per run after a 5M warmup, deterministic `run_batch` workload; both arms `--features jit` on the same idle machine; base is `origin/master` (0.10.6) with an identical benchmark copy; five alternating runs, median of pairwise ratios:

| run | base | cand |
|-----|------|------|
| 1 | 45.9 | 351.7 |
| 2 | 47.9 | 362.8 |
| 3 | 48.4 | 374.4 |
| 4 | 49.3 | 374.4 |
| 5 | 50.7 | 381.6 |

Median pairwise **7.59×** (M instr/s; base rejects the head at the AND and interprets).

**Whole-application headless A/B** (deterministic 900M-instruction EV Override boot workload): flat — native_calls 68,054 → 68,054, jit_retired within 30 of 16.3757M. The boot workload contains no AND-from-memory heads; the twenty-three blocked heads are gameplay-scene mass, on the same acceptance pattern as #108/#111/#112/#113 — a bundled session with this branch is the follow-up.

## Validation

- `memory_and_or_match_the_interpreter_with_exact_cycles` — a step() differential over the census exemplar (indexed and displaced AND, displaced OR) on a 68000: exact registers, exact logic flags with X preserved, and the family's cycle bound asserted never to undercharge the true 68000 count.
- `native_memory_and_or_match_portable` — native and portable agree on the merged word results, flags, retired count, and cycle charge.
- `memory_and_loop_matches_step` (integration) — the exemplar shape matches single-step interpretation exactly through the whole lifecycle.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1111,6 +1111,7 @@ fn bench_salvaged_prefix_loop() {
     for (index, word) in words.iter().enumerate() {
         bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
     }
+
     let prepare_cpu = || {
         let mut cpu = CpuCore::new();
         cpu.set_cpu_type(CpuType::M68040);
@@ -1821,8 +1822,8 @@ fn main() {
         bench_link_unlk_frame_loop();
         return;
     }
-    if only.as_deref() == Some("clr-abs") {
-        bench_clr_abs_loop();
+    if only.as_deref() == Some("memory-and") {
+        bench_memory_and_loop();
         return;
     }
     if only.as_deref() == Some("salvage-prefix") {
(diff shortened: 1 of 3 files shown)


rlanday:
From Claude Fable 5:

The bundled gameplay session promised in the PR body has now run, with this branch included. Head-level result: **every AND-from-memory blocker is gone from the census.**

**Instrument.** Each capped GUI session of EV Override (Systemless host, `trace-profile` build, 25 MHz cap, ~2.5 min of live gameplay) prints a ranked table at exit: the top forty rejected recording heads by rejected-hit count, each with the opcode that stopped its recording. The PR body's "twenty-three of the top forty" motivation was counted on this table from the prior session (bundle without this branch). The follow-up session ran the same bundle family plus this branch (and one unrelated recorder-side change, noted below).

**Arms, same table:**

| | prior session (no AND/OR) | this session (with AND/OR) |
|---|---|---|
| heads blocked at `C270`-family AND | **23 / 40** | **0 / 40** |
| heads blocked at MULU/MULS (`C2E8`/`C2F0`/`C0E9`/`C3F0` — out of scope here) | 3 | 6 |

The only `C270` appearing anywhere in the new session's dump is *inside* a recorded path (`00F08E60: C270/0800`, mid-prefix of a head that now records through the AND and fails later for an unrelated reason — no trace terminal). The vacated rows are occupied by the next classes: `4239` CLR-abs (6 heads), `61FF` BSR.L callees, `2D7C` immediate-store.

**Whole-session movement, with caveats.** Average retired ops per native call rose 98.2 → 122.6, and the record share of executed instructions rose 48.8% → 55.1%. Two honest qualifiers: sessions are live gameplay, so they are not instruction-count-matched and absolute totals are not comparable; and the session bundle also carried one recorder-side change (call-permission composition, PR #114), so the whole-session ratios are not attributable to this branch alone. The head-table membership above is the form-specific signal — the 23 AND heads can only leave that table because the admission landed.

Raw session logs are retained locally and I'm happy to attach either table in full.

benletchford (changes requested):
The AND/OR memory-to-register extension and its flag/parity coverage passed the full local CONTRIBUTING suite on the reviewed head, but this branch now conflicts with `master` after #109. Please rebase onto current master while preserving #109's blocker-only prefix salvage in `finish_recording`, all three salvage unit cases, `salvaged_prefix_loop_matches_step`, and the `salvage-prefix` benchmark alongside the AND/OR additions, then rerun the complete checks on the integrated result.

rlanday:
From Claude Opus 5:

Full suite on the rebased head: default 813/0 (including the Musashi/SST fixtures), lib jit-only 182/0, all-features 208/0, run_batch integration 34/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean.

rlanday:
From Claude Opus 5:

| run | base | cand |
|-----|------|------|
| 1 | 50.4 | 329.5 |
| 2 | 47.8 | 312.4 |
| 3 | 46.6 | 304.7 |
| 4 | 46.0 | 303.8 |
| 5 | 45.8 | 294.1 |

benletchford (commented):
Current `master` now includes #101 and #116, and this head conflicts in the trace-JIT integration. Please rebase while preserving continuation exit seeding/chaining and the absolute-CLR decode, native/portable execution, cycle tests, run-batch regression, and benchmark alongside the AND/OR memory-source additions; then rerun the full repository checks.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.9), preserving #101's continuation exit seeding/chaining and #116's absolute-CLR decode, native and portable execution, cycle coverage, run-batch regression, and benchmark alongside the AND/OR memory-source additions.

Full suite on the rebased head: default 818/0 (including the Musashi/SST fixtures), lib jit-only 191/0, all-features 217/0, run_batch integration 37/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

benletchford (approved):
Reviewed the complete diff and surrounding trace/JIT paths. AND/OR use the existing checked memory-to-register family, preserve sized register semantics and X, update NZVC correctly, and retain conservative cycle headroom. Focused native/portable and run_batch regressions are adequate. On head 093cc078, formatting, strict all-target/all-feature Clippy, default and all-feature tests/doctests, and package verification all pass locally; required GitHub checks are green.

PR #116, opened 2026-08-10
perf: admit absolute-addressed CLR into the memory-CLR family

rlanday:
## Why

The latest capped-session census of EV Override (Systemless host, `trace-profile`) puts absolute-addressed CLR next in line: **six of the top forty rejected recording heads** (ranked by rejected hits) block at `4239` — `CLR.B` to an absolute long address. The admission covers the whole absolute sub-family (`.B/.W/.L`, `(xxx).W` and `(xxx).L`), since the memory-CLR family already handles indexed and predecrement destinations and the absolute forms need only a constant-address arm.

## What

`CLR.B/W/L (xxx).W` and `(xxx).L` join the family. Three pieces beyond the decode arms:

- **Exact cycle entries.** The family's fallback arm would have *undercharged* the absolute forms (abs.W by 2, abs.L by 6 cycles). The new entries are exact — 16/24 for abs.W, 20/28 for abs.L (word-or-byte/long) — and are pinned against the 68000 step interpreter by test, not just against the timing table.
- **Constant-address emit.** The address is an `iconst`; the store passes the same SMC guard as every family store, so a CLR aimed at the trace's own code bails atomically on both engines.
- **Portable delegation.** The portable executor maps the absolute forms onto the interpreter's decoded-EA path, with the same pre-store code-range check.

## Measured effect

**Focused microbench** (filter `clr-abs`): all three widths against fixed absolute targets in a DBRA loop, 210M instructions per run after a 5M warmup, deterministic `run_batch` workload; both arms `--features jit` on the same idle machine; base is `origin/master` (0.10.6) with an identical benchmark copy; five alternating runs, median of pairwise ratios:

| run | base | cand |
|-----|------|------|
| 1 | 35.6 | 262.8 |
| 2 | 37.5 | 246.4 |
| 3 | 39.6 | 281.7 |
| 4 | 43.9 | 296.5 |
| 5 | 44.5 | 304.4 |

Median pairwise **6.84×** (M instr/s; base rejects the head at the CLR and interprets).

**Whole-application headless A/B** (deterministic 900M-instruction EV Override boot workload): exactly flat — every total identical between arms (backward_hits 553,701; rejected_hits 524,755; native_calls 68,054; jit_retired 16,375,726). The boot workload contains no absolute-CLR heads; the six blocked heads are gameplay-scene mass, on the same acceptance pattern as #108/#111/#112/#113/#115 — a bundled session with this branch is the follow-up.

## Validation

- `clr_to_absolute_decodes_with_exact_extents` — abs.W sign-extension (negative addresses), two-word abs.L assembly, and extension-word bookkeeping.
- `native_clr_to_absolute_matches_portable_and_bails_atomically` — all three widths: native and portable agree on registers, flags (Z set, N/V/C cleared, X preserved), memory, retired count, and cycle charge; a CLR aimed at the trace's own extension word bails on both paths with nothing committed.
- `clr_to_absolute_cycles_match_the_interpreter_on_a_68000` — one pass over all three widths compiled for a 68000 charges exactly what four step() calls charge.
- `mem_trace_clr_absolute_matches_step` (integration) — the whole record/compile/execute lifecycle with a re-dirtied target matches single-step interpretation exactly.
- Mutation-checked: removing the exact cycle arms fails the cycle pin (60 vs 74), removing the portable delegation fails parity, removing the sign-extension fails decode.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1023,6 +1023,56 @@ fn bench_link_unlk_frame_loop() {
     );
 }
 
+/// Exercise the absolute-addressed CLR forms the gameplay census
+/// flagged (4238/4239 heads): all three widths against fixed targets.
+fn bench_clr_abs_loop() {
+    // Outer cycles of 514 instructions end exactly at the outer label
+    // with the three targets cleared and the guard byte untouched.
+    const CYCLES: u32 = 408_560;
+    const INSTRS: u32 = CYCLES * 514;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x727F, // outer: MOVEQ #127,D1
+        0x4278, 0x4000, // inner: CLR.W ($4000).W
+        0x42B9, 0x0000, 0x4004, // CLR.L ($4004).L
+        0x4239, 0x0000, 0x4003, // CLR.B ($4003).L
+        0x51C9, 0xFFEE, // DBRA D1,inner
+        0x60E8, // BRA.S outer
+    ];
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    for address in 0x4000..0x4008 {
+        bus.write_byte(address, 0xAA);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.pc, CODE_BASE, "the run ends exactly at the outer label");
+    for address in (0x4000..0x4002).chain(0x4003..0x4008) {
+        assert_eq!(bus.read_byte(address), 0, "target byte cleared");
+    }
+    assert_eq!(bus.read_byte(0x4002), 0xAA, "untouched byte kept");
+    println!(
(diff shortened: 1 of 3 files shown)


rlanday:
From Claude Fable 5:

The bundled gameplay session promised in the PR body has now run with this branch included (same instrument as before: capped GUI session of EV Override, `trace-profile` build, ranked tables printed at exit).

**Head-level: all six absolute-CLR blockers are gone.** The prior session's top-forty rejected-heads table had six heads blocked at `4239`; this session's has zero, and `4239` ops now appear *inside* recorded prefixes (e.g. `00EFA8A0: 4239/00E7/AA61`, and the `00EFEF52` head from the prior census recording straight through its CLR).

**Whole-session movement, with the usual caveats:** record share of executed instructions rose 55.1% → **58.7%**, and the rejected-to-backward-hit ratio improved 90.4% → 81.5%. Sessions are live gameplay (not instruction-matched), and the bundle also carried one recorder-side change (the second permission-table way from #114), so the ratios are not attributable to this branch alone — the head-table membership above is the form-specific signal. Average ops per native call moved 122.6 → 95.6, which is composition, not regression: the newly converted CLR heads are short shapes joining the pool.

benletchford (changes requested):
The absolute CLR decode, checked-store behavior, flags, and cycle/parity coverage passed the full local CONTRIBUTING suite on the reviewed head, but this branch now conflicts with `master` after #109. Please rebase onto current master while preserving #109's blocker-only prefix salvage in `finish_recording`, all three salvage unit cases, `salvaged_prefix_loop_matches_step`, and the `salvage-prefix` benchmark alongside the absolute-CLR additions, then rerun the complete checks on the integrated result.

rlanday:
From Claude Opus 5:

Full suite on the rebased head: default 813/0 (including the Musashi/SST fixtures), lib jit-only 183/0, all-features 209/0, run_batch integration 34/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean.

rlanday:
From Claude Opus 5:

| run | base | cand |
|-----|------|------|
| 1 | 50.4 | 329.5 |
| 2 | 47.8 | 312.4 |
| 3 | 46.6 | 304.7 |
| 4 | 46.0 | 303.8 |
| 5 | 45.8 | 294.1 |

benletchford (approved):
The current head preserves the salvage machinery and the absolute-CLR implementation remains correct for address decoding, access width, SMC bail-before-commit behavior, flags, and cycle accounting. I validated the exact clean merge tree with current 0.10.8 master: formatting, strict all-target/all-feature Clippy, default and all-feature tests, strict private-item docs, and package verification all pass.

PR #117, opened 2026-08-10
perf: admit absolute-addressed TST into the memory-TST family

rlanday:
## Why

The latest capped-session census of EV Override (Systemless host, `trace-profile`) — the first with the absolute-CLR heads cleared — surfaces the next absolute-EA class behind them: **four of the top forty rejected recording heads** (ranked by rejected hits) block at absolute-addressed TST, three at `4A39` (`TST.B (xxx).L`) and one at `4A79` (`TST.W (xxx).L`), together ~460K projected instructions. The memory-TST family already handles indexed sources; the absolute forms need only a constant-address arm.

Two census neighbours are explicitly *not* in this PR: `0C79` (`CMPI.W #imm,(xxx).L`, three extension words) and `33F9` (`MOVE.W (xxx).L,(xxx).L`, four) exceed the recorder's two-extension-word shape and stay deferred with `2D7C` until that structural change happens.

## What

`TST.B/W/L (xxx).W` and `(xxx).L` join the family. TST reads and writes nothing, so no store guard is involved — the pieces are:

- **Exact cycle entries.** The family's flat charge was the indexed cost (14/18) and would have *overcharged* the absolute forms; the new entries are exact — 12/16 for abs.W, 16/20 for abs.L (word-or-byte/long) — and pinned against the 68000 step interpreter by test.
- **Constant-address emit** through the same window-checked load path as the indexed form.
- **Portable delegation** onto the interpreter's decoded-EA path.

## Measured effect

**Focused microbench** (filter `tst-abs`): all three widths against fixed absolute targets in a DBRA loop, 210M instructions per run after a 5M warmup, deterministic `run_batch` workload; both arms `--features jit` on the same idle machine; base is `origin/master` (0.10.6) with an identical benchmark copy; five alternating runs, median of pairwise ratios:

| run | base | cand |
|-----|------|------|
| 1 | 33.3 | 390.1 |
| 2 | 35.5 | 404.3 |
| 3 | 35.5 | 429.5 |
| 4 | 36.4 | 432.8 |
| 5 | 36.6 | 437.5 |

Median pairwise **11.89×** (M instr/s; base rejects the head at the TST and interprets — and TST's native form is a guard-free load, hence the larger ratio than the store families).

**Whole-application headless A/B** (deterministic 900M-instruction EV Override boot workload): exactly flat — every total identical between arms (backward_hits 553,701; rejected_hits 524,755; native_calls 68,054; jit_retired 16,375,726). The boot workload contains no absolute-TST heads; the four blocked heads are gameplay-scene mass, on the same acceptance pattern as #108/#111/#112/#113/#115/#116 — a bundled session with this branch is the follow-up.

## Validation

- `tst_from_absolute_decodes_with_exact_extents` — abs.W sign-extension (negative addresses), two-word abs.L assembly, extension-word bookkeeping.
- `native_tst_from_absolute_matches_portable_with_exact_flags` — all three widths with the flag matrix pinned per case (negative word → N; zero long → Z; positive byte → neither; X preserved throughout); registers, memory, and charge agree between engines, and memory is bit-identical before and after (TST writes nothing).
- `tst_from_absolute_cycles_match_the_interpreter_on_a_68000` — one pass over all three widths compiled for a 68000 charges exactly what four step() calls charge.
- `mem_trace_tst_absolute_matches_step` (integration) — the whole record/compile/execute lifecycle with the tested word re-dirtied every pass matches single-step interpretation exactly.
- Mutation-checked: removing the exact cycle arms fails the cycle pin (56 vs 58), removing the portable delegation fails parity, removing the sign-extension fails decode.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1818,6 +1818,55 @@ fn bench_set<B: BenchBus>(label: &str) {
     );
 }
 
+/// Exercise the absolute-addressed TST forms the gameplay census
+/// flagged (4A39/4A79 heads): all three widths against fixed targets.
+fn bench_tst_abs_loop() {
+    // Outer cycles of 514 instructions end exactly at the outer label
+    // with the targets untouched (TST writes nothing).
+    const CYCLES: u32 = 408_560;
+    const INSTRS: u32 = CYCLES * 514;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x727F, // outer: MOVEQ #127,D1
+        0x4A78, 0x4000, // inner: TST.W ($4000).W
+        0x4AB9, 0x0000, 0x4004, // TST.L ($4004).L
+        0x4A39, 0x0000, 0x4003, // TST.B ($4003).L
+        0x51C9, 0xFFEE, // DBRA D1,inner
+        0x60E8, // BRA.S outer
+    ];
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    for address in 0x4000..0x4008 {
+        bus.write_byte(address, 0xAA);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.pc, CODE_BASE, "the run ends exactly at the outer label");
+    for address in 0x4000..0x4008 {
+        assert_eq!(bus.read_byte(address), 0xAA, "TST writes nothing");
+    }
+    println!(
+        "batch     absolute TST loop       {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
 fn main() {
     println!("m68k microbench");
     let only = std::env::args().nth(1);
@@ -1895,6 +1944,10 @@ fn main() {
     }
     if only.as_deref() == Some("far-call-through") {
         bench_far_call_through_leaf();
+        return;
+    }
+    if only.as_deref() == Some("tst-abs") {
+        bench_tst_abs_loop();
         return;
     }
     if only.as_deref() == Some("clr-abs") {
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -2750,13 +2750,14 @@ impl JitTraceOp {
             Self::CmpiWordMem { .. } => 18,
             // TST is a four-cycle operation plus the indexed EA read
             // (M68000UM); byte and word reads have the same EA cost.
-            Self::TstMem { size, .. } => {
-                if size == Size::Long {
-                    18
-                } else {
-                    14
-                }
-            }
+            Self::TstMem { size, src } => match (size, src) {
+                (Size::Long, JitEa::AbsWord(_)) => 16,
+                (_, JitEa::AbsWord(_)) => 12,
+                (Size::Long, JitEa::AbsLong(_)) => 20,
+                (_, JitEa::AbsLong(_)) => 16,
+                (Size::Long, _) => 18,
+                _ => 14,
+            },
             // CLR is the M68000UM store cost (8, or 12 for long) plus the
             // EA calculation: indexed as in the read above, predecrement
             // per Table 8-2.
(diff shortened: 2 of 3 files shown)


rlanday:
From Claude Fable 5:

The bundled gameplay session promised in the PR body has now run with this branch included (same instrument: capped GUI session, `trace-profile`, ranked tables at exit).

**Head-level: all four absolute-TST blockers are gone** — no `4A39` or `4A79` rows remain in the top-forty rejected-heads table. The interesting part is *where those heads went*, and it is exactly what this PR's "two census neighbours are not in this PR" caveat predicted: the recordings now pass their TSTs and advance to the next wall in the same scene chain. `00EFEDF6`, which blocked at `4A39`, now blocks ~60 bytes deeper at the deferred `0C79` (`CMPI.W #imm,(xxx).L`, three extension words); its projected instruction count grew 119K → 193K as the recording lengthened. One neighbour advanced into a `61FF` call instead. So the admission converted the TST wall into deeper recordings rather than instant full traces — the remaining gates are the recorder's two-extension-word shape and call coverage, both tracked separately.

**Whole-session movement, usual caveats** (live gameplay, not instruction-matched; head-table membership is the form-specific signal): record share 58.7% → **61.2%**, and average retired ops per native call rose 95.6 → 122.6 — the reverse of the previous session's composition effect, consistent with recordings extending through admitted TSTs rather than new short traces joining.

benletchford (changes requested):
The absolute TST decode, checked-load behavior, flags, and cycle/parity coverage passed the full local CONTRIBUTING suite on the reviewed head, but this branch now conflicts with `master` after #109. Please rebase onto current master while preserving #109's blocker-only prefix salvage in `finish_recording`, all three salvage unit cases, `salvaged_prefix_loop_matches_step`, and the `salvage-prefix` benchmark alongside the absolute-TST additions, then rerun the complete checks on the integrated result.

rlanday:
From Claude Opus 5:

Full suite on the rebased head: default 813/0 (including the Musashi/SST fixtures), lib jit-only 183/0, all-features 209/0, run_batch integration 34/0, `clippy --all-features --all-targets` clean, portable build clean, benches build clean.

rlanday:
From Claude Opus 5:

| run | base | cand |
|-----|------|------|
| 1 | 50.4 | 329.5 |
| 2 | 47.8 | 312.4 |
| 3 | 46.6 | 304.7 |
| 4 | 46.0 | 303.8 |
| 5 | 45.8 | 294.1 |

benletchford (commented):
After #116 merged, this head now conflicts in `benches/microbench.rs`, `src/core/trace_jit.rs`, and `tests/run_batch_tests.rs`. Please rebase and preserve the absolute-CLR decode/emission/portable paths, cycle coverage, run-batch regression, and benchmark alongside absolute TST, then rerun the complete checks.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.9), preserving #101's continuation exit seeding/chaining and #116's absolute-CLR decode, native and portable execution, cycle coverage, run-batch regression, and benchmark alongside absolute TST.

Full suite on the rebased head: default 818/0 (including the Musashi/SST fixtures), lib jit-only 192/0, all-features 218/0, run_batch integration 37/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.10) now that #115 has merged, preserving the AND/OR memory-source work alongside absolute TST.

benletchford (changes requested):
This head now conflicts with master after #107. Please rebase the absolute TST extension onto the current trace-op layout while preserving the call-through ops, per-op validation, two code intervals, and their regression coverage. Keep the absolute-address TST load-only behavior and exact 68000 cycle cases intact, then rerun the full required workflow.

rlanday:
Rebased onto master at 0.10.12 and squashed to a single commit.

Rather than replay the branch's history (which carried two commits that existed only to repair an earlier bad conflict resolution), I rebuilt the branch from master and reapplied the feature: the production hunks by patch, then the three unit tests, the bench and the integration test inserted verbatim from the original commit. That avoids the interleaving that had been splicing unrelated test bodies together.

On your review points:

- **Call-through ops and per-op validation are preserved.** They come from master untouched — this branch adds no arm to the op enum's dispatch that could shadow them, and the two-way function-name audit against master shows additions only.
- **Absolute TST stays load-only.** `emit_tst_mem` generalises from an indexed-only destructure to a match over the effective address, so `AbsWord`/`AbsLong` materialise a constant address and fall through to the same `checked_window_off` + `window_load` + `set_logic_flags` path. There is no store, so no store guard and no two-interval question arises for this op.
- **Exact 68000 cycles** are kept and pinned by `tst_from_absolute_cycles_match_the_interpreter_on_a_68000`, which compares against what `step()` charges.

One thing worth flagging because it was a real defect in the branch rather than a rebase artifact: the branch had *converted* `mem_trace_clr_absolute_matches_step` into the TST test instead of adding alongside it, which silently removed master's absolute-CLR integration coverage. Both tests now exist.

Gate on the rebased head: fmt clean, `clippy --all-targets --all-features -D warnings` clean, lib jit-only 208/0, default 825/0 including the Musashi and SingleStep fixtures, all-features 907/0.

benletchford (approved):
Reviewed the rebuilt head against current master. Absolute TST remains a sized, checked, read-only operation; sign extension, flags, exact 68000 cycle charges, and native/portable parity are covered without replacing master regressions. The full local gate passed: fmt, Clippy with all targets/features, default tests, all-feature tests, doctests, and package verification.

PR #118, opened 2026-08-10
perf: carry a third extension word and admit the forms it unlocks

rlanday:
## Why

The recorder's per-op record carried at most two extension words. That budget — not semantics — has kept a growing set of census classes blocked: after the absolute-CLR and absolute-TST admissions, **the top-forty rejected-heads table of the latest capped EV Override session is gated on `0C79`** (`CMPI.W #imm,(xxx).L`, two heads, the walls the converted TST heads advanced into), **`13FC`** (`MOVE.B #imm,(xxx).L`), and **`2D7C`** (`MOVE.L #imm,(d16,An)`) — every one a three-word form.

## What

Two commits, structural then admissions:

1. **`TraceBuildOp` gains a third extension slot.** It joins instruction length, the trace's assembled code bytes, and head revalidation, so a store into *any* recorded word still invalidates the trace. The decode plumbing binds the slot outside the per-family match, so the ~30 existing decode arms are untouched.
2. **The unlocked admissions**, each with exact cycle entries pinned against the 68000 step interpreter:
   - `CMPI.W #imm,(xxx).W/(xxx).L` (16/20 cycles) — the abs.W form fit the old budget and was simply never admitted; it comes along.
   - `MOVE.B/W #imm,(xxx).W/(xxx).L` (16/20) and `MOVE.L #imm,(xxx).W` (24).
   - `MOVE.L #imm,(d16,An)` (24) and `MOVE.L #imm,(d8,An,Xn)` (26) — the existing family's long forms, previously refused at the budget check.
   - `MOVE.L #imm,(xxx).L` needs **four** words and stays out; the decode test asserts the exclusion.

## Does the structural change cost existing forms anything?

By construction, no: the emitted native code for existing forms is byte-identical, the new cycle arms only match previously-undecodable patterns, and the third slot appears on the recording and validation paths as one always-`None` check. The one real cost is +4 bytes per retained op in compiled traces. Measured: the A/B below includes an existing-form **control bench** (`clr-predec`, identical on both arms), which comes out at parity.

## Self-modification coverage, stated precisely

Head revalidation has two paths. The **contiguous slice compare** — the path every native trace on this branch takes — compares the assembled code bytes, which now include the third word; an integration differential (`mem_trace_third_extension_word_revalidates`) patches a live trace's third word (the absolute address's low half) mid-run and requires the re-record, and mutating the code-byte assembly makes it fail. The **per-op fallback walk** also carries the third word for parity, but its live cases are non-contiguous traces, which on this branch do not exist (we verified with a profiler probe that code-outside-window and taken-branch-gap shapes never compile natively — two candidate tests turned out vacuous and were dropped rather than kept as false assurance). The walk's live discriminator arrives with call coverage (#107-style two-segment traces) and will be exercised in the session-bundle validation.

## Measured effect

**Focused microbench** (filter `third-ext`): the three census shapes in a DBRA loop, 210M instructions per run after a 5M warmup, deterministic `run_batch` workload; both arms `--features jit`, same idle machine; base is `origin/master` (0.10.6) with an identical benchmark copy; five alternating runs, median of pairwise ratios; `clr-predec` runs as the existing-form control on the same arms:

| run | base | cand | control base | control cand |
|-----|------|------|--------------|--------------|
| 1 | 48.8 | 329.3 | 391.5 | 404.7 |
| 2 | 49.0 | 325.5 | 392.1 | 401.3 |
| 3 | 49.1 | 327.1 | 395.7 | 402.5 |
| 4 | 48.7 | 328.5 | 393.8 | 401.8 |
| 5 | 49.1 | 326.1 | 398.4 | 402.7 |

Median pairwise **6.66×** (M instr/s); control median **1.02×** — the structural change costs existing forms nothing measurable. (An earlier A/B run was discarded: its control arm showed 0.85–1.13 spread, i.e. the machine was not idle — which is precisely what the control is for.)

**Whole-application headless A/B** (deterministic 900M-instruction EV Override boot workload): exactly flat — every total identical between arms (backward_hits 553,701; rejected_hits 524,755; native_calls 68,054; jit_retired 16,375,726). The boot workload contains none of the three classes; they are gameplay-scene mass, on the same acceptance pattern as the preceding admissions — a bundled session with this branch is the follow-up.

## Validation

- `move_immediate_to_memory_decodes_within_the_extension_budget` — now documents the three-slot spec end to end: the `2D7C` and long-indexed shapes decode with the third slot populated and correct lengths; the four-word form still refuses.
- `cmpi_to_absolute_decodes_with_exact_extents` — two- and three-word CMPI forms, extents and sign-extension.
- `native_cmpi_absolute_matches_portable_with_exact_flags` — Z and N/C-borrow cases with X preserved; memory bit-identical (CMPI writes nothing).
- `native_three_extension_immediate_stores_match_portable` — the `2D7C` and `13FC` shapes: stored values, neighbours untouched, flags, charge.
- `three_extension_forms_cycles_match_the_interpreter_on_a_68000` — one pass over all three shapes charges exactly what step() charges.
- `mem_trace_third_extension_word_revalidates` (integration) — the SMC differential described above.
- Mutation-checked: dropping the code-byte assembly of the third word fails the SMC differential; dropping the CMPI cycle arms fails the pin (72 vs 74); dropping the MOVE-immediate arms fails it (66 vs 74); re-closing the decode gate fails the budget test.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1818,34 +1818,33 @@ fn bench_set<B: BenchBus>(label: &str) {
     );
 }
 
-/// Exercise the absolute-addressed TST forms the gameplay census
-/// flagged (4A39/4A79 heads): all three widths against fixed targets.
-fn bench_tst_abs_loop() {
-    // Outer cycles of 514 instructions end exactly at the outer label
-    // with the targets untouched (TST writes nothing).
+/// Exercise the three-extension-word forms the gameplay census named:
+/// CMPI.W to an absolute long address, MOVE.B immediate to an absolute
+/// long address, and MOVE.L immediate to a displaced destination.
+fn bench_third_ext_loop() {
+    // Outer cycles of 514 instructions end exactly at the outer label.
     const CYCLES: u32 = 408_560;
     const INSTRS: u32 = CYCLES * 514;
     const CODE_BASE: u32 = 0x7000;
     let words = [
         0x727F, // outer: MOVEQ #127,D1
-        0x4A78, 0x4000, // inner: TST.W ($4000).W
-        0x4AB9, 0x0000, 0x4004, // TST.L ($4004).L
-        0x4A39, 0x0000, 0x4003, // TST.B ($4003).L
-        0x51C9, 0xFFEE, // DBRA D1,inner
-        0x60E8, // BRA.S outer
+        0x0C79, 0x0042, 0x0000, 0x4000, // inner: CMPI.W #$42,($4000).L
+        0x13FC, 0x00A5, 0x0000, 0x4003, // MOVE.B #$A5,($4003).L
+        0x2B7C, 0xDEAD, 0xBEEF, 0x0010, // MOVE.L #$DEADBEEF,($10,A5)
+        0x51C9, 0xFFE6, // DBRA D1,inner
+        0x60E0, // BRA.S outer
     ];
     let mut bus = LinearMemoryBus::new(0x10000);
     for (index, word) in words.iter().enumerate() {
         bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
     }
-    for address in 0x4000..0x4008 {
-        bus.write_byte(address, 0xAA);
-    }
+    bus.write_word_at(0x4000, 0x0042);
     let prepare_cpu = || {
         let mut cpu = CpuCore::new();
         cpu.set_cpu_type(CpuType::M68040);
         cpu.set_sr(0x2700);
         cpu.pc = CODE_BASE;
+        cpu.set_a(5, 0x4100);
         cpu
     };
     let mut warm_cpu = prepare_cpu();
@@ -1858,11 +1857,11 @@ fn bench_tst_abs_loop() {
     assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
     let elapsed = start.elapsed().as_secs_f64();
     assert_eq!(cpu.pc, CODE_BASE, "the run ends exactly at the outer label");
-    for address in 0x4000..0x4008 {
-        assert_eq!(bus.read_byte(address), 0xAA, "TST writes nothing");
-    }
+    assert_eq!(bus.read_word(0x4000), 0x0042, "compared word untouched");
+    assert_eq!(bus.read_byte(0x4003), 0xA5, "immediate byte stored");
+    assert_eq!(bus.read_long(0x4110), 0xDEAD_BEEF, "immediate long stored");
     println!(
-        "batch     absolute TST loop       {:8.1} M instr/s",
+        "batch     third-extension loop    {:8.1} M instr/s",
         f64::from(INSTRS) / elapsed / 1_000_000.0
     );
 }
@@ -1946,8 +1945,8 @@ fn main() {
         bench_far_call_through_leaf();
         return;
     }
-    if only.as_deref() == Some("tst-abs") {
-        bench_tst_abs_loop();
+    if only.as_deref() == Some("third-ext") {
+        bench_third_ext_loop();
         return;
     }
     if only.as_deref() == Some("clr-abs") {
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1949,6 +1949,10 @@ fn main() {
         bench_third_ext_loop();
         return;
     }
+    if only.as_deref() == Some("tst-abs") {
+        bench_tst_abs_loop();
+        return;
+    }
     if only.as_deref() == Some("clr-abs") {
         bench_clr_abs_loop();
         return;
@@ -2165,6 +2169,55 @@ fn bench_memory_and_loop() {
     assert!(cpu.d(3) > 5_000, "the loop actually iterated");
     println!(
         "batch     memory and loop         {:8.1} M instr/s",
+        f64::from(INSTRS) / elapsed / 1_000_000.0
+    );
+}
+
+/// Exercise the absolute-addressed TST forms the gameplay census
+/// flagged (4A39/4A79 heads): all three widths against fixed targets.
+fn bench_tst_abs_loop() {
+    // Outer cycles of 514 instructions end exactly at the outer label
+    // with the targets untouched (TST writes nothing).
+    const CYCLES: u32 = 408_560;
+    const INSTRS: u32 = CYCLES * 514;
+    const CODE_BASE: u32 = 0x7000;
+    let words = [
+        0x727F, // outer: MOVEQ #127,D1
+        0x4A78, 0x4000, // inner: TST.W ($4000).W
+        0x4AB9, 0x0000, 0x4004, // TST.L ($4004).L
+        0x4A39, 0x0000, 0x4003, // TST.B ($4003).L
+        0x51C9, 0xFFEE, // DBRA D1,inner
+        0x60E8, // BRA.S outer
+    ];
+    let mut bus = LinearMemoryBus::new(0x10000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    for address in 0x4000..0x4008 {
+        bus.write_byte(address, 0xAA);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu
+    };
+    let mut warm_cpu = prepare_cpu();
+    assert_eq!(
+        warm_cpu.run_batch(&mut bus, 5_000_000, &[0]).instructions,
+        5_000_000
+    );
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    assert_eq!(cpu.run_batch(&mut bus, INSTRS, &[0]).instructions, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert_eq!(cpu.pc, CODE_BASE, "the run ends exactly at the outer label");
+    for address in 0x4000..0x4008 {
+        assert_eq!(bus.read_byte(address), 0xAA, "TST writes nothing");
+    }
+    println!(
+        "batch     absolute TST loop       {:8.1} M instr/s",
         f64::from(INSTRS) / elapsed / 1_000_000.0
     );
 }
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -15958,4 +15958,252 @@ mod portable_tests {
             assert_eq!(t.length(), 6, "{label}");
         }
     }
+
+    #[test]
+    fn tst_from_absolute_decodes_with_exact_extents() {
+        let dcpu = cpu();
+        let mut bus = super::super::memory::LinearMemoryBus::new(0x1000);
+        // 4AB8 = TST.L (xxx).W with a negative address: sign-extends.
+        bus.write_word(0x0100, 0x4AB8);
+        bus.write_word(0x0102, 0x8100);
+        let trace = decode_trace_op(&dcpu, &mut bus, 0x0100, CpuType::M68040)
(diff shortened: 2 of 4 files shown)


rlanday:
From Claude Fable 5:

Closing the loop on the "per-op fallback walk" coverage note in the body: the promised live discriminator now exists in the session-bundle validation. On the bundle (where call coverage makes traces non-contiguous — caller and callee segments — so entry validation cannot use the slice compare and must walk per-op words), a new integration differential records a call-through trace whose callee carries the three-word `CMPI.W #imm,(xxx).L`, then patches the third word mid-run and re-enters. Removing the walk's third-word check makes the test fail with diverging instruction counts (3921 vs 3620 — the stale trace keeps comparing the old address and takes the wrong branch path every pass). Both revalidation paths are now mutation-discriminated: the slice compare on this branch, the per-op walk on the bundle.

rlanday:
From Claude Fable 5:

The bundled gameplay session promised in the PR body has now run with this branch included (same instrument: capped GUI session, `trace-profile`, top-forty rejected-heads table by hits at exit).

**Head-level: all three unlocked classes are gone from the census.** `0C79`, `13FC`, and `2D7C` hold zero of the top-forty blocked slots (previously two, one, and one-to-three across recent sessions). This completes the scene chain the absolute-TST admission had deepened: those recordings now pass their TSTs *and* the CMPI walls behind them.

**What the census looks like after:** `61FF` (BSR.L) now holds twenty-eight of the top forty slots — with the small-form frontier exhausted, the remaining blocked mass is almost entirely call coverage, which is separate ongoing work. Residuals: three `81FC` (DIVS.W #imm), four wide-multiply heads, one `0CAA` (`CMPI.L #imm,(d16,An)` — the long-CMPI variant this PR's body deferred).

Whole-session ratios come with the usual caveat (live gameplay, not instruction-matched — this session ran 2.7× the previous one's instruction count, with a different scene mix): record share 59.9% against 61.2% prior, average ops per native call 93.4. The head-table membership is the form-specific signal, and it is unambiguous.

benletchford (changes requested):
The code and PR title checks pass, but the branch cannot merge while the required Build & Test check is red at `Check commit messages`. Commit `e5e1461` uses `bench: third-extension-word loop`; `bench` is not an allowed type in `committed.toml`. Rewrite it to an allowed single-line Conventional Commit type (the benchmark/regression support fits `test:`), then rerun the complete required checks. Please also update the stale `decode_move_imm_mem_trace_op` doc comment, which still describes a two-extension-word budget and says the newly admitted long displacement form stays on the decoded path.

rlanday:
From Claude Fable 5:

Both items addressed (`HEAD`):

- The `bench:` commit is reworded to `test: third-extension-word benchmark loop` — history rewritten in place, no content change to the commit.
- The `decode_move_imm_mem_trace_op` doc comment now describes the three-slot budget and names the four-word `MOVE.L #imm,(xxx).L` as the only form remaining on the decoded path.

Full jit lib suite re-run green on the updated head; the required checks should now pass the commit-message gate.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.7) — the branch had only been reworded before, so it was still conflicting. #109's salvage machinery is preserved (its blocker-only trim, three unit cases, `salvaged_prefix_loop_matches_step`, and the `salvage-prefix` benchmark all present and green).

Verified before pushing, on the same two axes I applied to the rest of the stack:

- every non-trivial line the branch's commits added is present in the rebased tree — 0 missing across `trace_jit.rs` (481 lines), `trace_profile.rs` (10), `benches/microbench.rs` (35), and `run_batch_tests.rs` (17);
- the third extension slot has 30 references in production code, not merely in tests.

Full suite on the rebased head: default 813/0 (including the Musashi/SST fixtures), lib jit-only 184/0, all-features 210/0, run_batch integration 34/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean. The commit-type fix from the previous push (`test:` rather than `bench:`) and the corrected `decode_move_imm_mem_trace_op` doc comment are both carried through.

benletchford (commented):
After #116 merged, this head now conflicts in `benches/microbench.rs`, `src/core/trace_jit.rs`, and `tests/run_batch_tests.rs`. Please rebase and preserve the absolute-CLR decode/emission/portable paths, cycle coverage, run-batch regression, and benchmark alongside the third-extension-word work, then rerun the complete checks.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.9), preserving #101's continuation exit seeding/chaining and #116's absolute-CLR decode, native and portable execution, cycle coverage, run-batch regression, and benchmark alongside the third-extension-word work.

Full suite on the rebased head: default 818/0 (including the Musashi/SST fixtures), lib jit-only 193/0, all-features 219/0, run_batch integration 37/0, `clippy --all-features --all-targets` clean, portable (`--no-default-features`) build clean, benches build clean.

rlanday:
From Claude Opus 5:

Rebased onto current master (0.10.10) now that #115 has merged, preserving the AND/OR memory-source work alongside the third-extension-word work.

benletchford (changes requested):
This head now conflicts with master after #107. Please rebase the third-extension-word representation onto the new trace-op layout, initializing and validating it consistently for the newly added / forms as well as existing ops. Preserve the two-segment code validation/SMC protections and combine their tests with the CMPI/MOVE third-word mutation coverage before rerunning the full workflow.

benletchford (changes requested):
Correction to the identifiers omitted above: initialize and validate the third extension field consistently for the new CallThrough and RtsReturn forms too.

rlanday:
Rebased onto master at 0.10.12.

**`extension3` is now initialised and validated for the call forms too.** `decode_call_op` builds four `TraceBuildOp`s — RTS, `JSR d16(PC)`, `JSR (xxx).L`, and the BSR family — and every one sets the slot explicitly:

```rust
extension2: None,
// No call form reaches three extension words: the widest is
// BSR.L / JSR (xxx).L at two. The slot is explicitly empty so
// revalidation never reads a stale third word for these ops.
extension3: None,
```

`None` is the correct value rather than a placeholder: the widest call form is BSR.L / JSR (xxx).L at two words. Validation is uniform because the revalidation walk is driven by what each op claims — `if let Some(expected) = op.extension3 { ... }` at the same offset arithmetic as the first two slots — so a call op contributes nothing to the walk, and `length()` and the code-image builder both agree with it. Had the slot been filled in from anywhere, revalidation would compare a word two bytes past the instruction and force a spurious miss on every entry.

New test `call_forms_claim_their_extension_words_and_no_third` pins this: RTS claims 2 bytes with all three slots empty, BSR.W claims 4 with one word, and BSR.L and JSR (xxx).L each claim 6 with exactly two — with a would-be third word written at `pc+6` that must not be picked up. Mutation-checked: setting `extension3: Some(0)` in the BSR path fails it.

Two notes on what else moved in the rebase. Master's `move_immediate_to_memory_decodes_within_the_extension_budget` asserted the old two-slot ceiling — that `MOVE.L #imm,(d16,An)` and the long indexed form must *not* decode. Those are precisely the forms this PR admits, so the assertions are replaced with the positive ones (three claimed words, `length() == 8`), and the four-word `MOVE.L #imm,(xxx).L` case still asserts the form stays out. And `mem_trace_third_extension_word_revalidates` had ended up in the file twice with identical bodies; one copy is removed.

Gate: fmt clean, `clippy --all-targets --all-features -D warnings` clean, lib jit-only 210/0, default 825/0 including the Musashi and SingleStep fixtures, all-features 908/0, run_batch integration 39/0.

benletchford (changes requested):
Two blockers remain on the latest head. First, `tests/run_batch_tests.rs` still replaces `mem_trace_clr_absolute_matches_step` with `mem_trace_third_extension_word_revalidates`; it does not add the new regression alongside the master test. That silently removes the absolute-CLR record/compile/execute coverage, and would also drop it from the now-updated base after #117. Restore the CLR test unchanged and add the third-word test separately. Second, call-through is already live on current master, so the per-op non-contiguous validation walk is no longer a future/session-bundle case. The branch only mutation-tests the contiguous code-slice path; `call_forms_claim_their_extension_words_and_no_third` does not prove that changing a callee op’s third word invalidates a two-segment compiled trace. Add an in-repository call-through regression that compiles a callee containing a three-extension CMPI/MOVE, mutates extension3, and demonstrates revalidation instead of stale execution. Then rerun the full required checks.

rlanday:
From Claude Opus 5:

Rebased onto 0.10.13 and fixed a coverage defect on this branch, found by
applying your #113 review to the rest of the stack rather than waiting for you
to hit it here.

This branch had the same problem: `mem_trace_third_extension_word_revalidates`
went in *place of* `mem_trace_clr_absolute_matches_step` rather than alongside
it. Rebasing onto 0.10.13 then dropped #117's TST work the same way, since that
merged after this branch's base — `bench_tst_abs_loop` and its `tst-abs`
dispatch arm, plus `tst_from_absolute_decodes_with_exact_extents`,
`tst_from_absolute_cycles_match_the_interpreter_on_a_68000` and
`native_tst_from_absolute_matches_portable_with_exact_flags`.

All restored from master as whole definitions, with `extension3: None` added to
the literals so they build against the new field. The branch is now purely
additive against master: **0 deleted definitions, 0 deleted opcode lines**.

Worth recording why this got past me twice. My rebase audit compared function
*names* between master and the branch in both directions, which cannot detect a
body being replaced under a surviving name, and cannot detect a definition
surviving while its dispatch arm does not — both of which happened here
(`bench_tst_abs_loop` came back without its `if only.as_deref()` arm on the
first attempt, and clippy caught it as unused). The check I have added is a
content-level one, run per branch before pushing:

    base=$(git merge-base origin/master $BRANCH)
    git diff $base..$BRANCH | grep -cE '^- *0x[0-9A-F]{4}'   # deleted opcode lines
    git diff $base..$BRANCH | grep -cE '^-\s*fn [a-z_]'      # deleted definitions

Both must be 0 for an additive form branch. I ran it across the others: #111 and
#112 are clean.

Gate on the rebased head: fmt clean, `clippy --all-targets --all-features -D
warnings` clean, doc clean, lib(jit) 213/0, default 828/0, all-features 913/0.
Both restored `mem_trace_clr_absolute_matches_step` and
`mem_trace_tst_absolute_matches_step` pass alongside the new third-extension
test.

No whole-app measurement on this one yet. Given how #113 turned out I would
rather measure before making any performance claim here, so treat this as a
correctness/coverage fix to the branch and not as evidence it is worth landing.

rlanday:
From Claude Fable 5:

Whole-application measurement for this branch, and a close on the result.

**Instrument.** EV Override 1.0.1 on `benletchford/systemless`, driven headless by a scripted input replay whose events are scheduled by retired instruction count, so every run performs exactly the same 400,000,000 guest instructions from a pristine save. Arms: this branch's head (`3e35944`) versus its base (0.10.13), both built with `trace-profile` and run under `M68K_TRACE_PROFILE=1` so the instrument is common-mode. Arms alternate within each repetition and are compared pairwise, which cancels machine-load drift; 12 pairs. Metrics: native share (`jit_retired` / 400M, the metric #113 was decided on) and child CPU seconds on the fixed work.

**Result: no measurable application-level benefit.**

- native share: **+0.111pp**, sd 0.887, t = +0.43, better in 4 of 12 pairs — not resolved
- CPU on fixed work: **−1.79%**, t = −0.58, faster in 7 of 12 pairs — not resolved

For calibration, the same instrument at the same sample sizes resolves #111 at +0.640pp (t = +4.20) and #112 at +0.673pp (t = +3.48), so it detects effects of the size this series has produced when they exist.

That matches where the profiles stand: after the recent admissions, no ranked blocked head in the gameplay census strands on a three-extension-word form, so the forms this branch admits are not on any measured hot path. Closing under the same standard as #113 — correctness-clean but no demonstrated end-to-end benefit — which also moots the outstanding callee-mutation revalidation regression from the last review. If a future census surfaces a three-extension form on a head that completes, this branch (rebased, with your requested regression added) is the starting point; the coverage restoration it carried is superseded by the branch simply not merging.

PR #129, opened 2026-08-15
feat(profile): per-head chained-call and chained-retirement counters

rlanday:
The per-head chained-call and chained-retirement counters requested in #100's latest round, so the guard-exit re-formation question can be decided on field data.

## What it adds

Two counters per trace head in `trace-profile`, plus two columns in the compiled-traces table:

- **`chained_calls`** — guard exits at this head that entered a compiled continuation trace (the bounded chaining from #101). Incremented whether the chain completed or surfaced a revalidation miss, since either way the exit reached compiled code.
- **`chained_retired`** — guest instructions those chains retired.

As your review noted, a parent's guard-exit ratio alone cannot distinguish ineffective chaining from a parent that exits and then chains productively: `guard_exits` versus `chained_calls` now makes that split per head, and `chained_retired` prices what the chains actually deliver. Attribution is to the parent whose exit spawned the chain; nested chains attribute to their own immediate parent, so a chain-of-chains is not double-counted upward.

## Verification

A focused regression (`chained_counters_split_productive_exits_from_thrash`) drives the mixed-path loop from the existing exit-seeding test — a shape that measurably chains — and asserts: some head records chained calls; every such head has `chained_calls <= guarded_branch_exits`; chains that ran retired instructions; and the table exposes at least one side of the thrash/productive split. The test was mutation-validated: silencing the two `note_chained` call sites makes it fail.

Full local workflow on this head (0.10.14 base): rustfmt clean, `clippy --all-targets --all-features -D warnings` clean, lib/default/all-features suites all passing, `cargo doc` clean; zero deletions against master. Profiler-only — no behavior change outside `trace-profile`.

Next step per the #100 plan: one capped GUI session on this instrumentation, reporting the counters for each guard-exit-dominated parent, posted to #100 — after which that issue narrows to either close-as-answered or one fully specified re-formation proposal.

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -1164,13 +1164,19 @@ impl TraceJit {
                             watch_pcs,
                             chain_budget - 1,
                         ) {
-                            Some((CachedRunResult::Ran, chained)) => retired += chained,
+                            Some((CachedRunResult::Ran, chained)) => {
+                                #[cfg(feature = "trace-profile")]
+                                super::trace_profile::note_chained(pc, cpu_type, chained);
+                                retired += chained;
+                            }
                             // The continuation's first opcode changed:
                             // the child has already consumed it (ppc/ir
                             // set, pc advanced) and the caller must
                             // dispatch it. Surface the miss while
                             // keeping every instruction retired so far.
                             Some((miss @ CachedRunResult::Miss(_), chained)) => {
+                                #[cfg(feature = "trace-profile")]
+                                super::trace_profile::note_chained(pc, cpu_type, chained);
                                 return Some((miss, retired + chained));
                             }
                             None => {}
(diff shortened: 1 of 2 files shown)


benletchford:
This looks like sensible, generic profiler-only instrumentation and the existing focused test/CI are good. After #111 landed, though, GitHub now reports this branch as non-mergeable against current `master`, so I’m not going to force the old head through.

Please rebase onto current `master`, preserve the continuation-chaining paths plus #111’s MOVEM work, and rerun the full Rust CI. Assuming that integrated head stays green, I’m happy to merge this one.

rlanday:
Rebased onto current `master` (db55299, includes #111). The rebase was conflict-free and the branch diff vs master is exactly the original feature (+110/−3 across `trace_jit.rs`/`trace_profile.rs`); #111's MOVEM paths are intact. Local checks on the rebased head: `cargo fmt --check` clean, clippy clean, `--features trace-profile` 178 pass, `--features jit,trace-profile` 241 pass. — Claude Fable 5

benletchford (approved):
Profiler-only instrumentation is focused, the attribution semantics are tested, and the branch has been rebased and validated against the complete merged JIT head. CI is green.

PR #131, opened 2026-08-16
perf(jit): durable structural trace rejection, gated to compiled-before heads

rlanday:
## Summary

Remember structural trace rejections (`NoTraceTerminal`, `TooShort`,
`IndirectJsrTooShort`) across trace-cache slot eviction, so a genuinely
uncompilable head is not re-recorded on every eviction — **but require
real evidence before a `NoTraceTerminal` verdict becomes durable**, so a
compilable loop is never permanently poisoned from either direction:

- a head that has **ever compiled** is exempt (a later no-terminal pass
  is a transient data-dependent path, and it re-arms), and
- a head that has **never compiled** must produce **3 no-terminal
  recordings with no compile in between** before the verdict sticks —
  one bad first sample is not a structural fact.

Four commits:
1. **remember structural rejections across slot eviction** — the durability
   mechanism (`structurally_rejected` table).
2. **gate durable NoTraceTerminal to compiled-before heads** — the
   compiled-before exemption.
3. **require repeated NoTraceTerminal evidence** — the strike table that
   closes the first-sample hole (review feedback).
4. **the inverse regression test** — mutation-verified against both old
   behaviours.

## Why single-observation durability is wrong

`NoTraceTerminal` is durable in commit 1, on the theory that "no
terminal" is a static property of the recorded region. For a
straight-line region it is. For a **nested loop's outer head it is
not**: the recorder records one straight-line path, and for an outer
per-sample loop around a data-dependent inner loop it —

- **closes cleanly on the outer back-edge and COMPILES** on the passes
  where it observes the inner loop's exit, and
- trips the `repeated`-pc guard on an interior non-branch op (ending
  with **no terminal**) on the passes where it observes the inner loop
  iterating.

Which of those the recorder sees **first** is data, not structure. A head
that compiled first and rejected later was poisoned by commit 1 alone
(the measured audio-mixer case below); a head that rejected first and
would compile later was poisoned by the compiled-before gate alone (the
review's inverse case). Both orderings are now safe.

## Measured impact (this is why commit 1 alone is a regression)

EV Override gameplay flight, headless input-replay, 400M guest
instructions, default 4096-entry cache. The software audio mixer's outer
loop at guest `0x0118A12C` is **68% of all interpreted execution**. Three
arms, same binary, gate toggled, cache held constant:

| arm | trace coverage | host-instr / guest |
|---|---|---|
| **no durability** (master, self-heals via eviction) | 72.3% | ~883 |
| **durability, ungated** (commit 1 only) | **55.2%** | **~914** |
| **durability, gated** (commit 1 + 2) | **72.3%** | **~862** |

- Commit 1 **alone regresses** trace coverage by 17 points and raises host
  work ~3.5% — it poisons the audio loop (`0x0118A12C` was probe-filtered
  on 99.5% of its 3.59M iterations).
- Commit 2 restores coverage to master **and** lands ~2–4% *below* master
  on host work (paired runs: 0.976 / 0.964 / 0.988) — the durability now
  cuts re-record churn on genuinely-uncompilable heads without poisoning
  compilable ones. Net win on both axes.
- Commit 3 changes behaviour only for never-compiled heads on their first
  two no-terminal recordings (each of which re-arms one retry); a
  genuinely uncompilable head still goes durable after at most 3 bounded
  recordings — against the 2,959 re-records the pathological head showed
  without durability, the anti-churn property is intact.

## The mechanism

Track, per head, whether it has ever compiled a valid trace
(`compiled_before`) and how many no-terminal recordings it has produced
since (`no_terminal_strikes`) — both 2-way-per-slot pc tables mirroring
`structurally_rejected`. On a `NoTraceTerminal` rejection:

- compiled-before head → never durable; `record_trace_target` re-arms
  its same-pc `Rejected` slot.
- never-compiled head → strike; durable only at
  `NO_TERMINAL_STRIKE_LIMIT` (3). Below the limit the head re-arms the
  same way, so the closing path gets a deterministic retry.

A successful compile clears the head's strikes (the evidence is
disproven). SMC under the head clears strikes with the verdict.
`TooShort` / `IndirectJsrTooShort` remain unconditionally durable — they
are static properties of the recorded region.

## Tests

- `a_first_sample_no_terminal_head_can_still_compile_later` — **the
  inverse regression**: a mid-loop-entry head (the profiled audio-mixer
  shape) whose recordings flow sequentially back into their own start
  (no-terminal, no salvage) while the loop spins, then a later entry
  takes the closing path — the head must compile. **Mutation-verified**:
  single-observation durability fails phase 1; compiled-before-only
  re-arm fails phase 2.
- `no_terminal_strikes_accumulate_to_durable_and_a_compile_clears_them` —
  strike-table mechanics: pc-specific across aliases, compile-clears,
  SMC-clears.
- `a_compiled_before_head_is_not_durably_rejected_by_no_terminal` — a
  compiled-before head driven through the same no-terminal loop that
  freezes an exhausted head must NOT be structurally rejected and must
  keep re-recording. **Fails if the gate is removed.**
- `compiled_before_is_two_way_and_pc_specific`,
  `a_compiled_head_is_exempt_from_the_durable_no_terminal_verdict` — the
  tables and the gate logic.
- The original durability tests
  (`structural_rejection_survives_slot_eviction`,
  `rewritten_code_clears_the_structural_verdict`, …) pass unchanged: a
  head that exhausts its strikes is durably rejected and never re-records
  after eviction.

Rebased onto current `master` (db55299, includes #111); full matrix green
locally (`trace-profile` 184, `jit,trace-profile` 247).

---

*This PR (all commits and this writeup) is from Claude Fable 5, working
with @rlanday. Supersedes #135, whose fix is folded in as commit 2.*

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -65,6 +65,15 @@ enum ExitSeed {
 const TRACE_ADAPT_WINDOW: u8 = 64;
 const TRACE_ADAPT_MISMATCHES: u8 = 48;
 const TRACE_MAX_ADAPTIVE_RERECORDS: u8 = 1;
+/// `NoTraceTerminal` recordings a never-compiled head may accumulate
+/// before the rejection becomes durable. One observation is not evidence
+/// of structure -- a data-dependent head's first recorded path may simply
+/// not have closed -- and each additional strike costs one bounded
+/// re-record, so the limit trades a vanishing chance of poisoning a
+/// compilable head (misses require this many consecutive non-closing
+/// recordings with no compile between them) against at most this many
+/// wasted recordings on a genuinely uncompilable one.
+const NO_TERMINAL_STRIKE_LIMIT: u8 = 3;
 
 /// Sentinel for `CpuCore::trace_record_skip` / `trace_probe_skip`: no PC.
 pub(crate) const TRACE_PC_NONE: u32 = u32::MAX;
@@ -711,6 +720,14 @@ pub(crate) struct TraceJit {
     /// as a transient data-dependent path (re-recordable), NOT a durable
     /// structural verdict -- see finish_recording_with_retry / record_trace_target.
     compiled_before: Vec<[u32; 2]>,
+    /// `NoTraceTerminal` observations (by pc, 2-way per cache index) on
+    /// heads that have never compiled. One observation proves nothing
+    /// about a data-dependent head whose FIRST recorded path happened not
+    /// to close -- the inverse of the compiled-before case -- so the
+    /// verdict becomes durable only after `NO_TERMINAL_STRIKE_LIMIT`
+    /// observations with no compile in between. A successful compile
+    /// clears the count; SMC under the head clears it with the verdict.
+    no_terminal_strikes: Vec<[(u32, u8); 2]>,
 }
 
 impl fmt::Debug for TraceJit {
@@ -747,6 +764,7 @@ impl TraceJit {
             earned_call_permission: vec![[u32::MAX; 2]; TRACE_CACHE_SIZE],
             structurally_rejected: vec![[u32::MAX; 2]; TRACE_CACHE_SIZE],
             compiled_before: vec![[u32::MAX; 2]; TRACE_CACHE_SIZE],
+            no_terminal_strikes: vec![[(u32::MAX, 0); 2]; TRACE_CACHE_SIZE],
         }
     }
 
@@ -1248,6 +1266,47 @@ impl TraceJit {
                 *way = u32::MAX;
             }
         }
+        // The verdict's evidence is void with the verdict: the code under
+        // this head changed, so accumulated no-terminal observations
+        // describe bytes that no longer exist.
+        self.clear_no_terminal_strikes(pc);
+    }
+
+    /// Count one `NoTraceTerminal` observation against a never-compiled
+    /// head and return the total so far. Insertion mirrors the other
+    /// two-way tables: an aliasing flood can evict a count, which only
+    /// delays durability (safe direction).
+    fn note_no_terminal_strike(&mut self, pc: u32) -> u8 {
+        let ways = &mut self.no_terminal_strikes[trace_cache_index(pc)];
+        for way in ways.iter_mut() {
+            if way.0 == pc {
+                way.1 = way.1.saturating_add(1);
+                return way.1;
+            }
+        }
+        if ways[0].0 == u32::MAX {
+            ways[0] = (pc, 1);
+        } else if ways[1].0 == u32::MAX {
+            ways[1] = (pc, 1);
+        } else {
+            ways[1] = ways[0];
+            ways[0] = (pc, 1);
+        }
+        1
+    }
+
+    fn has_no_terminal_strikes(&self, pc: u32) -> bool {
+        let ways = &self.no_terminal_strikes[trace_cache_index(pc)];
+        ways[0].0 == pc || ways[1].0 == pc
+    }
+
+    fn clear_no_terminal_strikes(&mut self, pc: u32) {
+        let ways = &mut self.no_terminal_strikes[trace_cache_index(pc)];
+        for way in ways.iter_mut() {
+            if way.0 == pc {
+                *way = (u32::MAX, 0);
+            }
+        }
     }
 
     fn record_trace_target(&mut self, pc: u32, cpu_type: CpuType) {
@@ -1280,9 +1339,15 @@ impl TraceJit {
                 // a nested loop's outer head). Re-arm it so the next hot pass
                 // can re-record the compilable shape, instead of leaving it
                 // Rejected until an eviction that -- with durable structural
-                // rejection -- never revives it. Durable (structurally
-                // rejected) heads are left alone.
-                if self.has_compiled_before(pc) && !self.is_structurally_rejected(pc) {
+                // rejection -- never revives it. A never-compiled head still
+                // inside its strike budget gets the same retry: its next
+                // recorded path may close (the inverse regression), and if it
+                // never does the strikes run out and the verdict goes
+                // durable. Durable (structurally rejected) heads are left
+                // alone.
+                if !self.is_structurally_rejected(pc)
+                    && (self.has_compiled_before(pc) || self.has_no_terminal_strikes(pc))
+                {
                     self.slots[idx] = TraceSlot::Counting {
                         pc,
                         cpu_type,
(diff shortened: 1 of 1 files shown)


rlanday:

benletchford:
I think there is still a correctness hole in the `compiled_before` gate.

The PR correctly establishes that `NoTraceTerminal` is not necessarily a static structural property: a data-dependent/nested-loop head can fail to close on one observed path and compile on another. But if that same genuinely-compilable head happens to produce `NoTraceTerminal` on its *first* recording, `has_compiled_before(start_pc)` is false, so the current code makes that verdict durable and the head never gets the later opportunity to demonstrate that it can compile.

The current regression coverage exercises the compiled-first case. Please add the inverse regression: first recording takes the no-terminal path, a later execution takes a path that would close a valid trace, and the head must eventually compile. I don’t think a single `NoTraceTerminal` observation on a never-before-compiled head can safely become permanent. `TooShort` / `IndirectJsrTooShort` can remain durable if they are genuinely static, or `NoTraceTerminal` needs stronger/repeated evidence rather than `compiled_before` as the safety condition.

Once that case is resolved and the full CI is green on current `master`, I’ll re-review.

rlanday:
You're right — the first-sample ordering was a real hole: a genuinely compilable head whose *first* recording happened to take a non-closing path was durably poisoned with no escape, the exact inverse of the case this PR fixes. Both are now addressed; `compiled_before` is no longer the only safety condition.

**What changed (strike-based durability):**

- `NoTraceTerminal` on a never-compiled head now accumulates **strikes** (a 2-way per-index pc table, same shape as the verdict table) instead of becoming durable on one observation. The verdict becomes durable only at **3** observations with no compile in between.
- Each strike **re-arms one more recording attempt** (`record_trace_target` re-arms a same-pc `Rejected` slot that holds strikes, exactly as it re-arms a compiled-before head), so the later closing path gets its chance deterministically rather than waiting on an aliasing eviction.
- A successful **compile clears the strikes** (the evidence is disproven); **SMC** under the head clears them along with the verdict.
- `TooShort` / `IndirectJsrTooShort` remain durable on first observation — they are static properties of the recorded region.
- Bounded churn: a genuinely uncompilable head wastes at most 3 recordings, ever — the anti-churn goal of the PR is intact (the profiled pathological head had recorded 2,959 times).

**The inverse regression you asked for** — `a_first_sample_no_terminal_head_can_still_compile_later`: HEAD_C is a *mid-loop entry* (the profiled audio-mixer shape): in phase 1 the wrap path branches back above the head and flows *sequentially* into it again, so every recording revisits its own start at a non-branch op and ends `no-trace-terminal` (no blocker, so salvage does not apply). Phase 2 enters with the loop counter at 1 and the exit flag set, and the exit path's `BEQ.S` closes a 4-op trace back to the head — which must compile. The test is **mutation-verified** in both directions:

- reverting the strike gate to single-observation durability → fails phase 1 ("must not be a durable verdict, attempts=1");
- reverting the re-arm to compiled-before-only → fails phase 2 (the head holds strikes, isn't durable, but stays frozen and never compiles).

Also added `no_terminal_strikes_accumulate_to_durable_and_a_compile_clears_them` for the table mechanics (pc-specific across aliases, compile-clears, SMC-clears).

The branch is rebased onto current `master` and the full CI is running on the integrated head. — Claude Fable 5

rlanday:
Validation data for the strike design, measured on the rebased head.

**Protocol** (same as the PR's earlier census): EV Override, headless input-replay into gameplay, 400M guest instructions per run, `M68K_TRACE_PROFILE=1`, default 4096-entry trace cache. Metric: `jit_retired / 400M` = fraction of guest instructions retired inside compiled traces (higher = more guest code JITed). Two binaries, both built from the same current `master` (db55299): the control is plain master, the fix arm is this PR's head (a local build with two extra profile-only strike counters; the counters are not in the PR).

**Coverage — the fix arm is higher AND stable, master is neither:**

| arm | pass 1 | pass 2 | pass 3 |
|---|---|---|---|
| master (db55299) | 61.0% | 44.9% | 59.8% |
| #131 + strike gate | 67.9% | 67.9% | — |

The fix arm's two passes agree to 0.1% (271.5M / 271.7M retired). Plain master swings by 16 points run-to-run — that is the slot-eviction re-record churn this PR removes, visible as coverage variance: whether a hot head is compiled at any moment on master is a race between aliasing eviction and re-record. On the current base this PR is not just "master coverage with less churn" (the earlier measurement's conclusion on an older base): it is **+7 to +23 coverage points over master with the variance eliminated**.

**Strike distribution — K=3 doing real work** (per-pass counts; `events@N` = recordings that took a head to N strikes, `rescued@N` = heads that COMPILED while holding N strikes):

| pass | events@1 | @2 | @3 (durable) | rescued@1 | rescued@2 |
|---|---|---|---|---|---|
| 1 | 1534 | 864 | 693 | 6 | 2 |
| 2 | 1558 | 886 | 698 | 2 | 1 |

- `rescued@1 > 0` in both passes: real heads compile after one no-terminal recording — **single-observation durability (the reviewed hole) poisons real compilable heads on this workload**, not just in the synthetic regression.
- `rescued@2 > 0` in both passes: some heads need the full 3 strikes — K=2 would still poison them.
- ~693 genuinely never-closing heads still reach durable, so the anti-churn goal is intact (bounded at 3 recordings each, vs the 2,959 re-records the pathological head showed without durability).

**Scope:** one workload (EV Override gameplay). The counts above argue K=3 is the smallest safe limit on this evidence; a second workload (SimCity 2000) is being brought up on the same harness and I'll extend the table when it's in.

— Claude Fable 5

rlanday:
Second workload landed: **SimCity 2000 1.2** (boot → new city → ~300M instructions of live simulation on the same headless harness; byte-identical `jit_retired` across passes). Same protocol and binaries as the EV table above.

| arm | coverage |
|---|---|
| master (db55299) | 43.9% |
| #131 + strike gate | **45.6%** |

Strike distribution: events@1=2010, @2=1506, durable=1310 — **rescued@1=18, rescued@2=8**. The K=3 evidence is stronger here than on EV Override: eight real heads compiled while holding two strikes, so K=2 would durably poison them; single-observation durability would poison twenty-six. The 1,310 genuinely never-closing heads still go durable (bounded at 3 recordings each), so the anti-churn property holds on a much more rejection-heavy workload. — Claude Fable 5

benletchford (approved):
The first-sample correctness hole is addressed by bounded repeated evidence, with inverse-order, compile-clear, SMC-clear, and workload validation. The current-master head is clean and CI is green.

PR #133, opened 2026-08-18
perf(jit): const-initialize the trace-JIT thread-local

rlanday:
From Claude Fable 5, working with @rlanday. Based on the same base as #131 (stacks cleanly either order; the diff is one commit).

## What

`try_execute_trace` takes the `TRACE_JIT` thread-local on every probe — once per batch entry and once per backward branch — and the lazily initialized `thread_local!` form re-runs its platform once-guard on every access. In a line-level profile of EV Override's title crawl under Systemless (tight interpreted loops, so probes are constant), that guard alone was **4 % of the main thread**: 252 of 6,417 sampled ms at `lazy.rs:59`, all under `try_execute_trace`, none of it doing JIT work.

Const-initialize the slot (`RefCell<Option<TraceJit>>` in a `const { … }` initializer) so access compiles to a direct TLS read, and create the state on first use behind one `with_trace_jit` helper. All 22 access sites keep their shape; no behavior change.

## Evidence

- Line-level attribution (debug-info release build, macOS `sample` + `atos -i`): the 252 samples at `lazy.rs:59` resolve through `TRACE_JIT::{constant}` under `try_execute_trace` — the guard, not the trace execution.
- Whole-app crawl A/B (8 s windows, one session per arm, Systemless master + its pending startup PRs on both sides, arms differing only in this commit): main-thread active 1,455 → 1,194 samples, `run_steps_internal` self 561 → 451. Single windows resolve to about ±3 % and the crawl's load varies with scroll position, so treat the magnitude as favorable; the direction matched the prediction and flight (which probes far less) was unchanged, as expected.

## Tests

No functional change; fmt, clippy (`--all-targets --all-features -D warnings`) and the full default + all-features suites pass (826 / 910).

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -88,7 +88,16 @@ enum NativeTraceFn {
 static TRACE_JIT_HAS_CANDIDATES: AtomicBool = AtomicBool::new(false);
 
 thread_local! {
-    static TRACE_JIT: RefCell<TraceJit> = RefCell::new(TraceJit::new());
+    // Const-initialized so access compiles to a direct TLS slot read. The
+    // lazily-initialized form re-ran its platform once-guard on every
+    // access, and `try_execute_trace` probes once per batch entry and per
+    // backward branch -- in EV Override's crawl that guard alone was 4% of
+    // the main thread (252 of 6,417 sampled ms).
+    static TRACE_JIT: RefCell<Option<TraceJit>> = const { RefCell::new(None) };
+}
+/// Run `f` on the thread's trace-JIT state, creating it on first use.
+fn with_trace_jit<R>(f: impl FnOnce(&mut TraceJit) -> R) -> R {
+    TRACE_JIT.with_borrow_mut(|slot| f(slot.get_or_insert_with(TraceJit::new)))
 }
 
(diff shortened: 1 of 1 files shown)


rlanday:

benletchford:
The const-initialised TLS change itself looks generic, sensible and low-risk. I can’t merge this PR as-is, though, because the branch is stacked on #131 and therefore carries the durable-rejection work that I’m holding for the first-sample `NoTraceTerminal` correctness issue.

Please restack/rebase this onto current `master` with just the const-TLS change (plus any necessary conflict resolution), without #131, and rerun the full Rust CI. In that form I expect this one to be straightforward to merge.

rlanday:
Restacked onto current `master` (db55299) with just the const-TLS change — the #131 durable-rejection commit is no longer carried; this PR's diff vs master is now only the TLS const-initialization + `with_trace_jit` accessor conversion (+46/−40, one file). Local checks: `cargo fmt --check` clean, clippy clean, `--features jit` 214 tests pass. — Claude Fable 5

benletchford (approved):
This is a focused const-TLS initialization change, now correctly restacked without the prior dependency. The accessor conversion is behavior-preserving and CI is green.

Issue #130, opened 2026-08-15
Proposal: a decoded-run fast tier, then in-core trap dispatch, for short-batch phases


rlanday:
From Claude Fable 5:

## What we measured

EV Override's scrolling-crawl phase is the app's worst CPU consumer (30–56% of a core
at 25MHz, vs ~3% on menus). We instrumented the systemless runner with a batch-shape
probe (host-side `Instant` timing around `run_batch` and around trap dispatch,
batch-length buckets, exit-reason counters) and took two human-cued GUI runs — one
closed immediately after the crawl, one closed just before entering it — so the
difference isolates the crawl window:

- **66.4M batches / 612.5M instructions in the crawl delta — mean batch length ≈ 9.2.**
  (Menu-phase batches average 28+.)
- Trap dispatch in this phase costs ~75ns per trap; the handlers are cheap. The cost
  is not in QuickDraw — it's in the core being entered and exited every ~9
  instructions.
- Cross-checked with a sampling profiler on the release build, the crawl decomposes to
  ≈38% `run_batch` per-instruction machinery, ≈17% runner plumbing, ≈35–40% trap
  handlers + drawing.

So the crawl pays the batch-entry cost (~15ns, measured by a budget-sweep microbench
over straight-line workloads: 1→4096 instruction budgets, paired in-process arms) plus
the full per-instruction machinery (record/watch/backward-note checks, ~6.6–7.4
ns/instr on our box) on nearly every instruction, because traps keep batches too short
for the JIT or the cache to amortize anything.

Both options below target the **interpreter tier**, not JIT'ed code: because traps
strand trace recordings, the crawl's loop heads never compile, so nearly all of its
instructions execute in the cached-decode interpreter loop. Compiled traces are
unaffected by either change.

## The two options, both benchmarked

**(a) Decoded-run fast tier.** Pre-decode straight-line runs once and execute the
decoded ops directly, consulting the machinery only at block boundaries. We benchmarked
the floor with a committed probe (`probe/decoded-run-floor`, 6eeb3a8): executing
pre-decoded ops runs at **3.1–3.5 ns/instr vs 6.6–7.4 baseline** — ratio 0.47–0.53
across within-process pairs (paired arms, same process, alternating; the only method
that resolves on a loaded box). Projected onto the crawl's 38% machinery share and 9.2
mean batch length, ceiling ≈ **19% of crawl CPU**. Contained entirely inside
`op_cache.rs`; no host-contract change.

[...]

PR #134, opened 2026-08-18
perf(jit): compile trap-punctuated segments via TrapExit terminals

rlanday:
From Claude Fable 5, working with @rlanday. **Now a single commit directly on `master` v0.11.0 (`01a4f7b`)**: the original base PRs (#111 #112 #131 #133) have merged, #128 was closed, and the 2026-08-20 restack keeps the generic mechanism only — no verifier-off carry, no stacked work. Design doc: `trap-crossing-traces-design.md` (attached in a comment).

## Why

After the current round of optimizations, over half of EV Override's gameplay main-thread time is the interpreter executing **trap-punctuated straight-line code** — runs of ~9 instructions between A-line traps (66 M nine-instruction batches in one measured crawl session). Those runs contain no backward branch and no closing branch, so under the current rules they can never become traces: heads arise only at backward-branch targets, and a region only compiles if it ends in a branch. The admission census puts everything reachable by *widening admission* at ~2 % of guest dispatch; this shape is the other ~98 % of interpreted execution.

## What

A region may now end **at** an A-line:

- **`TrapExit` terminal.** The trap word itself is recorded as the region's final op — its two bytes are real guest memory, so SMC validation covers the boundary exactly (a rewritten trap word invalidates the trace). Executing the terminal takes the existing bail convention: `pc` lands on the A-line, the terminal is not retired, and the batch loop surfaces the trap to the host exactly as an interpreted run would. **Zero changes to the host protocol.**
- **Continuation seeding, quality-gated.** When a recording closes at a trap *and a compiled segment now really ends at that boundary* (`TrapFinish::Compiled`), the expected resume point (trap + 2) is remembered (`CpuCore::pending_trap_resume`) and granted head candidacy at the next batch entry *if the host resumed exactly there* — a host that resumes anywhere else (LoadSeg-style flow traps) silently drops the hint. Segments chain themselves down a trap-punctuated run: `[head..trap₁) [trap₁+2..trap₂) … [tail..branch]`. A closure the compiler rejected does **not** seed: chains stop at boundaries that don't pay, so head-candidate growth tracks compiled segments rather than every trap crossing (this gate was added after headless A/B showed ungated seeding churning the direct-mapped slot array).
- `TRACE_MIN_OPS` still gates tiny segments, and the `LinearMemoryAlu` amortization refusal now applies to trap-terminal regions like any other (an earlier draft waived it; the restack dropped that exemption per review — a `TrapExit` ending adds trap dispatch and re-entry on top of the fixed entry costs, so the amortization economics only tighten). F-line, `TRAP #n`, exceptions, and non-sequential arrival keep the previous discard-the-recording behavior.

## Evidence

**Focused** (`microbench trap-segments` / `trap-segments-long`: DBRA loops with segments separated by host-emulated A-lines; 50 M instructions per run; interleaved pairs, arms differ only in this commit; within-pair ratios are the metric):

| shape | base → this PR (M instr/s, 3 pairs) | win |
|---|---|---|
| 5/4/2-op segments | 28.5→35.4, 31.1→36.3, 32.1→39.8 | +17–24 % |
| 9-op segments (the measured gameplay shape) | 34.9→50.3, 38.6→54.3, 39.4→57.4 | **+41–46 %** |

This is the experiment the decoded-run tier (#130) failed: native segments plus one probe/entry beat interpreting the run, where decoded replay did not — and the win scales with segment length, so the ~9-op trap-punctuated gameplay runs sit in the +40 % regime.

**Restack re-verification (2026-08-23):** the tables above were measured on the original stacked tree. Re-run on the restacked head against `01a4f7b` directly (same protocol: 3 interleaved pairs per shape, alternating order, 50 M instructions per run), the 9-op gameplay shape reproduces at **+39–50 %** (49.6→71.1, 53.7→74.4, 49.3→74.1 M instr/s) and the 5/4/2-op shape shows +6–21 % (39.1→41.4, 39.4→44.6, 40.1→48.6) — same ordering, with the short-segment spread widened by background host load. The bench segments are register-only, so the dropped amortization exemption does not affect them.

**Whole-app** (EV Override under Systemless, real GUI): four interleaved sessions (launch → title crawl ~20 s → flight ~20 s → close), baseline = systemless at a fixed commit + the pending PRs, candidate = the same + this PR. Metric: host work per guest instruction (macOS `/usr/bin/time -l` counters divided by the emulator's own retired-instruction total at close), which cancels session length and phase mix:

| pair | host instr / guest instr | host cycles / guest instr |
|---|---|---|
| 1 | 377.06 → 376.44 (−0.17 %) | 195.54 → 192.57 (**−1.52 %**) |
| 2 | 381.69 → 374.45 (−1.90 %) | 195.23 → 192.67 (**−1.31 %**) |

The baseline's cycles ratio is stable to ±0.2 % across three same-day sessions (194.81/195.54/195.23), so the consistent −1.3…−1.5 % cycles delta is well outside session noise.

Provenance: the whole-app pairs were collected on the pre-restack stacked tree; both arms shared the identical base (including then-open #128), so the delta isolates this commit's mechanism, but the exact figures predate the restack's two policy changes (the dropped amortization exemption above, and the re-record threshold now derived as `8 × TRACE_HOT_THRESHOLD` — the same value, 16). The focused microbench has been re-verified on the restacked head (previous section); the GUI pairs have not been re-collected.

**Boot-phase cost, disclosed:** headless boot A/B (400 M guest instructions, interleaved 3 pairs, host counters) bounds the m68k-side cost at ≈ +0.8 % host instructions during boot — boot's trap-punctuated code compiles segments but barely repeats them (census: +31.6 K native calls bought only +269 K retired). Two policy responses are in the commit: a chain quality gate (only a boundary that compiled seeds the next continuation) and deferred compilation (a head's first trap-boundary closure compiles nothing; it re-arms the head to count to `TRACE_TRAP_SEGMENT_HOT_THRESHOLD` — derived as `8 × TRACE_HOT_THRESHOLD` = 16 since the restack, so only re-proven-hot heads compile). The GUI numbers above are net of this tax.

**Where the rest of the win lives:** the microbench asymptote is the host trap dispatch round trip stage 1 deliberately keeps. Eliminating it (dispatching simple HLE traps from inside a running trace) is the stage-2 design sketched in the design doc — posted for discussion rather than implemented here.

## Tests

- A DBRA loop punctuated by two A-lines: all three segments compile (each verified to end at its boundary — `TrapExit` at the A-line's address, the tail on its real branch); executing the first segment retires exactly its real ops with `pc` parked on the trap; rewriting the trap word refuses stale execution (SMC). Runs under both the native and portable executors.
- `finish_recording_at_trap` refuses non-sequential arrival and non-A-line words, leaving the recording for the ordinary discard path.
- A `run_batch`-vs-`step` differential with host-emulated no-op traps stays bit-identical (pc/sr/all registers/memory/instruction counts) across 501 iterations.
- Full suites at the restacked head (re-run 2026-08-23): 839 default / 936 all-features; clippy `-D warnings`; fmt; CI green.

### DIFF
--- a/benches/microbench.rs
+++ b/benches/microbench.rs
@@ -1923,6 +1923,153 @@ fn bench_movem_frame_loop() {
     );
 }
 
+/// A trap-punctuated loop, a dominant profiled gameplay shape: short
+/// straight-line runs separated by A-lines the host emulates. The bench's
+/// dispatch loop plays the host (each A-line is a no-op that costs one
+/// batch round trip), so the number reflects segment execution plus the
+/// per-segment boundary -- the stage-1 trap-crossing question.
+fn bench_trap_segment_loop() {
+    const INSTRS: u64 = 50_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x5282, // head: ADDQ.L #1,D2
+        0x5283, // ADDQ.L #1,D3
+        0xD682, // ADD.L D2,D3
+        0x2003, // MOVE.L D3,D0- scratch
+        0x4A41, // TST.W D1
+        0xA123, // trap 1
+        0x5285, // ADDQ.L #1,D5
+        0xDA83, // ADD.L D3,D5
+        0x2005, // MOVE.L D5,D0
+        0x4A42, // TST.W D2
+        0xA124, // trap 2
+        0x5281, // ADDQ.L #1,D1
+        0x4A43, // TST.W D3
+        0x51CF, 0xFFE4, // DBRA D7,head
+        0x707F, // MOVEQ #127,D7
+        0x60DE, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(7, 0x7F);
+        cpu
+    };
+    let drive = |cpu: &mut CpuCore, bus: &mut LinearMemoryBus, budget: u64| -> u64 {
+        let mut retired = 0u64;
+        while retired < budget {
+            let chunk = (budget - retired).min(4_096) as u32;
+            let result = cpu.run_batch(bus, chunk, &[]);
+            retired += u64::from(result.instructions);
+            match result.exit {
+                m68k::BatchExit::BudgetExhausted => {}
+                m68k::BatchExit::AlineTrap { .. } => {} // host no-op dispatch
+                other => panic!("unexpected exit {other:?}"),
+            }
+        }
+        retired
+    };
+    let mut warm_cpu = prepare_cpu();
+    drive(&mut warm_cpu, &mut bus, 5_000_000);
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    let retired = drive(&mut cpu, &mut bus, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(2) > 5_000, "the loop actually iterated");
+    assert!(
+        cpu.d(1).abs_diff(cpu.d(2)) <= 1,
+        "the two per-iteration counters advance in lockstep"
+    );
+    println!(
+        "batch     trap segment loop       {:8.1} M instr/s",
+        retired as f64 / elapsed / 1_000_000.0
+    );
+}
+
+/// The measured real-world segment shape: ~9-op straight-line runs between
+/// A-lines (a dominant profiled gameplay pattern), so the boundary
+/// cost amortizes over twice the work of the short variant above.
+fn bench_trap_segment_loop_long() {
+    const INSTRS: u64 = 50_000_000;
+    const CODE_BASE: u32 = 0x6000;
+    let words = [
+        0x5282, // head: ADDQ.L #1,D2
+        0x5283, // ADDQ.L #1,D3
+        0xD682, // ADD.L D2,D3
+        0x2003, // MOVE.L D3,D0 - scratch
+        0x4A41, // TST.W D1
+        0x5284, // ADDQ.L #1,D4
+        0xD883, // ADD.L D3,D4
+        0x2004, // MOVE.L D4,D0
+        0x4A42, // TST.W D2
+        0xA123, // trap 1
+        0x5285, // ADDQ.L #1,D5
+        0xDA83, // ADD.L D3,D5
+        0x2005, // MOVE.L D5,D0
+        0x4A44, // TST.W D4
+        0x5286, // ADDQ.L #1,D6
+        0xDC85, // ADD.L D5,D6
+        0x2006, // MOVE.L D6,D0
+        0x4A45, // TST.W D5
+        0x5287, // ADDQ.L #1,D7 - clobbered by the DBRA counter reload path
+        0xA124, // trap 2
+        0x5281, // ADDQ.L #1,D1
+        0x4A43, // TST.W D3
+        0x51CF, 0xFFD2, // DBRA D7,head
+        0x7E7F, // MOVEQ #127,D7
+        0x60CC, // BRA.S head
+    ];
+    let mut bus = LinearMemoryBus::new(0x1_0000);
+    for (index, word) in words.iter().enumerate() {
+        bus.write_word_at(CODE_BASE + index as u32 * 2, *word);
+    }
+    let prepare_cpu = || {
+        let mut cpu = CpuCore::new();
+        cpu.set_cpu_type(CpuType::M68040);
+        cpu.set_sr(0x2700);
+        cpu.pc = CODE_BASE;
+        cpu.set_a(7, 0x8000);
+        cpu.set_d(7, 0x7F);
+        cpu
+    };
+    let drive = |cpu: &mut CpuCore, bus: &mut LinearMemoryBus, budget: u64| -> u64 {
+        let mut retired = 0u64;
+        while retired < budget {
+            let chunk = (budget - retired).min(4_096) as u32;
+            let result = cpu.run_batch(bus, chunk, &[]);
+            retired += u64::from(result.instructions);
+            match result.exit {
+                m68k::BatchExit::BudgetExhausted => {}
+                m68k::BatchExit::AlineTrap { .. } => {} // host no-op dispatch
+                other => panic!("unexpected exit {other:?}"),
+            }
+        }
+        retired
+    };
+    let mut warm_cpu = prepare_cpu();
+    drive(&mut warm_cpu, &mut bus, 5_000_000);
+    let mut cpu = prepare_cpu();
+    let start = Instant::now();
+    let retired = drive(&mut cpu, &mut bus, INSTRS);
+    let elapsed = start.elapsed().as_secs_f64();
+    assert!(cpu.d(2) > 5_000, "the loop actually iterated");
+    assert!(
+        cpu.d(1).abs_diff(cpu.d(2)) <= 1,
+        "the two per-iteration counters advance in lockstep"
+    );
+    println!(
+        "batch     trap segment loop long  {:8.1} M instr/s",
+        retired as f64 / elapsed / 1_000_000.0
+    );
+}
+
 /// PEA of an absolute address, rebalanced -- the profile's second-widest
 /// blocked head class after MOVEM. Base blocks the head at the PEA.
 fn bench_pea_abs_loop() {
@@ -2142,6 +2289,14 @@ fn main() {
     }
     if only.as_deref() == Some("movem") {
         bench_movem_frame_loop();
+        return;
+    }
+    if only.as_deref() == Some("trap-segments") {
+        bench_trap_segment_loop();
+        return;
+    }
+    if only.as_deref() == Some("trap-segments-long") {
+        bench_trap_segment_loop_long();
         return;
     }
     if only.as_deref() == Some("pea-abs") {
--- a/src/core/cpu.rs
+++ b/src/core/cpu.rs
@@ -367,6 +367,11 @@ pub struct CpuCore {
     // trace, so eviction can't wedge them.
     #[cfg_attr(feature = "serde", serde(skip))]
     pub(crate) trace_record_skip: [u32; 4],
+    /// Expected resume PC after an A-line that closed a trap-boundary
+    /// recording; consumed at the next batch entry to seed the
+    /// continuation segment's head candidacy. Transient host-side state.
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub(crate) pending_trap_resume: Option<u32>,
     #[cfg_attr(feature = "serde", serde(skip))]
     pub(crate) trace_probe_skip: [u32; 4],
     #[cfg_attr(feature = "serde", serde(skip))]
@@ -526,6 +531,7 @@ impl CpuCore {
             fm_base: 0,
             fm_len: 0,
             trace_record_skip: [super::trace_jit::TRACE_PC_NONE; 4],
+            pending_trap_resume: None,
             trace_probe_skip: [super::trace_jit::TRACE_PC_NONE; 4],
             trace_record_skip_at: 0,
             trace_probe_skip_at: 0,
--- a/src/core/execute.rs
+++ b/src/core/execute.rs
@@ -620,6 +620,18 @@ impl CpuCore {
         let mut retired: u32 = 0;
         let mut probe_on_entry = true;
 
+        // A recording that closed at an A-line on the previous batch named
+        // its expected continuation; if the host resumed exactly there, the
+        // continuation becomes a trace-head candidate (counting, then
+        // recording, like any backward-branch target). A host that resumed
+        // anywhere else silently drops the hint.
+        if let Some(expected) = self.pending_trap_resume.take()
+            && expected == self.pc
+        {
+            trace_jit::record_trace_target(self.pc, self.cpu_type);
+            probe_on_entry = true;
+        }
+
         loop {
             // The trace JIT's headroom guard compares against
             // `cycles_remaining`; keep it topped up so it can never gate a
@@ -715,10 +727,36 @@ impl CpuCore {
                 }
             };
             if let Some(exit) = exit {
-                // The caller may emulate a surfaced trap and resume at an
-                // unrelated guest PC. Never let an in-progress path recording
-                // cross that host-controlled execution boundary.
-                trace_jit::stop_recording(self, trace_jit::RecordingStop::TrapOrException);
+                // An A-line reached by falling through a recorded region is
+                // a trap-boundary terminal: the region compiles ending at
+                // the trap and the expected resume point (the word after
+                // the A-line) gains head candidacy on the next batch entry,
+                // so segments chain themselves down a trap-punctuated run
+                // (docs/trap-crossing-traces-design.md). Everything else --
+                // F-line, TRAP #n, exceptions, non-sequential arrival --
+                // keeps the previous behavior: the caller may resume at an
+                // unrelated guest PC, so the recording is discarded rather
+                // than allowed to cross a host-controlled boundary.
+                if matches!(exit, BatchExit::AlineTrap { .. }) {
+                    match trace_jit::finish_recording_at_trap(self) {
+                        // Chain quality gate: only a segment the compiler
+                        // accepted right up to this boundary earns a seeded
+                        // continuation; a rejected closure must not extend
+                        // the head chain past code that will never pay.
+                        trace_jit::TrapFinish::Compiled => {
+                            self.pending_trap_resume = Some(self.ppc.wrapping_add(2));
+                        }
+                        trace_jit::TrapFinish::Closed => {}
+                        trace_jit::TrapFinish::None => {
+                            trace_jit::stop_recording(
+                                self,
+                                trace_jit::RecordingStop::TrapOrException,
+                            );
+                        }
+                    }
+                } else {
+                    trace_jit::stop_recording(self, trace_jit::RecordingStop::TrapOrException);
+                }
                 return BatchResult {
                     instructions: retired,
                     exit,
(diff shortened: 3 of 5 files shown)


rlanday:
From Claude Fable 5 — the design doc behind #134, posted for the stage-2 conversation as much as the stage-1 record. Stage 1 (this PR) keeps every host trap round trip; stage 2 (§3B) is the design that would remove them, and it needs your take on the HLE dispatch surface before it's worth prototyping.

---

# Trap-crossing traces: design

*Claude Fable 5, 2026-08-17. Status: stage-1 prototype implemented and
measured (see §7); stages 2+ pending discussion with Ben.*

## 1. The problem, quantified

After the 2026-08-17 optimization pass (systemless #662–#669, m68k-rs
#111/#112/#128/#131/#133), the dominant remaining cost in EV Override's
gameplay phases is the interpreter executing guest instructions the
trace JIT cannot cover:

- Run 10 (Instruments, 65 s boot→crawl→flight): "runner loop self" =
  30–35 % of the main thread in crawl and flight, plus m68k core self
  ~22 % — over half of active main-thread time is per-instruction
  interpretation.
- Line-level attribution shows it is not one hot spot: `cached_decode`
  table loads, per-op execute, endian swaps, mem-op resolution — the
  interpreter's distributed per-instruction overhead (~100+ host
  cycles per guest instruction against a compiled trace's ~handful).
- The trace-profile census puts *admission* fixes (heads blocked by an
  unsupported opcode) at ~2 % of guest dispatch. The other ~98 % of
  interpreted execution never becomes a candidate at all.

Why: **traces require loops.** Heads arise only at backward-branch
targets, and a region only compiles if it ends in a branch
(`NoTraceTerminal` otherwise). EV Override's hot code is
trap-punctuated: the measured crawl executes ~9-instruction runs
between A-line traps (66 M nine-instruction batches in one crawl
session). A straight-line run from one trap to the next contains no
backward branch and no closing branch — structurally invisible to the
JIT, forever.

## 2. Prior evidence that shapes the design

- **Decoded-run tier (m68k-rs #130, closed):** decoding trap-to-trap
  runs into a replayable form was built and measured *null* whole-app:
  at ~9 instructions per run, per-batch entry/exit overhead (state
  sync, probe, dispatch) cancels the per-instruction win. Any design
  that pays a host-boundary round trip per short segment inherits this
  result.
- **Wait elision (systemless #644):** trap-punctuated *poll* cycles are
  already parked by the exact-cycle prover at the systemless level, so
  compiling them is not double work — a compiled poll segment burns
  fewer host cycles per iteration during the (bounded) unproven period.
- **JIT batch-entry cost** (decoded-run-tier design doc §batch-entry):
  probe + validate + `call_native` + state sync ≈ tens of ns — the
  budget a 9-op segment must amortize.

## 3. Design space

**A. Segmented traces (stage 1, prototyped).** Keep the existing
"trace = compiled region, host dispatches traps between regions"
contract. Add:
1. a `TrapExit` trace terminal: a region may end *at* an A-line; the
   A-line word itself is recorded as the terminal op (its two bytes are
   real guest memory, so SMC validation stays byte-exact), execution of
   the terminal sets `pc` to the trap's address and exits the trace
   normally; the batch loop then fetches the A-line and surfaces it to
   the host exactly as today (zero systemless changes);
2. post-trap head seeding: when a recording is stopped by a trap (or a
   `TrapExit` trace completes), the expected resume PC (trap + 2) is
   remembered on the CPU; if the next batch entry lands there, that PC
   gains head candidacy. Segment heads chain themselves down a
   trap-punctuated run; a loop whose body contains traps becomes
   [head..trap₁) [trap₁+2..trap₂) … [trapₙ+2..closing-branch], each an
   ordinary trace.

Costs one probe+entry per segment — exactly the overhead the tier
experiment measured. Stage 1 therefore measures whether *native*
segments (vs the tier's replayed decode) clear the bar, and its
numbers bound the boundary overhead empirically for stage 2.

**B. Inline HLE dispatch (stage 2, design only).** Native trace code
calls back into the host trap dispatcher mid-trace (a `extern "C"`
trampoline carrying the opcode; the dispatcher runs against synced
state; the trace resumes at the next op). Removes the per-segment
boundary entirely — one probe+entry covers a whole trap-punctuated
region. Requires: a re-entrancy contract for the dispatcher (it may
mutate PC/registers/memory arbitrarily — the trace must revalidate or
bail on any deviation from the recorded resume PC), a way to thread
the `&mut dyn AddressBus` + dispatcher context through `call_native`,
and agreement that HLE handlers never re-enter the CPU (today's
handlers do not, but nothing enforces it). This is the architectural
conversation with Ben; stage 1's numbers tell us whether it is worth
having.

**C. Not pursued:** whole-region AOT of trap-punctuated code
(recompilation cost, SMC exposure), and interpreter micro-optimization
(measured today: decode-table repacks null after boot; the cost is
genuinely distributed).

## 4. Stage-1 semantics (as implemented)

- `JitTraceOp::TrapExit` — terminal; recorded with the A-line's own
  opcode word at the A-line's PC. `ends_trace() == true`;
  `max_cycles == 0`; retiring it does **not** count a guest
  instruction and charges no cycles; it stores `pc = trap_pc` and ends
  the trace.
- Recording: `run_batch_inner`'s A-line exit path, where recording was
  previously discarded (`RecordingStop::TrapOrException`), now — for
  A-line traps only, with a live recording whose ops all precede the
  trap — appends `TrapExit` and finishes with
  `RecordingEnd::TrapBoundary`, which compiles.
- Compile gates: `TRACE_MIN_OPS` still applies (a too-short segment is
  rejected and remembered structurally, as before). The
  `LinearMemoryAlu` amortization refusal is *waived* for
  `TrapBoundary` regions — deliberately, so the measurement reflects
  segmentation itself rather than a policy interaction; if stage 1
  ships, this policy gets its own measurement.
- Seeding: `CpuCore::pending_trap_resume` is set (trap_pc + 2) when a
  recording ends at a trap **and a compiled segment now really reaches
  that boundary** (`TrapFinish::Compiled`); consumed at batch entry by
  comparing against the entry PC and granting candidacy through the
  ordinary `record_trace_target` path (Counting → threshold → record).
  A host that resumes anywhere else (LoadSeg, ExitToShell, …) simply
  never matches, and the stale expectation is cleared on the next batch
  entry.
- **Chain quality gate** (added after the first headless A/B): the
  first build seeded the continuation whenever a recording *closed* at
  a trap, even when the closure then rejected (TooShort etc.). In the
  real app that extends head chains through code the compiler has
  already refused — every trap crossing plants another candidate, and
  the steady-state samples showed `finish_recording_at_trap` +
  `stop_recording` still hot late into boot (record/reject churn plus
  direct-mapped alias pressure). The gate seeds only past a boundary
  that has actually paid: a rejected or salvaged-short closure returns
  `TrapFinish::Closed` and the chain stops there.
- F-line and other exceptions: unchanged (recording still discarded).

## 5. Correctness argument

- The terminal's bytes are the guest's own A-line word: SMC validation
  covers the full recorded byte range including the boundary, so a
  rewritten trap word invalidates the trace like any other patched
  instruction.
- State at exit is the same as the interpreter's at the same point: a
  trace exit already syncs registers/flags/PC (prefix-salvage relies
  on this); `TrapExit` adds nothing new — PC lands on the A-line and
  the existing Miss → full dispatch → `BatchExit::AlineTrap` path runs
  the host trap exactly as before.
- Seeding is only a *candidacy* hint; a wrong hint creates a Counting
  entry that either never fires or records a region that stands on its
  own merits.

## 6. Risks

- **Net-null (the tier's fate):** per-segment boundary cost eats the
  win at 9-op segments. This is exactly what stage 1 measures; a null
  here with native segments is strong evidence that only stage 2 can
  reach the interpreted share.
- Trace-cache pressure: segment heads are new 4096-slot direct-mapped
  entries; alias churn (the #100 class-2 mechanism) could evict loop
  heads. Mitigation if seen: cap seeded heads or tag them for
  preferential eviction.
- Compile-time growth at boot (more regions compile): bounded by
  #128's verifier removal; measured in the whole-app arms.

## 7. Measurement plan / results

Arms: systemless v0.16.0 fixed; m68k `perf/pending-combined` (the
five pending PRs, already measured at −7.8 % host cycles per guest
instruction) vs that plus stage 1. Metrics: host instructions & cycles
**per guest instruction** over one full comparable session per arm
(the ratio protocol from the 2026-08-17 proof), plus the trace-profile
census (compiled-segment counts, churn) and boot triples.

**Focused microbench** (two DBRA loops with segments separated by
host-emulated A-lines; 50 M instructions per run; interleaved pairs on
a busy box, so within-pair ratios are the metric):

| shape | baseline -> stage 1 (M instr/s, 3 pairs) | win |
|---|---|---|
| short: 5/4/2-op segments | 28.5->35.4, 31.1->36.3, 32.1->39.8 | +17-24 % |
| long: 9-op segments (the measured gameplay shape) | 34.9->50.3, 38.6->54.3, 39.4->57.4 | **+41-46 %** |

Segmentation clears the boundary-cost bar that nulled the decoded-run
tier, and the win scales with segment length exactly as the
boundary-cost model predicts: at the crawl's measured ~9-op runs the
mechanism sits in the +40 % regime. The remaining asymptote is the
host trap dispatch round trip itself -- stage 2's territory. Whole-app numbers pending the approved GUI pair.

**Headless boot A/B** (400 M guest instructions of real EV Override
boot, `--headless`, interleaved 3 pairs, host counters from
`/usr/bin/time -l`): the *ungated* first build regressed — host
instructions +0.78 % (1077.8 G → 1086.1 G, stable to 4 digits across
pairs) and host cycles +3–8 %, the cycle-vs-instruction asymmetry
pointing at cache damage from head-chain churn. Steady-state samples
showed `finish_recording_at_trap`/`stop_recording` still hot late in
boot and no cranelift frames at all (compile time innocent). The chain
quality gate (§4) is the response; its own interleaved pair is the
gating result for whether stage 1 proceeds to GUI measurement.
**Gated-pair result:** the gate did *not* recover the headless-boot
cost — the interleaved pair (same protocol) still shows +0.81 % host
instructions (1077.9 G → 1086.8 G) with cycles +1–10 %. Census diff
attributed it: +31.6 K native calls bought only +269 K retired (boot's
trap segments compile but barely repeat — 17/40 sampled compiled
shapes are trap-terminal, none reach the top-40 retired table), and a
full-run sample split the added spend three ways: closure handling
(+559 samples), probe cost (+519), compile machinery (+593, 3.6×
baseline cranelift).

**Deferred compilation** (in the final commit): a head's first
trap-boundary closure compiles nothing — it re-arms the head to count
with `TRACE_TRAP_SEGMENT_HOT_THRESHOLD = 16`; only heads that
re-prove that hot record again and compile. Its interleaved pair still
shows +0.40/+0.68/+1.24 % host instructions (mean ≈ +0.8 %, box
noisier): the residual is the recording/probe tax itself on boot code
that never repeats, not compile spend. Conclusion: headless boot
bounds the cost at ≈ +0.8 % m68k-side host instructions; the win this
design exists for lives in the GUI gameplay phases (the measured ~9-op
trap-punctuated runs at 66 M batches/session), which headless boot
cannot exercise. The GUI proof pair decides.

**GUI proof pair (the deciding measurement).** Four sessions run by
Ryan, interleaved (baseline = systemless-m68kprs: systemless 7fa0789 +
m68k pending-combined; candidate = systemless-trapseg3: + the stage-1
commit). Each session: launch, crawl ~20 s, flight ~20 s, close.
Metric: host work per guest instruction (`/usr/bin/time -l` counters /
"Total instructions" at close), which cancels session length and phase
mix:

| pair | host instr / guest instr | host cycles / guest instr |
|---|---|---|
| 1 | 377.06 → 376.44 (**−0.17 %**) | 195.54 → 192.57 (**−1.52 %**) |
| 2 | 381.69 → 374.45 (**−1.90 %**) | 195.23 → 192.67 (**−1.31 %**) |

The baseline's cycles ratio is stable to ±0.2 % across three sessions
the same day (194.81 / 195.54 / 195.23), so the consistent −1.3…−1.5 %
cycles delta is ~6× the observed session noise: stage 1 is a real
whole-app CPU reduction, net of the ≈+0.8 % boot-phase tax. The
remaining headroom (the trap dispatch round trip itself) is stage 2's
territory.

rlanday:
From Claude Fable 5, working with @rlanday — an update on where this line of work led, so the design-doc comment above doesn't point at a dead end.

**This PR (stage 1) stands on its own** and is unchanged: trap-punctuated segments compile via `TrapExit` terminals, GUI proof pair shows −1.3…−1.5% host cycles per guest instruction net of the ≈+0.8% boot tax. Nothing below changes that; it's ready to review on its own merits.

**Stage 2 (inline HLE dispatch, §3B of the design doc) was prototyped and measured null — do not pursue it.** The prototype works (microbench 6–9× on the trap-segment loops, the boundary-removal ceiling), but whole-app it is −0.4% ± 1.3% host instructions, i.e. a null. The reason is structural: SANE (`$A9EB`, the trap we'd inline) is 57% of *dispatches* but only **14 samples** of a full-run flight profile — the per-trap boundary is a rounding error of whole-app CPU, so removing it can't move the needle. Stage 1 keeps the trap round trip and that turned out to be fine.

**Where the gameplay CPU actually is (execution-weighted flight census).** Chasing the null, we attributed interpreted execution by where it happens rather than by dispatch frequency. Two findings:

- Trace *coverage* in flight is ~46%. The entire op-admission + call-through + call-target-seeding opportunity is ~2–3% combined (all measured, all small/overlapping). Coverage work is largely spent.
- **95% of interpreted execution is guard-exit / alternate-path bleed inside loops that are *already* compiled**, and **64% of all interpreted execution is one loop** — EV Override's software audio mixer (`0x0118A12C`), a per-sample clamp/pack/store. Its four `if (x>127) x=127;`-style clamps are data-dependent skip-one-instruction forward branches; the JIT compiles *linear* traces, so each clamp mispredict side-exits to the interpreter (~72% of that loop's iterations).

**The real follow-on is if-conversion, not stage 2.** Compiling short forward conditional branches that skip register-only ops as branchless `select`/`smin`/`smax` (the primitives `emit_guarded_branch` already uses) makes the clamp loop single-path and fully JITable — recovering ~25% of all guest instructions from one change. Writeup with the reconstructed guest C and the mechanism is drafted; happy to open it as a separate PR.

Net: merge/close this on the stage-1 evidence; stage 2 is closed as a measured null; if-conversion is the next lever.

benletchford:
The `TrapExit` / trap-boundary segmentation mechanism is interesting and much of it is generic, but I don’t think the current stacked PR is ready to merge as a generic library improvement.

It currently carries the pending stack including #128 and #131, both of which I’m holding for safety/correctness reasons, so it needs to be restacked onto current `master` without those before this change can be evaluated independently.

I’d also separate the correctness mechanism from workload-specific policy tuning. In particular, the deliberate `LinearMemoryAlu` amortisation-policy waiver and `TRACE_TRAP_SEGMENT_HOT_THRESHOLD = 16` are justified here from the EV Override measurements. That is useful experimental evidence, but I don’t want an application-tuned threshold/exception to become the default library policy without broader justification. Either make those decisions demonstrably generic across representative workloads or expose them as tunable/opt-in policy, while keeping the core TrapExit/resume/SMC semantics and differential tests generic.

Please restack onto current `master` after the accepted generic changes, exclude #128/#131, preserve the trap-boundary correctness tests, and rerun the full Rust CI.

benletchford:
Triage decision: leaving this open, but not merging the current head. The core `TrapExit` idea remains worth reviewing; the present branch is conflicted and still carries #128’s verifier-off change, which was explicitly declined, plus workload-tuned amortisation/threshold policy that has not been separated or made opt-in. Please restack only the generic trap-boundary mechanism and its correctness tests onto current `master`, then rerun CI.

rlanday:
Restacked per the triage instructions — the head is now a single commit on current `master` (v0.11.0) containing the TrapExit mechanism, both executors, and the correctness tests, nothing else:

- **No #128 carry** — the verifier-off change is gone with the old stack.
- **The amortisation-gate exemption for trap-ending segments is dropped**, not made opt-in: on reflection it wasn't defensible even as a knob — a TrapExit ending adds trap dispatch and re-entry on top of the fixed validation/ABI costs, so the economics the gate models only get tighter for trap segments. They now face the same gate as every other region.
- **The deferred-trap re-record threshold is no longer a standalone tuned constant**: it's expressed as `8 × TRACE_HOT_THRESHOLD` with the structural rationale in the comment (the value only needs to exceed how often one-shot startup code repeats while staying trivial for repeating gameplay loops — any small multiple of the base threshold satisfies that).
- Composes with the merged durable-rejection work (the strike-era `Counting` constructors carry the new `deferred_trap` state). Full suite locally: 258 pass. — Claude Fable 5

rlanday:
A measurement note from testing this PR alongside the other open
branches, recorded here so the PR history explains it: **headless census
numbers taken on this branch before 2026-08-24 ran ~4% less guest time
than other arms, and the cause was our measurement harness, not this
PR.**

**Observation.** On the fixed 400M-instruction SimCity 2000 headless
census, a build of this PR's head alone ended at guest tick 113503 where
`master`, #141, and #142 all ended at 118533. The per-screenshot tick
curve showed the entire −5,030-tick divergence accumulating during the
first ~25M instructions (boot); afterwards both arms advanced ticks at an
identical rate.

**Root cause (harness-side).** The Systemless harness proves boot wait
loops are exact idle cycles — same trap site, identical CPU snapshot,
write journal restored — and fast-forwards the guest clock for each
proven cycle. The snapshot compared `change_of_flow`, the m68k core's
internal did-the-last-instruction-branch bookkeeping, which is a trace
and loop-mode heuristic input rather than architectural state. Its value
at a trap site depends on whether execution arrived via the interpreter
or a compiled trace. This PR's `TrapExit` traces cover the boot wait's
trap-punctuated segments, so arrivals flipped the flag, identity proofs
failed on that single field, and each lost proof was a lost tick
fast-forward: over the first 40M instructions, successful exact-repeat
proofs dropped 60,645 → 55,520 (−5,125 ≈ the −4,995 lost ticks) while
CPU-state proof failures rose 43,888 → 55,636. Field-level sampling
showed `change_of_flow` was the *only* differing field in every sampled
failure — in both arms; the baseline's 43.9k background failures were
the same flag at other sites.

**Resolution, verified.** The harness snapshot now normalizes
`change_of_flow` (it was never restored from the snapshot, only
compared). With the fix, a build of this PR's head and a build of the
#141+#142 stack produce **bit-identical harness behavior** on the 40M
boot window — final tick 84,328 in both arms with every wait-stats
counter equal (exact repeats 75,906; CPU-state failures 1,458, down from
43,888/55,636). Two bonuses: the baseline itself recovered ~15k ticks of
boot fast-forwarding it had been losing to the same flag (its arrivals
also mix traced and interpreted execution), and prover round trips
nearly halved. The emulation was never wrong — `TrapExit` preserved the
host trap protocol exactly as the PR body states; the harness was
over-strict about non-architectural state.

One residual note for review: any *external* consumer of proven-idle
parking with similarly strict state comparison would see the same
interaction whenever a JIT compiles part of a wait loop; the general
rule this suggests is that wait-identity proofs should compare
architectural state only.

PR #138, opened 2026-08-19
perf(jit): expand traces with effective-address ops and short-branch if-conversion

rlanday:
## What this does

This PR lets the trace JIT keep both sides of a small forward conditional branch in one compiled trace. It also admits two previously unsupported address-generation forms that appeared frequently in the gameplay census:

- `LEA (xxx).W/.L,An` — put an absolute address in an address register.
- `PEA (An)` — push the address already held in an address register.

The result is fewer guard exits and longer useful traces without speculatively executing guest instructions from the wrong branch.

This is stacked on #134 (`5108b70`) as requested. The two commits are mechanically independent of `TrapExit`; if #134 is declined, they can be retargeted to `master` without changing the implementation.

### The small amount of 68k terminology used below

- `D0`–`D7` are data registers. `A0`–`A7` are address registers. `A7` is the stack pointer.
- The condition-code register contains flags such as carry (`C`) and zero (`Z`). Conditional branches test these flags.
- A 68k word is 2 bytes. Most instructions begin with one word and may have extension words after it.
- `PC` is the address at which execution continues. In a JIT side exit, `PPC` identifies the guest instruction that caused the exit.
- “Retired” means a guest instruction actually executed. An instruction skipped by a branch must not be counted.

### If-conversion

Consider this short forward branch (addresses are illustrative):

```asm
$0100  BCS.S  $0104       ; if C = 1, skip the next instruction
$0102  MOVEQ  #99,D3      ; runs only when C = 0
$0104  BNE.S  $0108       ; a later branch may still leave the trace
```

Previously the recorder committed to the branch direction it happened to observe. Seeing the other direction later caused a guard exit, so a tiny diamond in a hot loop could repeatedly terminate the trace.

The recorder now represents the first branch as `CondSkip`. It records:

- which 68k condition is being tested;
- how many following trace operations belong to the fall-through side;
- whether the original branch was the 2-byte `.S` or 4-byte `.W` form, so cycle accounting remains exact.

Native code emits a real conditional `brif` around those operations. They are not executed speculatively: with `C = 1`, `MOVEQ` is skipped; with `C = 0`, it runs normally. The portable executor implements the same decision so both execution engines can be compared directly.

The regression test fixes `Z = 0` so the later `BNE` exits, then checks both values of `C`:

| Initial state | Result for `D3` | Retired through the later exit | Cycles | Exit location |
|---|---:|---:|---:|---|
| `C = 1`, `D3 = 7` | `7` (the move was skipped) | 2 (`BCS`, `BNE`) | 20 | `PC=$0108`, `PPC=$0104` |
| `C = 0`, `D3 = 7` | `99` | 3 (`BCS`, `MOVEQ`, `BNE`) | 22 | `PC=$0108`, `PPC=$0104` |

The cycle difference is the real 68000 behavior: a taken conditional branch costs 10 cycles; a not-taken short branch costs 8; `MOVEQ` costs 4. A separate test covers the word-displacement form, whose not-taken cost is 12 cycles.

The recorder recognizes both observation orders. If the branch was first recorded as taken, it discovers the skipped instructions by decoding forward to the join. If it was recorded as not taken, those instructions have already been observed; the recorder converts the existing run and resumes at the join without recording it a second time.

### `LEA` and `PEA`

`LEA` means “load effective address.” For example:

```asm
LEA  ($0340).W,A2
```

puts address `$00000340` in `A2`. It does not read memory at `$0340`, and it does not change the condition flags. The `.W` form sign-extends a 16-bit address; the `.L` form carries a full 32-bit address. These are therefore simple register-producing operations for the JIT.

`PEA` means “push effective address.” For example, starting with `A2=$00000340` and `A7=$00000800`:

```asm
PEA  (A2)
```

must leave `A7=$000007FC` and bytes `00 00 03 40` at `$07FC..$07FF`. It must not dereference `$0340`, consume an extension word, or change the flags. `PEA (A7)` is the subtle case: it pushes the old value `$00000800` and only then leaves `A7=$000007FC`.

During review, `PEA (An)` was found to have been represented as an address mode with an extension word even though this encoding has none. That could advance the guest `PC` incorrectly and calculate the wrong value. It now has its own `PeaInd` trace operation in both executors. Tests place another valid opcode immediately after `PEA` so consuming a nonexistent extension word cannot pass unnoticed, and cover both an ordinary address register and `A7`.

This was a bug in the unmerged PR implementation, not in upstream. Upstream currently contains neither this `PEA (An)` admission nor the new absolute `LEA` admission, so there is no merged LEA change requiring a separate corrective PR.

### Safety boundary

Conversion is limited to a clean forward region of at most 4 operations and 12 guest bytes. Every instruction must tile the skipped byte range exactly and belong to an explicit allowlist. Calls, traps, nested control flow, `MOVEM`, stack-changing operations, and other trace terminators remain ineligible.

Some admitted operations access memory, but they remain inside the generated conditional block and execute only on the architecturally selected path. The limits are conservative compile-size policy, not workload-specific matching.

## Guard-exit classification cost

Once a trace can contain a `CondSkip`, the retired count is data-dependent, so guard exits are classified from architectural state (`PPC` plus the legal successor addresses) rather than an operation index. That classification scans the trace's operations on every partial return. Compilation now also records a 128-bit mask with a bit set only for operations holding a runtime branch prediction, so guard-free traces reject after one zero check and guarded traces visit only their guarded operations. This matches the classifier shape used by the dispatcher work in #141; on this base it changes classification speed only, and the deterministic census below reproduces the pre-mask arm bit-for-bit.

One known interaction for whoever lands #141 after this PR: a run whose `CondSkip` block was skipped retires fewer instructions than the trace's operation count, and "partial call" detection remains retirement-based (`retired < op count`). Benign here — it costs one extra host entry per native call and nothing else — but combined with #141's link exits it would suppress link chaining, because clean-link detection requires a non-partial call. The second PR to land should redefine "partial" by control flow (for example a reserved bit in the packed native return) instead of instruction count.

## Correctness fixes made during review

- Added the distinct, extension-free `PeaInd` operation described above.
- Carried a runtime retirement total into a guard after `CondSkip`. A static trace index is wrong because one direction executes the inline operations and the other direction skips them.
- Classified a guard exit from the exiting branch's `PPC` and legal successor addresses instead of treating that data-dependent retirement total as an operation index.
- Preserved `.S` versus `.W` branch length so both portable and native execution charge the correct taken/fall-through cycles.

The prior retirement error did not change guest register or memory results. It did inflate profiling: a later native guard could report a skipped instruction as retired, and the coverage census uses retired guest instructions as its numerator. The original SC2K `+2.93`-point result is therefore withdrawn; a fresh binary built from the corrected tree measures `+0.82` points.

## Tests and checks

- 167 library tests with `--no-default-features`
- 233 library tests with `--features jit`
- 267 library tests with `--features jit,trace-profile`
- `cargo clippy --all-targets --features jit,trace-profile -- -D warnings`
- `cargo fmt --check` and `git diff --check`

New regression coverage includes:

- portable `PEA (An)` with a following opcode that detects false extension-word consumption;
- native/portable `PEA (A2)` and `PEA (A7)` parity;
- both `CondSkip` directions followed by a real guarded branch, with exact result, retirement, cycles, `PC`, `PPC`, and exit classification — including the compiled trace's guard-mask contents;
- exact short- and word-displacement fall-through cycle bounds;
- recorder detection in both observation orders and native/portable parity for the supported shapes;
- rejection at the operation/byte limits and for unsafe or non-tiling regions.

## Whole-application measurement

The metric is JIT-retired guest instructions divided by a fixed 400-million-instruction budget. It is coverage, not wall-clock speed. EV Override's guest timers use wall time, so base and arm runs were interleaved in pairs; running each arm as a separate block had overstated the result. SC2K was deterministic in this run.

The arm was rebuilt from the final source tree (`eaec29d7db0ceba92a27219ae4c3b7285c300fd8`, which adds only the precomputed classification mask; census binary MD5 `6477f240e67af2af1df5b402a8177fd2`). The mask is classification-speed-only: on the deterministic SC2K workload the masked arm reproduces the pre-mask arm exactly — `187,333,811` JIT-retired instructions and `12,342,135` native calls in both builds.

| Workload | Pair | #134 base | This PR | Delta |
|---|---:|---:|---:|---:|
| EV Override | 1 | 72.80% | 74.02% | **+1.22 pts** |
| EV Override | 2 | 73.07% | 75.10% | **+2.03 pts** |
| EV Override | 3 | 73.00% | 75.22% | **+2.22 pts** |
| SimCity 2000 | 1 | 46.01% | 46.83% | **+0.82 pts** |

At EV head `$0118A136`, guard exits fall from `189,483–189,993` to zero. Average retired operations per invocation rise from about `25.6` to `53.15`. The head itself adds about `5.3M` retired instructions; the remainder of the whole-application delta comes from other trace changes.

For SC2K, total JIT retirement rises by `3,296,246` instructions. Two heads that appear in the arm's top 40 (`$008D2812` and `$008D281C`) retire about `2.72M` and `2.46M`, but those are not independent additions: `$008D26F0` falls by about `1.84M`, and other heads move as work is consolidated into longer traces. The whole-application number already nets those shifts out.

Coverage rising does not by itself prove a speedup. `CondSkip` adds a host conditional branch and dynamic accounting, so wall-clock or host-cycle performance should be measured separately from this census.

## Why the impact is not larger

The converted EV head is highly reproducible, but it represents only about 1.3 points of a 400M-instruction run. The earlier expectation of a much larger recovery also included durable outer-loop rejection behavior that #134's base already fixes. This PR addresses small forward diamonds; it cannot capture arbitrary side paths, calls, or dispatch loops, so the remaining interpreted share is not one hidden instance of the same problem.

The `4 op / 12 byte` limits may leave some compatible diamonds unconverted. The useful next step is execution-weighted rejection-reason profiling, followed by a measured cap sweep. Increasing the limits without that data risks larger native functions and more compile work for cold shapes while missing the hot structural blockers below.

### Related work kept out of this PR

SC2K's bytecode dispatcher is a materially larger lever than short forward diamonds: its missing path uses PC-relative indexed `MOVE` operations and computed `JMP` dispatch/case bodies, which this PR does not touch. That work — link exits, PC-relative `MOVE` sources, guarded computed jumps, and the entry-cost reductions they expose — is published separately as #141 with its own measurements. Hot side-path capture (heads such as `$008C7642` and `$008F12B0`, ~125K guard exits each in the corrected SC2K run) and execution-weighted conversion diagnostics for the `CondSkip` caps remain open follow-ups after both PRs.

## Commit structure

The branch contains exactly two commits on #134's head:

1. admit `LEA (xxx).W/.L` and corrected `PEA (An)` into traces;
2. add bounded forward-branch if-conversion, its accounting/tests, and the precomputed guard-classification mask.

The fixture repairs are folded into the feature commit because the feature alone changes the branch shapes those accounting fixtures were meant to exercise. Keeping them separate would leave an invalid bisect point.

Original implementation by Claude Fable 5; rebasing and measurement by Claude Opus 5; review, correctness fixes, the guard-mask classifier, regression tests, and a corrected census by GPT-5.6; verification of the mask port, refreshed paired measurement, and this revision by Ox Alpha. The corresponding co-authorship trailers are preserved in the commits.

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -227,6 +227,24 @@ pub(crate) enum JitTraceOp {
     /// the trace. The batch loop then fetches the A-line and surfaces it
     /// to the host exactly as an interpreted run would.
     TrapExit,
+    /// If-conversion of a short forward conditional block: the recorded
+    /// op is the Bcc word at its own pc (SMC-covered). The following
+    /// `skip_ops` trace ops are a conditional block executed only when the
+    /// branch is NOT taken (`condition` false); when taken they are
+    /// skipped. Retires one guest instruction for the branch plus one per
+    /// skip op actually executed -- data-dependent, so a trace containing a
+    /// CondSkip reports a runtime-computed retired count.
+    ///
+    // Constructed by the recorder in step 2 (native codegen + detection);
+    // step 1 lands the op, the portable executor, and its semantics test.
+    #[allow(dead_code)]
+    CondSkip {
+        condition: u8,
+        skip_ops: u8,
+        /// Encoded Bcc length: 2 for Bcc.S, 4 for Bcc.W. The 68000 charges
+        /// different fall-through cycles for the two encodings.
+        length: u8,
+    },
     Nop,
     MoveReg {
         src: JitDirectReg,
@@ -327,6 +345,15 @@ pub(crate) enum JitTraceOp {
     /// no condition codes.
     LeaIndex {
         src: JitEa,
+        dst: u8,
+        cycles: i32,
+    },
+    /// LEA `(xxx).W`/`(xxx).L,An`. The absolute address is sign-extended
+    /// (word form) or taken whole (long form) while recording, so
+    /// execution loads a constant into the address register -- no memory
+    /// access, no condition codes.
+    LeaAbs {
+        address: u32,
         dst: u8,
         cycles: i32,
     },
@@ -498,6 +525,12 @@ pub(crate) enum JitTraceOp {
         reg: u8,
         displacement: i16,
     },
+    /// `PEA (An)`: push the address register's value through the checked
+    /// window. This has no extension word and costs four fewer 68000 cycles
+    /// than the displacement form.
+    PeaInd {
+        reg: u8,
+    },
     /// `PEA (d16,An)`: push the effective address through the checked
     /// window. The address is computed from the pre-decrement stack pointer
     /// state, no condition code changes, and every check precedes the store
@@ -656,6 +689,11 @@ struct CompiledTrace {
     /// between them is deliberately unguarded.
     callee_start: u32,
     callee_end: u32,
+    /// Bit `n` is set when operation `n` is a recorded branch guard. Trace
+    /// entry can reject guard-free partial returns after one zero check, and
+    /// guarded traces inspect only the guarded operations instead of walking
+    /// up to 128 recorded operations.
+    guarded_ops: u128,
     /// A recorded interior branch is a path prediction eligible for adaptive
     /// rerecording. Cleared after the one allowed rerecord so completed traces
     /// stay off the accounting path.
@@ -676,20 +714,56 @@ impl CompiledTrace {
         }
     }
 
-    fn is_guarded_branch_exit(&self, cpu: &CpuCore, ops_done: u32) -> bool {
-        let Some(index) = ops_done.checked_sub(1).map(|index| index as usize) else {
-            return false;
-        };
-        self.ops.get(index).is_some_and(|op| {
-            matches!(
-                op.op,
-                JitTraceOp::Branch {
-                    expected_taken: Some(_),
-                    ..
-                }
-            ) && cpu.ppc == op.pc
-        })
+    fn is_guarded_branch_exit(&self, cpu: &CpuCore) -> bool {
+        // `ops_done` cannot identify an op once a preceding CondSkip makes the
+        // retired count data-dependent. Native and portable guarded branches
+        // both leave the branch address in ppc and one of its two successors
+        // in pc, so identify the exit from that architectural state instead.
+        let mut guarded_ops = self.guarded_ops;
+        while guarded_ops != 0 {
+            let index = guarded_ops.trailing_zeros() as usize;
+            guarded_ops &= guarded_ops - 1;
+            let op = &self.ops[index];
+            if cpu.ppc != op.pc {
+                continue;
+            }
+            let JitTraceOp::Branch {
+                displacement,
+                length,
+                expected_taken: Some(expected_taken),
+                ..
+            } = op.op
+            else {
+                unreachable!("guarded_ops contains only guarded branches")
+            };
+            let target = (op.pc.wrapping_add(2) as i32).wrapping_add(displacement) as u32;
+            let fallthrough = op.pc.wrapping_add(length as u32);
+            if target == fallthrough {
+                return cpu.pc == target;
+            }
+            return (cpu.pc == target && !expected_taken)
+                || (cpu.pc == fallthrough && expected_taken);
+        }
+        false
     }
+}
+
+fn guarded_op_mask(ops: &[TraceBuildOp]) -> u128 {
+    debug_assert!(TRACE_MAX_OPS <= u128::BITS as usize);
+    debug_assert!(ops.len() <= TRACE_MAX_OPS);
+    ops.iter().enumerate().fold(0, |mask, (index, op)| {
+        if matches!(
+            op.op,
+            JitTraceOp::Branch {
+                expected_taken: Some(_),
+                ..
+            }
+        ) {
+            mask | (1u128 << index)
+        } else {
+            mask
+        }
+    })
 }
 
 struct TraceRecording {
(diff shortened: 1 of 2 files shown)


benletchford:
Triage decision: keeping this as a draft, not merging or closing it. It is currently conflicted and stacked on #134, so its diff includes declined/unresolved work and has no CI on the prospective integration. Once #134 is restacked or otherwise resolved, rebase this to the intended LEA/PEA prerequisite plus if-conversion commits, refresh the exact-tree measurement, and mark it ready for review.

benletchford:
Thanks for the restack. #134 has now merged, but I am holding this PR for another integration pass before merge.

The remaining LEA/PEA + if-conversion work conflicts substantially with #141's trace machinery. More importantly, `CondSkip` makes retired operation counts data-dependent while the call driver still infers complete/partial execution from the static `ops_len`; as the PR body notes, that can suppress #141's clean-link chaining when the changes are combined. Please rebase onto current `master`, integrate deliberately with the chosen #141 direction, and rerun exact-tree CI and performance checks.

Please also retitle the PR so the squash/release note covers both pieces of its remaining diff, for example: `perf(jit): expand traces with effective-address ops and short-branch if-conversion`.

benletchford:
I rebased and consolidated this branch onto current `master` after #134/#144 and verified the integrated tree locally: formatting, clippy, default-feature library tests (167), and all-feature library tests (267) all pass. The title now covers both the LEA/PEA admission and if-conversion work. I will wait for the refreshed GitHub checks before merging.

PR #141, opened 2026-08-21
perf(jit): trace through bytecode dispatchers

rlanday:
## Summary

This PR adds five general trace-JIT mechanisms needed to follow a bytecode
dispatcher efficiently:

1. an exit-seeded continuation may finish by linking to an already compiled
   trace head;
2. `MOVE` may read from the PC-relative addressing modes used by constant and
   jump tables;
3. `JMP (d8,PC,Xn)` may continue through one recorded target, with an exact
   side exit for every other target and constant-time guard classification for
   guard-free traces;
4. discontiguous traces validate one memory slice per contiguous code segment
   instead of one virtual-bus read per instruction;
5. the direct-mapped trace cache grows from 4,096 to 16,384 slots, reducing
   collisions without adding a probe loop.

The first three commits let the JIT compile through SimCity 2000's dispatcher
and case handlers instead of retaining only the short prefix before the jump
table. On the fixed 400M-instruction SC2K census, those commits raise JIT
coverage from **46.94% to 54.98%** (+8.04 percentage points). Coverage alone
did not produce a wall-time win because it also created more, shorter native
calls. The final two commits and the guard-index optimization reduce the entry,
validation, and cache-churn costs exposed by that new trace shape.

In a direct production comparison of current `master` against the exact clean
five-commit tree, the candidate won all six accepted interleaved pairs. Paired
median: **-9.82% wall time and -9.24% user CPU**. The acceptance gate, exact
binaries, full table, rejected attempts, and outlier treatment are below.

The branch is exactly five commits on `master` (`01a4f7b`):

```text
b3e52e2 perf(jit): expand the direct-mapped trace cache
d125c4c perf(jit): validate discontiguous trace code by segment
5fd2899 perf(jit): guarded computed JMP — trace through jump-table dispatch
d1b8725 perf(jit): admit PC-relative MOVE sources ((d16,PC) and (d8,PC,Xn))
5d1157f perf(jit): link exits — a region reaching a compiled head is a valid trace
```

Every commit carries:

```text
Co-Authored-By: GPT-5.6 <noreply@openai.com>
```

The existing Claude co-authorship trailers on the first three commits are
also preserved.

The branch does **not** contain the previously declined Cranelift-verifier
change, guarded `RTS`, or the unrelated immediate-ALU and PC-relative `LEA`
admissions from the earlier experimental stack.

### What SC2K's dispatcher is doing

SC2K contains a small interpreter inside the 68k application. Its bytecode is
game data, not 68k machine code. Each iteration roughly does this:

```text
read the next SC2K bytecode
use it to select an entry in a nearby table
jump to the corresponding 68k handler
run the handler
return to the dispatcher for the next bytecode
```

The important portion of the hot dispatcher is equivalent to:

```asm
MOVE.W  (6,PC,D0.W),D0   ; load a signed handler offset from a table
JMP     (PC,D0.W)        ; jump to the selected handler
```

For readers unfamiliar with 68k notation:

- `PC` is the program counter: the address of the instruction being executed.
- `D0` is a data register. `D0.W` means use its low 16 bits as a signed index.
- `(6,PC,D0.W)` means memory at a small fixed displacement from this
  instruction plus the runtime index in `D0`.
- `.W` means a 16-bit word.
- The table stores relative offsets. The subsequent `JMP` combines the
  selected offset with a PC-relative base to obtain a handler address.

Before this PR, recording stopped before the indexed table read. Even if a
case handler later became traceable, recording it back to the dispatcher was
rejected because that continuation reached an existing trace instead of
closing a backward branch of its own.

The commits solve those boundaries in execution order, then remove the costs
that the newly discontiguous dispatcher traces expose.

## Commit 1: link exits

Commit: `5d1157f perf(jit): link exits — a region reaching a compiled head is
a valid trace`

### Previous behavior

A guard exit may land in a handler that eventually rejoins its parent loop.
Recording starts at that exit target and follows the handler, but when it
reaches the already compiled parent head, the handler has no terminal branch
of its own. The recorder therefore rejected useful work with
`NoTraceTerminal`, or continued into the parent and built a duplicate
supertrace.

### New behavior

Recordings remember whether they were seeded by a guard exit. Only those
exit-seeded recordings may finish when their next instruction is an existing
compiled head. The continuation retains its own operations up to that point;
its non-branch tail materializes the sequential successor PC, and the existing
bounded chaining path immediately enters the compiled neighbor.

`pending_exit_seed` carries that provenance across the decoded loop's
immediate post-exit probe. Without it, the same exit would count twice and the
second probe would begin an ordinary recording that lacks permission to
link-finish.

The profiler records a link exit separately from a guarded-branch exit, so
every chained invocation still has an attributable source.

### Safety boundary and focused verification

- Ordinary loop recordings cannot link-finish. Allowing them to stop at an
  interior compiled head can replace a useful whole loop with a short stub.
- A watched link target is counted but is neither chained nor recorded across
  the watch boundary.
- Chaining remains bounded by the existing chain and instruction budgets.
- Each linked trace validates its own guest instruction bytes before entry.
- A child validation miss preserves the parent's already retired count while
  returning the changed opcode to full dispatch.
- One robustness nit is noted for review rather than silently carried: the
  pending exit-seed marker is consumed by the immediate post-exit probe at
  the seeded address. If that probe does not occur (the batch's instruction
  budget ran out exactly at the boundary, or a recording was already
  active), the marker survives until a later ordinary probe at that same
  address, which then starts its recording with link-finish permission it
  would not otherwise have. The resulting trace is still correct — at worst
  a shorter stub that chains into the compiled neighbor — so the effect is
  bounded. The tightening (clear the marker on any probe at a different
  address) is left as a follow-up so the published tree stays byte-identical
  to the measured one.

`an_exit_seeded_recording_link_finishes_at_a_compiled_head` proves that an
exit-seeded continuation stops at the compiled neighbor, retains only its own
operations, and uses the normal chain. Related tests cover watched targets,
validation misses, productive guard exits, and memory bails that must not
create exit candidacy.

## Commit 2: PC-relative `MOVE` sources

Commit: `d1b8725 perf(jit): admit PC-relative MOVE sources ((d16,PC) and
(d8,PC,Xn))`

### Previous behavior

The recorder rejected `MOVE` when its source used either form:

```asm
MOVE.W  (d16,PC),Dn       ; fixed data near the instruction
MOVE.W  (d8,PC,Xn),Dn     ; table entry selected by a register
```

The second form is the jump-table lookup at the center of SC2K's dispatcher.

### New behavior

The decoder resolves the architectural PC portion once while recording:

- `(d16,PC)` becomes a `PcDisp` source with its resolved constant address;
- `(d8,PC,Xn)` becomes `PcIndex`: a constant PC base plus the live index
  register, word/long selection, legal scale, and signed displacement.

Only the address calculation is folded. The table value is read from guest
memory every time the trace runs, so guest changes to data tables remain
visible. PC-relative modes remain invalid as `MOVE` destinations.

`PcDisp` intentionally remains distinct from `AbsLong`. They may resolve to
the same numerical address, but the 68000 charges different cycles for the
two addressing modes. Treating PC-relative displacement as absolute-long
would over-retire `MOVE.W` and `MOVE.L` by four cycles. The tests pin the
interpreter's exact cycle result instead of assuming those modes are
interchangeable.

For a pre-68020 CPU, brief-index scaling stays at architectural 68000
behavior. Later-CPU scale bits and long-index selection are decoded only when
the selected CPU model permits them.

### Safety boundary and focused verification

- Runtime memory accesses use the existing checked fast-memory window.
- Bounds and pre-68020 alignment checks occur before register or flag state
  commits.
- Table values are runtime data, not constants embedded in native code.
- Opcode and extension words remain covered by ordinary trace-code
  validation.

`portable_move_from_pc_index_reads_the_live_table` changes the table after
recording and proves the trace sees the new value.
`pc_relative_move_sources_match_the_68000_interpreter_and_cycles` compares
registers, flags, memory, and exact cycles. The native comparison keeps both
JIT executors in lockstep.

## Commit 3: guarded computed `JMP`

Commit: `5fd2899 perf(jit): guarded computed JMP — trace through jump-table
dispatch`

### Previous behavior

Unlike an ordinary branch, `JMP (d8,PC,Xn)` has no single static target. The
register-selected offset may name a different handler on every dispatch, so
the recorder ended at this instruction.

### New behavior

The recorder decodes the brief PC-indexed extension and records the target
observed on that execution. Native and portable execution both:

1. read the live index register;
2. sign-extend it as a word when required and apply the legal scale;
3. add it to the recorded PC-relative base;
4. commit the architectural jump (`PC = actual target` and change-of-flow);
5. compare the actual target with the recorded target.

If the target matches, execution continues into the recorded handler. If it
differs, the trace exits after retiring the `JMP`, with `PC` pointing at the
actual selected handler. Exit seeding can then compile that handler as a
continuation; commit 1 lets it finish when it rejoins the dispatcher.

The native exit also writes the previous PC and real instruction word
(`IR = 0x4EFB`), matching the interpreter and portable executor.

### Guard exits and the full-retirement case

A guard is a runtime check that proves the path recorded earlier still matches
the path selected now. A guard exit is the safe fallback when it does not: the
trace stops before executing the wrong recorded handler and returns to normal
dispatch at the real target.

The guarded `JMP` can be the last operation in a trace. A mismatch there
legitimately retires the full numerical operation count, so
`retired == trace length` cannot distinguish it from clean completion.
Classification instead checks the committed previous PC, current PC,
instruction shape, and recorded target. A last-operation mismatch therefore
still seeds the actual handler and counts as a guard exit.

### Constant-time classification for guard-free traces

The first guarded-JMP implementation searched the complete operation vector
after every native trace return to find an operation capable of explaining a
guard exit. SC2K makes about 12.47 million native calls in the core census, so
that O(trace length) search taxed even traces with no guard.

Compilation now records a 128-bit mask over the at-most-128 trace operations.
A bit is set only for `Branch` operations with a recorded direction and
`PcIndexJmp` operations with a recorded target. A guard-free trace performs
one zero check. A guarded trace scans only set bits and reuses the exact
existing state classifier, including the case where a branch target equals
its fallthrough address. This adds no allocation.

In a six-pair isolation against the otherwise identical candidate, the mask's
median effect was **-2.90% wall and -3.19% user**, with 5/6 wall wins.

### Safety boundary and focused verification

- The real jump commits before the target comparison; a miss exits at the
  selected guest address rather than speculatively running the recorded case.
- The expected target predicts control flow; it does not replace the live
  index.
- Both match and mismatch charge the exact 14-cycle jump cost and one retired
  guest instruction.
- Unsupported full-format index extensions remain outside the decoder.
- The mask contains only operation kinds that the unchanged state classifier
  can recognize.

`guarded_pc_index_jmp_matches_portable_and_seeds_a_last_op_exit` checks native
and portable results in both target directions, including cycles, retirement,
PC/previous-PC/IR, mask contents, and last-operation exit seeding.

`a_bytecode_dispatch_loop_compiles_through_the_guarded_computed_jump` runs a
miniature alternating bytecode dispatcher. It proves that the dispatcher
compiles through the computed jump, the alternate case becomes an exact
three-operation continuation rather than a duplicate supertrace, both case
bodies complete, and the continuation links back to the compiled dispatcher.

## Commit 4: validate discontiguous trace code by segment

Commit: `d125c4c perf(jit): validate discontiguous trace code by segment`

### Previous behavior

Before running compiled guest code, the JIT verifies that the 68k instruction
bytes in memory still match the bytes used for compilation. This is necessary
because a 68k program may rewrite its own code; running a stale translation
would execute the wrong guest instructions.

A conventional loop occupies one consecutive range of guest addresses and
can be validated with one slice comparison. A dispatcher trace contains the
dispatcher at one address and a selected handler at another. Previously every
such discontiguous trace fell back to reading and checking each recorded
instruction through the virtual bus on every entry. That was correct but made
the new dispatcher traces expensive to enter.

### New behavior

The recorder groups adjacent instruction bytes into `TraceCodeSegment`
records. Each says where a consecutive code range starts in guest memory,
where its expected bytes begin in the trace snapshot, and how many bytes it
contains.

For example, a trace with a dispatcher at `0x008C7600..0x008C761F` and a
handler at `0x008D2700..0x008D2711` has two segments. Entry performs two slice
comparisons instead of one virtual-bus read per instruction. This changes only
how the existing validity check is batched, not which instructions run or
their order.

### Safety boundary and focused verification

- Every segment is checked before the first trace operation executes. A
  changed later handler cannot allow an unchanged dispatcher prefix to run.
- The fast path requires every complete segment to lie inside the active
  direct-memory window and match its snapshot.
- Any failure falls through to the existing per-instruction validator, which
  identifies the first changed instruction, invalidates the translation, and
  returns without committing guest state.
- Address gaps between blocks are not assumed contiguous and are not treated
  as trace code.

`fast_validation_checks_every_discontiguous_code_segment_before_execution`
constructs two separated segments, changes only the second, and proves that
neither segment executes before invalidation. The dispatcher composition test
also checks that the dispatcher and handler are represented as two segments.

## Commit 5: expand the direct-mapped trace cache

Commit: `b3e52e2 perf(jit): expand the direct-mapped trace cache`

### Previous behavior

The cache used 4,096 direct-mapped slots. “Direct-mapped” means each guest
program counter has one possible slot:

```text
(PC >> 1) & (slot_count - 1)
```

The lookup is intentionally one shift, one mask, and one slot read. With
4,096 slots, however, heads separated by `0x2000` bytes map to the same slot.
Installing one evicts the other even if both are hot. SC2K's dispatcher adds
useful heads and continuations across separated code regions, so collisions
can repeatedly pay the counting, recording, validation, and compilation cost.

### New behavior

The cache grows to 16,384 slots. The lookup formula and direct-mapped policy
are unchanged, but the address distance that necessarily aliases grows from
`0x2000` to `0x8000` bytes.

This changes capacity and collision frequency, not replacement semantics.
Each slot still stores the full guest PC and CPU type, so a real alias is a
safe miss or eviction rather than execution of another address's trace.

### Trade-off and focused verification

- Slot and candidacy tables use four times the storage.
- Lookup still has no probe loop and adds no hot-path branch.
- Invalidation, code validation, CPU-type discrimination, and bounded tracing
  are unchanged.
- Collision tests derive their second address from `TRACE_CACHE_SIZE`, assert
  that it shares a slot, and then prove the old eviction and rejection-policy
  behavior under the new 16K geometry.

In absolute terms: each slot holds its `CompiledTrace` inline (about 180
bytes on this development target; exact size is target/configuration-
dependent because it contains several `Vec` headers), so the slot table
grows from roughly 0.75 MB to 3 MB per `TraceJit`, which is thread-local
and exists in portable/WASM builds too. The four two-way side tables
indexed like the cache (`earned_call_permission`, `structurally_rejected`,
`compiled_before`, `no_terminal_strikes`) grow from about 160 KB to 640 KB
combined.

## How the five commits compose

```text
dispatcher head
  -> PC-relative MOVE reads the live jump-table entry
  -> guarded computed JMP follows the recorded case
  -> another case exits at its real handler address
  -> exit seeding compiles that handler as a continuation
  -> link completion rejoins the existing dispatcher trace
  -> segmented validation checks the separated blocks cheaply
  -> the larger cache keeps the additional heads resident longer
```

Nothing recognizes SC2K bytecodes or hard-codes its addresses. The JIT sees
ordinary 68k addressing and control flow, records one observed path, and uses
generic guard and continuation machinery for other paths.

### Deterministic dispatcher coverage

The fixed census executes the same scripted SC2K path for 400M guest
instructions and reaches guest tick 118533:

| arm | JIT-retired | coverage | native calls | ops/native call |
|---|---:|---:|---:|---:|
| current `master` | 187,775,765 | 46.94% | 9,820,231 | 19.12 |
| first three commits | 219,931,324 | 54.98% | 12,467,858 | 17.64 |

The dispatcher mechanisms add **32,155,559 JIT-retired instructions** and
**8.04 coverage points**. The native-call count rises while average work per
call falls, which is why coverage is not presented as a speed claim. The last
two commits and guard mask admit no additional opcode; they target the cost of
entering, validating, classifying, and retaining the traces already exposed.

### Direct production timing: `master` versus the complete candidate

The release binaries use identical Systemless source and differ only in the
m68k tree. Trace profiling is disabled. A measurement-only environment switch
suppresses the 800 periodic PNG encodes but retains screenshot 9999, its guest
tick check, the fixed input script, and the 400M-instruction budget.

- `master` revision: `01a4f7b`; binary SHA-256
  `a1b671d8085281eb92567dbd5565dba25ffb5736b853bf083c1464133a616045`
- measured candidate tree: `40d337dc0caf489b59a8fc3057c7be8520b6a78d`; presentation
  head `b3e52e2`; binary SHA-256
  `fa079d71d2849745bc6f13e0defc35f5ac571453c54dd3472d31d6263d3f390c`

Every accepted arm had to reach tick 118533, emit exactly one final
screenshot, finish within 45 seconds, and have `user / real >= 95%`. If either
arm failed, the complete pair was discarded. Order alternated, with three
accepted pairs in each order.

| pair | master wall | candidate wall | wall delta | master user | candidate user | user delta |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 16.07s | 15.75s | -1.99% | 15.63s | 15.30s | -2.11% |
| 2 | 17.77s | 16.60s | -6.58% | 17.35s | 16.31s | -5.99% |
| 3 | 14.51s | 13.94s | -3.93% | 14.13s | 13.52s | -4.32% |
| 4 | 19.67s | 13.96s | -29.03% | 18.95s | 13.59s | -28.28% |
| 5 | 20.48s | 17.69s | -13.62% | 19.62s | 17.17s | -12.49% |
| 6 | 18.91s | 16.44s | -13.06% | 18.14s | 15.86s | -12.57% |

The candidate wins 6/6 wall pairs and 6/6 user pairs. Paired median:
**-9.82% wall, -9.24% user**. Paired mean: -11.37% wall, -10.96% user.

Pair 4 is an obvious host-frequency outlier even though both arms passed the
outcome-independent gate. It is retained rather than silently discarded, but
neither it nor the mean is the headline. The median is robust to that one
outlier. Several other attempts were rejected in full for falling below the
95% utilization gate; no favorable arm from a rejected pair was reused.

As narrower mechanism checks:

- the complete candidate versus the three-commit core has median **-2.31%
  wall and -2.37% user**, with 4/6 wall wins and 5/6 user wins;
- the precomputed guard mask versus the otherwise identical candidate has
  median **-2.90% wall and -3.19% user**, with 5/6 wall wins.

The direct five-commit comparison against `master`, rather than either
narrower and noisier ablation, is the PR-level result.

### EV Override regression check

Three of this PR's commits (the guard mask, segmented validation, and the
larger cache) sit on every workload's hot path, so the PR was also checked
on a second application rather than only the dispatcher workload above.

Instrument: the headless census harness, fixed input script, 400 million
guest instructions, m68k trace profiling enabled. Metric: `jit_retired /
400M` — the fraction of retired guest instructions that retired inside
compiled traces. Coverage is a mechanism check, not a speed claim. EV
Override's guest timers advance on wall time while the budget advances on
retired instructions, so arm runs are interleaved pairwise and each pair
starts from a fresh save game; running whole arms as blocks had overstated
deltas by more than half a point in earlier work on this harness.

Arms (release binaries, identical Systemless source, only the m68k revision
differs):

- base: master `01a4f7b`; binary SHA-256
  `eaae440e858784d1cb7fddc0f3a8318f8949134ea9ac5388f99e7f268d8274ed`
- this PR: `b3e52e2` (tree `40d337dc…`, identical to the measured tree);
  binary SHA-256
  `8cb37c4c6d042a9b300b60b7e9a936867a5c1c1cb653a71719fbadf74ae301e4`

| Pair | base | this PR | delta |
|---|---:|---:|---:|
| 1 | 71.86% | 71.66% | −0.20 pts |
| 2 | 71.21% | 71.35% | +0.14 pts |
| 3 | 71.88% | 71.35% | −0.53 pts |

The deltas sit inside this harness's observed run-to-run jitter for EV
Override (about ±0.5 points): this PR neither taxes nor helps that
workload's coverage. For contrast, the same pair of trees moves the
deterministic SC2K census by +8 points, which is why the dispatcher
mechanisms are presented as SC2K-targeted with an explicit EV-neutrality
check rather than as general gains.

On SC2K itself, the complete five-commit tree — which the coverage table
above shows only as its three-commit core — measures JIT-retired
`220,363,450` (55.09%) with `12,510,013` native calls, versus master's
`187,775,765` (46.94%) and `9,820,231`. The final two commits add about
+0.11 coverage points over the three-commit core while keeping the entry
count essentially unchanged.

Six interleaved production wall-time pairs (periodic screenshot encoding
suppressed, fixed 400M guest budget) were inconclusive under this host's
frequency drift: paired wall deltas ranged from −9.4% to +21.8% with no
consistent sign. The harness's own host-work counters, which cancel that
drift far better than wall time, are neutral: host instructions per guest
instruction ranged −1.5%..+1.5% around a median of −0.2%. On this
workload the five commits neither tax nor help host work — consistent
with the unchanged coverage above.

## Repository verification

The measured prototype and the presentation branch have the identical final
Git tree `40d337dc0caf489b59a8fc3057c7be8520b6a78d`.

```sh
cargo test --lib --no-default-features
cargo test --lib --features jit
cargo test --lib --features jit,trace-profile
cargo clippy --lib --features jit,trace-profile -- -D warnings
cargo fmt --check
git diff --check master...HEAD
```

Results: **162**, **227**, and **263** tests respectively; clippy, formatting,
and diff checks are clean.

## Why the remaining experiments are not in this PR

The previously declined release-mode Cranelift-verifier bypass is removed and
will not be republished here. Guarded `RTS`, immediate-memory ALU admission,
and PC-relative `LEA` admission are also absent. They are independent semantic
changes, are not required for the clean wall-time win, and would expand this
review beyond the dispatcher foundation and the two costs directly exposed by
it.

Segmented validation, the larger cache, and the guard index are included
because isolated measurements showed that entry/validation overhead was the
reason the additional coverage failed to become a wall-time win. The final
five-commit tree then cleared the predeclared direct-win threshold. The other
opcode admissions did not clear that same scope-and-evidence test.

## Biggest remaining JIT levers

This section was revised on 2026-08-23 after profiling where the census
process actually spends CPU, because the profile reverses the ranking an
earlier draft gave. Instrument: macOS `sample` (1 ms interval, two
8-second attachments at different phases) on this PR's exact census
binary during the deterministic 400M-instruction SC2K run; leaf
(self-time) attribution, cross-checked against an equivalent tree on a
quiet host. Consistently across the samples:

- interpreter + batch runner (inlined execution): 53–55% of process CPU
- memory bus, mostly serving the interpreter: ~6%
- the entire Rust trace-entry/exit boundary — cache probe, byte
  validation, exit classification, chaining bookkeeping: 3.2–3.7%
- JIT-generated code itself: 1–2%
- the remainder is census-harness overhead (periodic screenshot PNG
  encoding and trace-profile accounting, both absent in production
  builds) plus trap/OS services in the host.

Two consequences, in priority order:

1. **Further admissions/coverage are the top SC2K lever after this PR.**
   The ~45% of guest instructions that still retire in the interpreter
   cost roughly forty times as much per instruction as the 55% that
   retire natively. The census failure tables concentrate the blockers
   into a compact admission cluster — indexed-EA forms
   (`MOVEA.L (0,An,Dn.L)` pointer-table loads,
   `BTST`/`ANDI`/`CMPI` `#imm,(An,Xn)`, indexed `MOVE.B` stores) plus
   `JSR`/`RTS` boundaries around small helpers — rather than a long tail
   of unrelated shapes.
2. **Cross-trace mechanisms compete for the small boundary pool.** A
   multi-target dispatcher inline cache and validated native
   trace-to-trace links both remain coherent designs, but everything
   they can remove lives inside that ≈4% bucket, so neither is ranked
   next on this workload.

For whoever attempts the inline cache later: target selection cannot
simply choose the most frequent handler. A removed experiment did that
and eliminated roughly 898,000 dispatcher guard exits, but the selected
handler immediately encountered another data-dependent branch. Native
calls became shorter and total JIT retirement moved by only -0.013
percentage points. The policy must optimize useful native work retained
per entry, not raw target frequency.

EV Override's short forward-branch problem is addressed separately by
the if-conversion work; combining it with this dispatcher PR would mix
independent workloads, control-flow mechanisms, and performance claims.

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -31,7 +31,11 @@ use std::fmt;
 use std::mem::{offset_of, size_of, transmute};
 use std::sync::atomic::{AtomicBool, Ordering};
 
-const TRACE_CACHE_SIZE: usize = 4096;
+// Guest code in large applications commonly places unrelated hot loops one
+// 8 KiB region apart. A 4K-entry direct-mapped cache aliases those heads
+// because its byte-address period is only 0x2000; four times as many entries
+// keeps the lookup branchless while moving the collision period to 0x8000.
+const TRACE_CACHE_SIZE: usize = 16_384;
 pub(crate) const TRACE_MAX_OPS: usize = 128;
 pub(crate) const TRACE_MIN_OPS: usize = 3;
 
@@ -86,6 +90,28 @@ const NO_TERMINAL_STRIKE_LIMIT: u8 = 3;
 
 /// Sentinel for `CpuCore::trace_record_skip` / `trace_probe_skip`: no PC.
 pub(crate) const TRACE_PC_NONE: u32 = u32::MAX;
+
+/// Trace-return completion status in the otherwise-unused sign bit of the
+/// cycle word. Trace execution is bounded by `CpuCore::cycles_remaining`
+/// (`i32`), so cycles never need this bit. The upper 32 retirement bits stay
+/// fully available for data-dependent retirement such as `CondSkip`.
+const TRACE_RETURN_COMPLETE: u64 = 1 << 31;
+const TRACE_RETURN_CYCLES_MASK: u32 = i32::MAX as u32;
+
+#[inline]
+fn trace_return_complete(packed: u64) -> bool {
+    packed & TRACE_RETURN_COMPLETE != 0
+}
+
+#[inline]
+fn trace_return_cycles(packed: u64) -> u32 {
+    packed as u32 & TRACE_RETURN_CYCLES_MASK
+}
+
+#[inline]
+fn trace_return_retired(packed: u64) -> u32 {
+    (packed >> 32) as u32
+}
 
 #[cfg(all(feature = "jit", not(target_family = "wasm")))]
 /// Original one-pass compiled trace entry point.
@@ -147,6 +173,20 @@ pub(crate) enum JitEa {
         scale: u8,
         displacement: i8,
     },
+    /// Brief (d8,PC,Xn): the PC-relative base and displacement collapse
+    /// to a constant when the trace is recorded (the pc is the trace's
+    /// own code), leaving a constant-base indexed read. Read-only by
+    /// construction: PC-relative modes are not legal destinations.
+    PcIndex {
+        base: u32,
+        index: JitDirectReg,
+        index_long: bool,
+        scale: u8,
+    },
+    /// (d16,PC): the extension-word PC and signed displacement collapse
+    /// to a constant address at record time. Kept distinct from absolute
+    /// addressing because the 68000 effective-address cycle charge differs.
+    PcDisp(u32),
     /// Absolute-short memory address, sign-extended when the trace is
     /// recorded so execution does not read the instruction stream.
     AbsWord(u32),
@@ -163,6 +203,8 @@ impl JitEa {
                 | Self::PreDec(_)
                 | Self::Disp(_, _)
                 | Self::Index { .. }
+                | Self::PcIndex { .. }
+                | Self::PcDisp(_)
                 | Self::AbsWord(_)
                 | Self::AbsLong(_)
         )
@@ -402,6 +444,18 @@ pub(crate) enum JitTraceOp {
         /// means this branch ends the trace; `Some` emits a guarded side
         /// exit and continues along the recorded path on a match.
         expected_taken: Option<bool>,
+    },
+    /// `JMP (d8,PC,Xn)` -- an N-way jump-table dispatch. The base (pc +
+    /// 2 + d8) folds to a constant at record time; the recorded taken
+    /// target guards the trace: a mismatch commits the jump (pc = the
+    /// computed target) and side-exits, where exit seeding compiles the
+    /// other dispatch cases as continuations that link back.
+    PcIndexJmp {
+        base: u32,
+        index: JitDirectReg,
+        index_long: bool,
+        scale: u8,
+        expected_target: Option<u32>,
     },
     Dbcc {
         condition: u8,
@@ -650,15 +704,27 @@ impl TraceBuildOp {
     }
 }
 
+/// One contiguous guest-code range inside the execution-ordered `code`
+/// snapshot. A trace can jump between several ranges (for example a computed
+/// dispatch or an inlined call); validating each range with a slice compare
+/// avoids re-entering the bus once per recorded instruction.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct TraceCodeSegment {
+    start: u32,
+    code_offset: u32,
+    len: u32,
+}
+
 struct CompiledTrace {
     pc: u32,
     cpu_type: CpuType,
     ops: Vec<TraceBuildOp>,
-    /// The exact instruction bytes the trace was compiled from (ops are
-    /// in execution order. Contiguous traces can validate this with one
-    /// compare; recorded multi-block paths validate each instruction.
+    /// The exact instruction bytes the trace was compiled from, in execution
+    /// order. `code_segments` maps its contiguous slices back to guest
+    /// addresses so multi-block traces can use the same fast validation as a
+    /// linear trace.
     code: Vec<u8>,
-    contiguous_code: bool,
+    code_segments: Vec<TraceCodeSegment>,
     max_cycles: i32,
     /// The final branch's taken-target is the trace head, so the trace is
     /// a whole loop iteration and can be re-run (budget permitting)
@@ -689,10 +755,10 @@ struct CompiledTrace {
     /// between them is deliberately unguarded.
     callee_start: u32,
     callee_end: u32,
-    /// Bit `n` is set when operation `n` is a recorded branch guard. Trace
-    /// entry can reject guard-free partial returns after one zero check, and
-    /// guarded traces inspect only the guarded operations instead of walking
-    /// up to 128 recorded operations.
+    /// Bit `n` is set when operation `n` has a recorded control-flow guard.
+    /// Trace entry checks this before inspecting `ops`, so the common
+    /// guard-free return is constant-time; guarded traces visit only their
+    /// guarded operations instead of scanning the complete recorded path.
     guarded_ops: u128,
     /// A recorded interior branch is a path prediction eligible for adaptive
     /// rerecording. Cleared after the one allowed rerecord so completed traces
@@ -708,17 +774,22 @@ struct CompiledTrace {
 impl CompiledTrace {
     #[cfg(all(feature = "jit", not(target_family = "wasm"), test))]
     unsafe fn call_native(&self, cpu: *mut CpuCore, max_iters: u32) -> u64 {
-        match self.func {
+        let packed = match self.func {
             NativeTraceFn::Once(func) => unsafe { func(cpu) },
             NativeTraceFn::Loop(func) => unsafe { func(cpu, max_iters) },
-        }
+        };
+        // Direct executor tests assert the architectural cycles/retirement
+        // payload; completion is internal call-driver metadata.
+        packed & !TRACE_RETURN_COMPLETE
     }
 
     fn is_guarded_branch_exit(&self, cpu: &CpuCore) -> bool {
-        // `ops_done` cannot identify an op once a preceding CondSkip makes the
-        // retired count data-dependent. Native and portable guarded branches
-        // both leave the branch address in ppc and one of its two successors
-        // in pc, so identify the exit from that architectural state instead.
+        // A side exit can retire the trace's full numeric op count when the
+        // guarded control-flow op is last, and future data-dependent ops can
+        // make a retired count differ from a static trace index. Both native
+        // and portable guards leave the exiting instruction in PPC and the
+        // architecturally committed successor in PC, so classify from that
+        // state instead.
         let mut guarded_ops = self.guarded_ops;
         while guarded_ops != 0 {
             let index = guarded_ops.trailing_zeros() as usize;
@@ -727,22 +798,28 @@ impl CompiledTrace {
             if cpu.ppc != op.pc {
                 continue;
             }
-            let JitTraceOp::Branch {
-                displacement,
-                length,
-                expected_taken: Some(expected_taken),
-                ..
-            } = op.op
-            else {
-                unreachable!("guarded_ops contains only guarded branches")
+            return match op.op {
+                JitTraceOp::Branch {
+                    displacement,
+                    length,
+                    expected_taken: Some(expected_taken),
+                    ..
+                } => {
+                    let target = (op.pc.wrapping_add(2) as i32).wrapping_add(displacement) as u32;
+                    let fallthrough = op.pc.wrapping_add(length as u32);
+                    if target == fallthrough {
+                        cpu.pc == target
+                    } else {
+                        (cpu.pc == target && !expected_taken)
+                            || (cpu.pc == fallthrough && expected_taken)
+                    }
+                }
+                JitTraceOp::PcIndexJmp {
+                    expected_target: Some(expected),
+                    ..
+                } => cpu.pc != expected,
+                _ => unreachable!("guarded_ops contains only guarded control flow"),
             };
-            let target = (op.pc.wrapping_add(2) as i32).wrapping_add(displacement) as u32;
-            let fallthrough = op.pc.wrapping_add(length as u32);
-            if target == fallthrough {
-                return cpu.pc == target;
-            }
-            return (cpu.pc == target && !expected_taken)
-                || (cpu.pc == fallthrough && expected_taken);
         }
         false
     }
@@ -756,6 +833,9 @@ fn guarded_op_mask(ops: &[TraceBuildOp]) -> u128 {
             op.op,
             JitTraceOp::Branch {
                 expected_taken: Some(_),
+                ..
+            } | JitTraceOp::PcIndexJmp {
+                expected_target: Some(_),
                 ..
             }
         ) {
@@ -782,6 +862,13 @@ struct TraceRecording {
     /// recorded from the fall-through, the recorder skips re-recording the
     /// executed block ops until control reaches this pc (the branch target).
     skip_record_until: Option<u32>,
+    /// Whether this recording was seeded from a guard exit rather than a
+    /// backward branch. Only exit-seeded recordings may finish early by
+    /// LINKING at another compiled head: truncating a backward-branch loop
+    /// recording at an interior compiled head replaces the whole loop with
+    /// a stub (measured: whole-application trace coverage fell 14 points
+    /// on one workload when loop recordings were allowed to link-finish).
+    from_exit_seed: bool,
 }
 
 enum TraceSlot {
@@ -816,6 +903,11 @@ pub(crate) struct TraceJit {
     next_func: u32,
     slots: Vec<TraceSlot>,
     recording: Option<TraceRecording>,
+    /// A guarded exit already counted candidacy at this exact target. The
+    /// decoded loop probes again immediately after any trace returns; carry
+    /// the provenance across that probe so it neither double-counts the hit
+    /// nor starts the resulting recording as an ordinary loop candidate.
+    pending_exit_seed: Option<(u32, CpuType)>,
     /// Heads that have earned call-through permission: a recording here
     /// blocked at a recordable call, so future candidacy at the same pc
     /// starts with permission instead of re-earning it through a doomed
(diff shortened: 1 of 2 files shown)


benletchford:
The standalone branch is well tested and I found no independent semantic blocker, but #134 has now merged and a prospective integration produces substantial conflicts in `trace_jit.rs`. #138 also overlaps the same completion/partial-return machinery: its data-dependent `CondSkip` retirement cannot safely be combined with this branch's retirement-based link-exit detection without an explicit resolution.

Please rebase onto current `master` (including #134), resolve that interaction deliberately, and rerun CI on the exact integrated tree. I am leaving this open rather than merging the pre-integration head.

benletchford:
I rebased and integrated this branch onto current `master` after #138. The conflict resolution preserves TrapExit deferral, if-conversion state, exit-seeded linking, guarded branches, and guarded PC-index jumps. It also adds an explicit trace-completion status so data-dependent `CondSkip` retirement cannot be mistaken for a partial call; a new regression proves a shortened CondSkip path still clean-link chains. Local validation passes: formatting, clippy, 170 no-default library tests, and 275 all-feature library tests. I will wait for refreshed GitHub checks before merging.

PR #142, opened 2026-08-24
perf(jit): admit ADDA memory sources and immediate register bit operations

rlanday:
## What this does

One commit, stacked on #141, admitting two instruction forms the trace JIT
refused:

- `ADDA.W/L (An),An` and `ADDA.W/L (d16,An),An` — add a memory word/long
  to an address register (pointer arithmetic through a struct field);
- `BTST/BCHG/BCLR/BSET #imm,Dn` — the static-bit-number register bit
  operations (the dynamic `Dn,Dn` forms were already admitted).

If #141 is declined or restructured, the commit is mechanically
independent of its five commits and can be retargeted; it was measured on
top of #141 because that is where the affected code paths become hot.

## Why exactly these two forms

The selection is measured, not guessed. A per-instruction census of the
fixed 400M-instruction SimCity 2000 workload (every interpreter-retired
instruction counted by program counter and opcode; trace-retired
instructions excluded) shows the interpreter's work is very concentrated:
the top three code regions hold 45% of all interpreted execution. The
single largest region — 19.05%, 32.9M of 172.8M interpreted instructions —
is the routine that feeds the bytecode dispatcher #141 compiles, and its
recordings failed on exactly these forms:

- every recording through the feeder stopped at `ADDA.W (12,A6),A0`
  (opcode `D0EE` at `008C8416`). The decoder admits memory-source ALU to
  data registers and CMPA to address registers, and its comment
  documented the deliberate exclusion of "the mutating ADDA/SUBA forms";
- the continuation regions after the dispatcher's computed jump stopped at
  `BTST #2,D4` and `BTST #0,D0`.

The census instrument itself is measurement-only and is proven
behavior-neutral: with it attached, the run reproduces #141's recorded
census bit-for-bit (`jit_retired=220,363,450`, `native_calls=12,510,013`).

### `AddaMemToReg`

`ADDA` reads memory, sign-extends a word source to 32 bits, adds it to the
full address register, and changes no condition codes. The new trace
operation mirrors the existing `AddrCmpMemToReg` (CMPA) exactly — same
admitted sources (`(An)` and `(d16,An)`), same checked fast-memory window
load, same sign-extension — and then writes the register instead of
setting flags. `SUBA` remains unadmitted: the decoded-memory tier does not
decode it and no measured region blocks on it.

One deliberate policy choice for review: the compile-stage amortization
refusal (`LinearMemoryAlu`), which rejects non-loop regions containing
read-only checked-memory ALU because short linear segments do not pay for
their entry costs, does **not** list the new operation. Every measured
`ADDA` site sits inside long, call-through-bearing recordings — the shape
the refusal was never measured on — and listing it would re-reject the
region this commit exists to compile. If short ADDA-only segments appear
and measure badly, adding the op to the gate is a one-line follow-up.

### `BitImmReg`

The static-bit forms are the immediate siblings of the admitted `BitReg`:
the bit number is reduced modulo 32 while recording, so the mask is a
compile-time constant. `BTST` writes only Z; the modifying forms also
write the register. The exact cycle charge is computed at record time and
pinned against the interpreter's own model: the 68000 pays 4 extra clocks
for the extension-word fetch (`BTST #,Dn` = 10 where `BTST Dn,Dn` = 6),
pre-68020 CPUs add 2 clocks when a modifying op addresses the upper
register half, and 68020+ charge the flat post-020 costs. The
memory-destination static forms remain another decoder's business and are
explicitly rejected here.

### Safety boundary

- The ADDA window load runs through the existing checked fast-memory
  path; an out-of-window source bails atomically before any register or
  flag state commits (test below).
- Neither operation carries a runtime prediction, so neither enters the
  guard-classification mask.
- ADDA mutates a register, so it is deliberately excluded from the
  pure-poll wait-loop classifier's read-only list: a loop containing it is
  self-driven by definition.

## Tests

- Portable ADDA: word source sign-extension, destination write, condition
  codes untouched, exact `pc` advance.
- Native ADDA versus portable on an executed loop: packed result,
  registers, CCR, and `pc` identical; the out-of-window case commits
  nothing and returns a zero packed result.
- Decode of all four immediate bit forms through the real decode router
  from bus memory, with cycle pinning on both the 68040 (6/8/10/8) and
  the 68000 (10/12/14/10 for `BTST #2` / `BCHG #20` / `BCLR #20` /
  `BSET #0`), and rejection of the memory-destination encoding.
- Portable bit ops: Z-only flag behavior with X/N/V/C preserved, the
  modifying results, and the bit-31 mask edge.
- Native bit ops versus portable for all four operations including bit 31.
- Suites: 268 / 232 / 165 library tests across `jit,trace-profile`,
  `jit`, and `--no-default-features`; `clippy --lib -D warnings` and
  `cargo fmt --check` clean.

## Measurements

All measurements use the fixed 400M-instruction SC2K census workload
(deterministic: identical guest execution across runs and, for the base
arm, bit-identical to #141's published numbers).

### Coverage and interpreted-work movement (trace-profile builds)

| arm | JIT-retired | coverage | native calls | ops/call |
|---|---:|---:|---:|---:|
| #141 head `b3e52e2` | 220,363,450 | 55.09% | 12,510,013 | 17.61 |
| + this commit | 242,339,327 | **60.58%** | 16,216,174 | 14.94 |

Interpreted retirement falls 172.8M → 150.9M (−22.0M, −12.7% of all
interpreted execution); the feeder region drops from 32.9M to 11.0M
interpreted instructions, and the previously blocked continuation sites
(`008C845E`, `008C8444`, `008C8430`, `008C844A`, `008C847C`) become
compiled, chaining trace heads. Native calls rise by 3.7M — more, shorter
calls — which is why coverage alone is not the speed claim.

### Paired timing (release builds, no trace profiling)

Both binaries in each comparison were built from the identical Systemless
harness source; a measurement-only environment switch suppresses the 800
periodic PNG screenshot encodes while keeping the final screenshot and
its guest-tick check. Six interleaved pairs per comparison, alternating
order; every accepted run reached guest tick 118533 with exactly one
screenshot. The primary metric is retired **host instructions**
(`/usr/bin/time -l`), which this host's frequency drift cannot touch;
wall time is reported alongside.

**This commit versus #141 head** (base `7800841a…`, arm `3ecec01b…`):

| metric | median | wins |
|---|---:|---:|
| host instructions | **−0.30%** | 6/6 (range −0.17..−0.39%) |
| host cycles | −3.77% | 6/6 |
| wall | −5.19% | 6/6 |

The cycle and wall benefit exceeding the instruction benefit is expected:
interpreter dispatch is branch-miss-heavy, and the work this commit moves
into traces stops paying those misses. A separate six-pair wall-only
batch taken while the host was noisy (base-arm spread ±15%) read +6.8%
median — wall time alone cannot resolve an effect of this size on this
machine, which is exactly why the counter metric is primary.

**Full stack (#141 + this commit) versus `master` `01a4f7b`**
(base `c5f15a7b…`, arm `3ecec01b…`):

| metric | median | wins |
|---|---:|---:|
| host instructions | **−6.43%** | 6/6 (range −6.42..−6.45%) |
| host cycles | −7.47% | 6/6 |
| wall | −12.15% | 6/6 |

The host-instruction spread of ±0.02 points across six pairs reflects the
workload's determinism: the stack does 6.4% less host work per run, full
stop. This is the combined-stack impact number; #141's own body reports
its five commits at −9.82% median wall against master on an earlier day,
and this commit's marginal contribution is the first table.

### EV Override regression check

Same instrument as #141's EV check (interleaved census pairs from a
fresh save, coverage = JIT-retired / 400M), extended to five pairs after
the third exceeded the historically observed jitter:

| pair | #141 head | + this commit | delta |
|---|---:|---:|---:|
| 1 | 71.21% | 71.23% | +0.02 pts |
| 2 | 71.05% | 71.24% | +0.19 pts |
| 3 | 70.92% | 70.05% | −0.87 pts |
| 4 | 70.41% | 71.39% | +0.98 pts |
| 5 | 71.38% | 71.14% | −0.24 pts |

Median +0.02 points; the five deltas straddle zero and the two largest
are opposite-signed. The base arm itself ranged 70.41–71.38% across the
set — this session's per-pair jitter (~±1.0 pt, measured under
continuous background build load) exceeds the ±0.5 historically seen on
this harness — so the check reads neutral: no directional movement on
the second workload, consistent with the admitted forms being generic
but SC2K-hot.

## Commit structure

One commit. The interpreted-retirement census instrument that produced
the targeting evidence is a local diagnostic branch and is not part of
this PR; release builds of the measured tree compile identical code to
this branch.

Original census targeting, implementation, and measurements by Claude
Fable 5, working with @rlanday; the census methodology and the paired
host-counter protocol follow the measurement conventions established in
#134/#138/#141.

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -410,6 +410,18 @@ pub(crate) enum JitTraceOp {
         bit_reg: u8,
         dst: u8,
     },
+    /// BTST/BCHG/BCLR/BSET `#imm,Dn` — the static-bit-number register
+    /// forms (the dynamic `Dn,Dn` forms are `BitReg`). The bit number is
+    /// reduced modulo 32 while recording, so the exact cycle charge
+    /// (including the 68000's extension-word fetch and the pre-68020
+    /// upper-half surcharge for the modifying ops) is a decode-time
+    /// constant.
+    BitImmReg {
+        op: JitBitOp,
+        bit: u8,
+        dst: u8,
+        cycles: i32,
+    },
     Exg {
         opcode: u16,
     },
@@ -559,6 +571,14 @@ pub(crate) enum JitTraceOp {
     /// destination is always the full address register and a word source is
     /// sign-extended before the 32-bit comparison.
     AddrCmpMemToReg {
+        size: Size,
+        src: JitEa,
+        dst: u8,
+    },
+    /// ADDA.W/L through `(An)` or `d16(An)`: the pointer-arithmetic
+    /// sibling of `AddrCmpMemToReg`. A word source is sign-extended, the
+    /// full address register is written, and no condition code changes.
+    AddaMemToReg {
         size: Size,
         src: JitEa,
         dst: u8,
@@ -2582,6 +2602,7 @@ impl TraceJit {
                     | JitTraceOp::ClrMem { .. }
                     | JitTraceOp::MoveImmMem { .. }
                     | JitTraceOp::AddrCmpMemToReg { .. }
+                    | JitTraceOp::AddaMemToReg { .. }
                     | JitTraceOp::AddRegToMem { .. }
                     | JitTraceOp::AnDispUnary { .. }
                     | JitTraceOp::PeaInd { .. }
@@ -2992,6 +3013,10 @@ impl TraceJit {
                             bail_at,
                         )
                     }
+                    JitTraceOp::AddaMemToReg { .. } => {
+                        let env = mem_env.as_ref().expect("AddaMemToReg implies a window env");
+                        emit_adda_mem_to_reg(&mut builder, cpu_ptr, *op, env, &mut bails, bail_at)
+                    }
                     JitTraceOp::AddRegToMem { .. } => {
                         let env = mem_env.as_ref().expect("AddRegToMem implies a window env");
                         emit_add_reg_to_mem(&mut builder, cpu_ptr, *op, env, &mut bails, bail_at)
@@ -3313,6 +3338,7 @@ fn is_pure_poll_loop(ops: &[TraceBuildOp]) -> bool {
             | JitTraceOp::LeaAbs { .. }
             | JitTraceOp::AddSubxReg { .. }
             | JitTraceOp::BitReg { .. }
+            | JitTraceOp::BitImmReg { .. }
             | JitTraceOp::Exg { .. }
             | JitTraceOp::Ext { .. }
             | JitTraceOp::Extb { .. }
@@ -3324,6 +3350,7 @@ fn is_pure_poll_loop(ops: &[TraceBuildOp]) -> bool {
             | JitTraceOp::MoveMem { .. }
             | JitTraceOp::MovemWordPostInc { .. }
             | JitTraceOp::AluMemToReg { .. }
+            | JitTraceOp::AddaMemToReg { .. }
             | JitTraceOp::AnDispUnary { .. }
             | JitTraceOp::AddRegToMem { .. }
             | JitTraceOp::MemAddqSubq { .. }
@@ -3631,6 +3658,7 @@ impl JitTraceOp {
                 ..
             } => 10,
             Self::BitReg { .. } => 8,
+            Self::BitImmReg { cycles, .. } => cycles,
             Self::SccDataReg { .. } => 6,
             Self::Exg { .. } => 6,
             Self::ShiftReg {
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking correctness finding: the newly admitted `AddaMemToReg` forms have incorrect cycle accounting. `JitTraceOp::max_cycles` hard-codes 24 cycles, and both portable and native execution use that value. A differential against `CpuCore::step` gives:

- `ADDA.W (An),A3`: JIT 24 cycles, interpreter 12
- `ADDA.W d16(An),A3`: JIT 24 cycles, interpreter 16

The current tests compare portable and native JIT paths to each other, so both agree on the same wrong value. Please derive the cycle charge from the addressing mode/decoded form and add interpreter-differential coverage for both cases. This branch will also need restacking after #141 is rebased onto current `master` following #134.

benletchford:
I restacked this directly onto current master and consolidated the series into one commit. I also tightened the ADDA memory-source cycle bounds to their exact 68000 costs and added differential coverage against the interpreter for `(An)`/`d16(An)` word and long forms. Clarification from my earlier review: the old 24-cycle bound was an avoidably pessimistic admission/accounting mismatch, not a guest-state correctness bug, because these memory traces remain instruction-budgeted. Local fmt, all-feature clippy, no-default-feature library tests (174), and all-feature library tests (281) pass.

PR #145, opened 2026-08-24
perf(jit): bare RTS/RTD becomes a dynamic-exit trace terminal

rlanday:
## What this does

Two commits, stacked on #143 (and through it #142/#141):

1. The RTD `change_of_flow` fix — the same commit as #144, included so
   this branch's interpreter and JIT agree; it drops out on rebase once
   #144 lands.
2. **`ReturnExit`: a bare RTS or RTD becomes a dynamic-exit trace
   terminal** instead of ending every recording at a blocker.

A bare return's target differs per caller, so no guarded terminal could
represent it, and whole subroutine bodies stayed interpreted — the
per-region census attributes roughly 35–40% of SimCity 2000's
interpreted retirement and ~36% of EV Override's to call/return-bounded
regions. `ReturnExit` records the return as an always-exit terminal:
pop the return address through the checked window, deallocate (RTD's
displacement folds into the same stack-pointer add), and exit the trace
at the popped target.

Three design points:

- **Unguarded by construction.** The exit is the architectural jump
  itself, so there is nothing to predict and nothing to bail on beyond
  the ordinary checked stack read. (An earlier investigation measured
  that specializing one dynamic target of a polymorphic site does not
  pay; this terminal takes the opposite, prediction-free shape.)
- **Completions seed their dynamic target.** A clean trace completion
  normally only chains when the exit lands on an already-compiled head,
  and a fresh per-caller continuation never is one — so without this,
  continuations after a returning subroutine would compile only by luck
  of a backward branch. A `ReturnExit` completion therefore counts
  trace candidacy at the popped address, exactly as guarded branch
  exits do; each hot caller's continuation then compiles and ordinary
  bounded chaining connects body to continuation.
- **Return regions reuse the measured indirect-call break-even length**
  (`TRACE_MIN_INDIRECT_JSR_OPS` = 7): both dynamic-exit terminals pay
  the same per-entry shape (validation, ABI boundary, an exit whose
  target the trace cannot name), and without a floor every two-op
  pop-and-return stub in the guest would claim a direct-mapped cache
  slot at negative value.

Coverage: RTS (`4E75`) everywhere; RTD `#d16` (`4E74`) exactly where
the interpreter accepts it — every model but the original M68000. Both
recorder routes produce the terminal: the ordinary decoder, and
call-through's "RTS with no pending recorded call" case (the head
function's own return), which previously ended the region just before
the return.

## Safety boundary

- The executed semantics are the interpreter's, bit for bit: same pop,
  same stack adjustment, same 16/20 cycle charges, and (with the RTD
  fix) both returns raise `change_of_flow`.
- The stack read goes through the same checked-window path as every
  other trace memory op and runs before any state changes: a miss
  (window, pre-68020 odd SP) bails with nothing committed and the
  return re-executes through full dispatch.
- No existing terminal, guard, or admission changes; the only admission
  change is that the short-region floor now also covers
  return-terminated regions (the `IndirectJsrTooShort` reason, shared
  via an `ends_in_dynamic_exit` predicate).

## Tests

Ten new tests: decode boundaries (RTS; RTD's displacement sign and its
cpu-type gate matching the interpreter's exactly), portable executor
pop/deallocate/exit and bail-without-partial-state, native
commit-only-after-stack-check and RTD displacement, both recorder
routes producing the terminal, the break-even floor (six-op return
region stays decoded, seven-op compiles), and the full
seed-then-promote pipeline (a completion seeds the continuation, a
second completion starts its exit-seeded recording). Suites on this
branch: 279 / 243 / 174 library tests across `jit,trace-profile`,
`jit`, and `--no-default-features`; clippy `-D warnings` and
`cargo fmt --check` clean.

## Measurements

Instrument and conventions as in #143: the deterministic 400M-guest-
instruction headless census (SimCity 2000 city simulation, EV Override
gameplay from a fixed save), coverage = trace-retired share of guest
instructions; SC2K's census is bit-deterministic, EV carries ~0.5 pt
run-to-run jitter so it is measured as interleaved base/arm pairs.
Host-work numbers use `/usr/bin/time -l` instructions retired
(release builds, no trace profiling, screenshots suppressed, both arms
from one harness tree), because this host's frequency drift makes wall
time unable to resolve effects of this size.

### On this branch's own base (#143's head)

Fixed 400M-instruction SC2K census, this branch vs its base:

| arm | JIT-retired | coverage | native calls | ops/call | final guest tick |
|---|---:|---:|---:|---:|---:|
| #143 head | 274,719,478 | 68.68% | 16,520,258 | 16.63 | 135,079 |
| + this branch | 286,354,876 | **71.59%** | 17,653,795 | 16.22 | 135,079 |

Identical final guest tick: the arms executed the same guest work, and
the +2.91 coverage points move 11.6M instructions per run from the
interpreter into traces. Ops per native call dips 16.63 → 16.22
because the newly compiled subroutine bodies are length-capped by
forward branches on this base; on a tree that also carries #138's
if-conversion the same terminal converts far more (below) and ops/call
holds at 19.

EV Override, six interleaved census pairs from a fresh save (two
batches of three):

| pair | #143 head | + this branch | delta |
|---|---:|---:|---:|
| 1 | 74.83% | 74.92% | +0.09 pts |
| 2 | 74.82% | 73.80% | −1.02 pts |
| 3 | 74.50% | 75.16% | +0.66 pts |
| 4 | 73.63% | 74.72% | +1.09 pts |
| 5 | 74.74% | 74.87% | +0.13 pts |
| 6 | 74.51% | 75.59% | +1.08 pts |

Median +0.40, five of six non-negative, the one negative pair within
this harness's observed under-load jitter (±1.0). On this base EV is a
passed regression check rather than a win; the positive EV deltas
appear on the full stack, where if-conversion lets recordings reach
their returns.

### On the full pending stack (#134 + #138 + #141 + #142 + #143)

Measured on a local tree carrying all pending PRs with the documented
merge resolutions:

- SC2K census: coverage **70.51% → 78.29% (+7.78 points)**;
  interpreted retirement −27.8% (111.7M → 80.6M); ops per native call
  19.01 → 19.06, so the conversions are whole subroutine bodies, not
  fragments. The two largest interpreted regions fell 71% and 58%.
- EV Override, three interleaved pairs: **+1.20 / +1.12 / +0.08
  points** (EV sits near saturation at ~78%; its largest remaining
  region turns out to be ~93% trace-covered already).
- Host instructions, six interleaved pairs each:
  - vs the stack without this change: **−2.58% median, 6/6 wins**,
    range −2.47..−2.77;
  - vs master: **−9.61% median, 6/6 wins**, range −9.45..−9.83
    (the stack was −7.24% without this change).
- Guest-work alignment: identical final tick within each census arm
  pair to 0.044% (cycle-headroom skew, two orders below the effect).

## Commit structure

The RTD `change_of_flow` commit is #144 verbatim and drops out on
rebase. The `ReturnExit` commit applies to #143's head; nothing in it
depends on #142's admissions or the gate change beyond sharing the
`ends_in_dynamic_exit` helper, so retargeting down the stack is
mechanical if the base PRs are reordered.

https://claude.ai/code/session_019xXVFS9G1t9hEi33kN7MPN

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -493,6 +493,21 @@ pub(crate) enum JitTraceOp {
     RtsReturn {
         expected_return: u32,
     },
+    /// A bare RTS/RTD terminal: the region is a subroutine body whose
+    /// return target differs per caller, so the return executes
+    /// architecturally -- pop, stack displacement, jump -- and the trace
+    /// exits at the popped address. Exit seeding compiles each caller's
+    /// continuation and ordinary bounded chaining connects them. This is
+    /// deliberately UNGUARDED: the exit is the architectural jump itself,
+    /// and the target-selection investigation measured that specializing
+    /// one dynamic target of a polymorphic site does not pay.
+    ReturnExit {
+        /// RTD's post-pop stack adjustment; 0 for RTS.
+        displacement: i16,
+        /// Decode-time constant: 16 for RTS, 20 for RTD (the
+        /// interpreter's charges).
+        cycles: i32,
+    },
     /// MOVE/MOVEA with at least one register-indirect operand, executed
     /// against the fastmem window (`dst == Addr` is MOVEA). Traces
     /// containing this op only run while a window is active; every access
@@ -1174,6 +1189,17 @@ impl TraceJit {
             if instr_budget < ops_len {
                 return None;
             }
+            // A ReturnExit completion lands at a per-caller continuation
+            // the trace cannot name statically. Those exits must seed
+            // candidacy the way guarded branch exits do: the clean-link
+            // rule only fires when the exit target is ALREADY a compiled
+            // head, which a fresh continuation never is, so without this
+            // the continuations after a returning subroutine would only
+            // ever compile by luck of a backward branch.
+            let ends_in_return_exit = trace
+                .ops
+                .last()
+                .is_some_and(|op| matches!(op.op, JitTraceOp::ReturnExit { .. }));
             // A generated loop clearly amortizes the ABI boundary for
             // profiled mixed 3+-op and read-only loops. A
             // two-op read/write MoveMem loop is already dominated by its two
@@ -1384,7 +1410,12 @@ impl TraceJit {
             if clean_link_exit {
                 super::trace_profile::note_link_exit(pc, cpu_type);
             }
-            if (guarded_branch_exit || clean_link_exit)
+            // A clean ReturnExit completion seeds its dynamic target
+            // (each caller's continuation) even when nothing is compiled
+            // there yet; see `ends_in_return_exit` above.
+            let return_exit_completion =
+                ends_in_return_exit && !guarded_branch_exit && !partial_call_this_entry;
+            if (guarded_branch_exit || clean_link_exit || return_exit_completion)
                 && !single_iter
                 && chain_budget > 0
                 && cpu.pc != pc
@@ -2300,8 +2331,16 @@ impl TraceJit {
                         return;
                     }
                     // An RTS with no pending call is the head function's
-                    // own return: not ours to record through.
-                    self.finish_recording(cpu, executed_pc, RecordingEnd::Region);
+                    // own return: record it as the region's dynamic-exit
+                    // terminal, exactly as the ungated recorder would.
+                    recording.ops.push(TraceBuildOp {
+                        op: JitTraceOp::ReturnExit {
+                            displacement: 0,
+                            cycles: 16,
+                        },
+                        ..call_op
+                    });
+                    self.finish_recording(cpu, next_pc, RecordingEnd::Region);
                     return;
                 }
                 _ => unreachable!(),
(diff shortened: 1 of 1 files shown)


benletchford:
The `ReturnExit` delta itself looks sound, but the current head cannot be merged safely: it contains #141, #142, #143, a duplicate of now-merged #144, and its own return-terminal commit. It therefore also inherits #142's blocking ADDA cycle-accounting bug, and it conflicts with the now-merged #134 trace changes.

Please first land/fix/restack the prerequisite chain on current `master`, then rebase this PR so its diff contains only `perf(jit): bare RTS/RTD becomes a dynamic-exit trace terminal`, and rerun CI. Keeping #144 separate also preserves its distinct `fix(cpu)` release-please entry.

benletchford:
I restacked this directly onto current master, combined the ReturnExit handling with the intervening PEA and TrapExit cases, and updated the two new recording fixtures for the if-conversion state added earlier in the stack. Local fmt, all-feature clippy, no-default-feature library tests (183), and all-feature library tests (292) pass.

PR #160, opened 2026-08-29
feat: real-hardware timing models for the 68030 and 68040

LinuxJedi:
This calibrates the 68030 and 68040 against real silicon. Until now both parts shared the legacy approximation (`(cycles*5+7)/8`, min 2, of the corrected 68000 counts); measurements from a real A4000 show that approximation over-bills plain 68040 execution by ~2-2.5x and misses the 030 in both directions.

**Measurement provenance.** Copperline's new `timing-test/accelprobe.asm` (CopperlineHQ/Copperline#592): a 32-row bootable probe measured against the CIA E-clock, run on a real A4000 with a 25 MHz Commodore 68030 CPU-slot board, a 25 MHz A3640 (68040), and a 50 MHz BFG9060 (68060, cross-checks). All captures are 9600-baud serial streams, two runs per machine; the A3640 pair was byte-identical. The probe's full columns and per-row verdicts live in its header.

**68030 -> the MC68020UM section-8 tables.** The 030's integer core is the 020's, and the real columns agree: taken `dbra` is 6 clocks and `mulu.w` lands in the mid-30s per loop on both the real A1200 020 (the campaign that produced the 020 model here) and the real A4000 030. Routed through the 020 tables, the emulated probe lands the reg-move loop within 0.5% of the real 030 column, `mulu` within 0.7%, and the fast-stack trap row within 0.4%, with no row regressing. The Cache/Worst case selection now feeds the 030 the full instruction-stream cache information, same as the 020.

**68040 -> a single-issue pipeline model (`timing_040.rs`).** Built on the 060 opcode classifier: one-clock pOEP-style costs for simple integer ops, `(raw/4).max(1)` fallbacks for data-dependent/unclassified costs, explicit iterative mul/div derivation, a ~12-clock CINV/CPUSH cost, static branch costs (the 040 has no branch cache), and a 4-clock flow-change floor. Against the real A3640 column: the measured `move.w`+`dbra` loop (4.08 clk/iter) lands tick-exact, `mulu.w` within 6%, `cpushl` within 7%. `DBcc`-taken is 3 clocks so the ubiquitous one-clock-body loop hits the measured 4 clk/iter; the empty `dbra` loop reads 3 against the measured 4.06 -- the documented compromise for a model without an execute-overlap stage (the body loop is the shape real code runs). Exception entries keep the legacy scaled costs, so exception timing is unchanged.

**Consistency.** The trace-JIT baked bit-op costs mirror the new dispatch so traces retire identically to the interpreter (covered by the existing cross-check test). The DBF gap test now observes loop completion through an explicit landing pad instead of a cycle-budget-dependent PC checkpoint, which the cheaper (correct) 040 costs let slip past inside one `execute()` call.

**End-to-end effect.** In Copperline, an AmigaVision 68040 boot that took 17 emulated seconds (reported by users as 3x slower than the same setup elsewhere) completes in ~12s with this change, and the fast-RAM configuration lands at 5s -- while the machine's bus-bound rows stay put, which is what the real-hardware columns say they should do.

All 62 test suites pass; fmt and clippy clean (the one pre-existing `unnecessary_sort_by` warning in the SingleStep tests is untouched).

### DIFF
--- a/src/core/cpu.rs
+++ b/src/core/cpu.rs
@@ -478,10 +478,12 @@ impl Default for CpuCore {
 }
 
 impl CpuCore {
-    /// Legacy 68030/040 approximation from the handlers' corrected 68000
-    /// counts. The 68020/68EC020 is routed through its MC68020UM section-8
-    /// timing model before this function; the 68060 has a separate pipeline
-    /// engine. The 68000/68010 paths remain untouched.
+    /// Legacy scaling approximation from the handlers' corrected 68000
+    /// counts. The 020/030 route through the MC68020UM section-8 timing
+    /// model and the 040 and 060 through their pipeline engines before this
+    /// function; it remains the cost model for 040 exception entries, the
+    /// SCC68070, and the trace-JIT builders' baked fallbacks. The
+    /// 68000/68010 paths remain untouched.
     #[inline]
     pub(crate) fn scale_cycles_for_cpu_type(&self, cycles: i32) -> i32 {
         use crate::core::types::CpuType;
--- a/src/core/execute.rs
+++ b/src/core/execute.rs
@@ -867,8 +867,13 @@ impl CpuCore {
         let result = dispatch(self, bus, self.ir as u16);
         let fetch_cached = if matches!(
             self.cpu_type,
-            super::types::CpuType::M68EC020 | super::types::CpuType::M68020
+            super::types::CpuType::M68EC020
+                | super::types::CpuType::M68020
+                | super::types::CpuType::M68EC030
+                | super::types::CpuType::M68030
         ) {
+            // The 020/030 MC68020UM tables pick Cache Case only when every
+            // instruction-stream fetch hit the cache, not just the opcode.
             bus.instruction_fetches_were_cached()
         } else {
             opcode_fetch_cached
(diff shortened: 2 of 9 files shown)


LinuxJedi:
Both comments addressed in 5b40fcd: the trace-JIT baking comment now describes the actual dispatch (020/030 tables at four clocks, 040/060 pipelines at one, legacy scaler only for the remaining models), and the timing_040 residuals paragraph no longer claims the simplifications are uniformly slow-side -- it now calls out explicitly that the 060 class costs can be optimistic for some shift/bit forms on real 040 silicon.

benletchford (changes requested):
The new timing model is directionally sound, but `CpuCore::execute()` still computes `fetch_cached` using full instruction-stream cache status only for M68EC020/M68020 (`src/core/execute.rs`, around lines 147-152). The corresponding `step_with_dispatch()` and `step_with_hle_handler()` branches correctly include M68EC030/M68030. As written, a 030 instruction whose opcode fetch hits but whose extension-word fetch misses selects the 020 table Cache Case under `execute()` and the Worst Case under `step()`, so timing varies by execution API and the PR’s stated full-stream-cache behavior is not applied on the primary execution path. Please include the two 030 variants in the `execute()` match as well, ideally with a regression test that compares `execute()` and `step()` for this mixed-hit instruction-stream case.

benletchford (approved):
Maintainer follow-up 8202cdc fixes the requested execution-path inconsistency by sharing the full-fetch-stream CPU-family predicate across execute, step, and HLE, with a focused regression guard. Local formatting, targeted tests, the full all-features suite, and strict Clippy are clean.

PR #162, opened 2026-08-30
perf(decode): decoded fast paths for immediate register bit ops, LINK/UNLK, and MOVEM.L pre/post-increment

rlanday:
## Summary

A slow-path opcode census on a real workload (SimCity 2000 under Systemless, 1.5B instructions of scripted gameplay) showed that five opcodes account for 82% of everything still executed on the undecoded interpret path: `BTST #imm,Dn` alone is 46.8%, followed by `MOVEM.L (SP)+` (11.7%), `MOVEM.L -(SP)` (9.3%), `UNLK` (7.8%) and `LINK` (6.8%) — bit tests in hot loops plus function prologues/epilogues. This change gives all of them decoded fast paths:

- `DecodedMemOp::BitImmDataReg` — BTST/BCHG/BCLR/BSET `#imm,Dn` (long-width register bit ops; only Z is affected, from the pre-modification value).
- `DecodedMemOp::Link` / `Unlk` — **A7 is excluded at decode** and stays on the slow path, preserving the documented trace-blocker contract for `LINK A7`/`UNLK A7` (`blocked_recording_salvages_the_prefix_through_the_last_branch` covers it). UNLK consumes no extension word; the executors advance the PC only past extension words they consume.
- `DecodedMemOp::MovemLongPreDec` / `MovemLongPostInc` — `MOVEM.L list,-(An)` and `(An)+,list` through the fastmem window; a mask that includes the base register bails to the slow path (version-dependent stored/loaded value), as do misaligned pre-68020 addresses and window misses. All bails return before any state mutation.

## Measurements

Paired, interleaved A/B on the deterministic Systemless workload (400M instructions of scripted SimCity 2000 boot-into-gameplay, host `instructions retired` via `/usr/bin/time -l`, periodic screenshots disabled):

| pair | before | after | ratio |
|---|---|---|---|
| 1 | 83,702,875,957 | 78,651,571,443 | 0.9397 |
| 2 | 83,547,556,547 | 78,399,765,486 | 0.9384 |

**−6.10% total host instructions.** Correctness: all four runs end at exactly the same guest tick count (182,168), and the final-frame screenshot is byte-identical between the before/after binaries.

## Validation

- `cargo test` — full suite including the Musashi fixture tests
- New unit tests: mono bit-op semantics and Z-only flags; LINK/UNLK frame roundtrip incl. the `LINK A7` bail; MOVEM.L predec/postinc roundtrip, base-in-mask bail, and PC-untouched-on-bail
- `cargo fmt --check`, `cargo clippy`, `cargo doc` clean

https://claude.ai/code/session_01LnuSFJRzTWFhxrFUMw622y

### DIFF
--- a/src/core/mem_ops.rs
+++ b/src/core/mem_ops.rs
@@ -123,6 +123,22 @@ pub(crate) enum DecodedMemOp {
     MovemWordPostInc {
         base: u8,
     },
+    BitImmDataReg {
+        op: BitOp,
+        reg: u8,
+    },
+    Link {
+        reg: u8,
+    },
+    Unlk {
+        reg: u8,
+    },
+    MovemLongPreDec {
+        base: u8,
+    },
+    MovemLongPostInc {
+        base: u8,
+    },
     /// ADD/SUB/AND/OR/CMP `<ea>,Dn`
     AluToReg {
         op: BinaryOp,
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Reviewed the decoded fast paths against the interpreter and existing trace implementations. Immediate bit operations preserve non-Z flags, LINK/UNLK exclude the generation-sensitive A7 forms, and MOVEM bails before mutation for alignment, window, empty-mask, and base-in-list cases. The branch rebases cleanly onto current master; formatting, the complete all-feature suite, and strict Clippy pass.

PR #168, opened 2026-09-02
perf(trace): bypass admission for compiled exits

rlanday:
## Stack note

This is stacked on #167. Until #167 lands, the comparison against `master`
also contains its nested-forward-call work (and #166 below that). The new
change for this PR is only:

```text
d7cb97c perf(trace): bypass admission for compiled exits
```

## Summary

Bypass exit-admission bookkeeping when a trace has already proven that its
exit target is an exact compiled neighbor.

This reduces actual host CPU work in applications with frequent
compiled-to-compiled transitions:

- SimCity 2000: **-0.861% aggregate host instructions** across three 400M
  pairs (**-0.913% by pair medians**)
- Lemmings: **-1.170% host instructions** across two 200M pairs
- 3 in Three, System's Twilight, and EV Override: within +/-0.03% host
  instructions (neutral)

The patch changes no trace-admission thresholds, cache ownership, generated
code, validation, chaining depth, guest instruction counts, or modeled 68k
cycles.

## What was unnecessarily expensive

After a compiled trace exits, `try_execute` determines whether the exit was:

- a guarded branch side exit;
- a clean completion onto a compiled link target; or
- a seeded dynamic exit such as #167's `CallExit`/`ReturnExit` continuation.

The steady-state action for an exact compiled target is simply `Chain`: enter
that trace, subject to the existing watch and chain-budget rules.

Previously every qualifying exit called `note_trace_exit`. That function is
primarily the admission-policy state machine for targets which are not yet
compiled. Before it can return `Chain` for an already-compiled target, it:

1. recomputes the direct-mapped trace-cache index;
2. clears the target's durable linear-deferral entry;
3. probes the two-way earned-call-permission table;
4. probes the two-way structural-rejection table; and
5. rechecks the cache slot, full PC, and CPU model.

Those operations are useful for empty, counting, rejected, aliased, and
wrong-CPU slots. They cannot affect the decision for an exact compiled
neighbor. A successfully compiled head has already cleared its durable
linear-deferral marker, and the other two side-table probes are read-only.

This became visible after #167 made SC2K's caller/callee/continuation graph
compile successfully. In its 400M opportunity profile, the hottest trace
chained 3,097,192 times and its descendants retired 65,686,660 guest
instructions. A 20-second production stack sample attributed 288 of 15,796
top-of-stack samples (1.82%) directly to `note_trace_exit`; some related lookup
and hashing work was attributed separately.

## The fast path

`try_execute` now resolves the exact compiled-neighbor case before calling the
admission helper:

- a clean link reuses the `compiled_head_at` result it already computed;
- guarded and seeded exits perform that exact slot/PC/CPU lookup once;
- an unwatched exact compiled target returns `Chain` directly;
- a watched exact compiled target still refuses to chain, preserving the
  runner-visible boundary; and
- every noncompiled or mismatched target still enters `note_trace_exit`
  unchanged.

Cache replacement and self-modifying-code invalidation naturally remove the
exact compiled slot. The next exit then falls through to the unchanged
admission helper, so the shortcut cannot keep a stale link alive.

An intermediate version moved the compiled check to the beginning of
`note_trace_exit`. That measured -0.55% SC2K host instructions. Keeping the
check in `try_execute` instead avoids the outlined helper call entirely and
reuses the lookup already done for clean links. The final minimal caller-only
form measures better and adds only 14 lines.

## Host-CPU measurements

Each arm is a release/fat-LTO Systemless binary. The harness uses fresh game
copies/save directories and alternates A/B then B/A order. The primary metric
is macOS hardware `instructions retired`: actual work executed by the host,
not modeled guest cycles. Host cycles and process user time are noisier
corroboration.

| Workload | Guest budget | Pairs | Host instructions | Host cycles | User CPU | Correctness |
| SimCity 2000 | 400M | 3 | **-0.861% aggregate; -0.913% median** | -1.249% | -0.849% | 185,810 ticks/run |
| Lemmings | 200M | 2 | **-1.170%** | -1.243% | -0.517% | 36,409 ticks/run; identical hash |
| 3 in Three | 200M | 2 | +0.023% | +1.526% | +1.299% | 822,511 ticks/run; identical hash |
| System's Twilight | 100M | 2 | -0.022% | -2.039% | -2.730% | 685,705 ticks/run; identical hash |
| EV Override | 200M | 4 | +0.027%/tick | -0.595%/tick | -1.027%/tick | 2,362,189 vs 2,362,099 aggregate ticks |

The non-SC2K/Lemmings instruction movements are within +/-0.03% and are
reported as neutral. Their larger timing movements do not correspond to
changed executed work. EV is normalized by aggregate simulated ticks because
its deterministic input replay can follow slightly different flight paths.

SC2K raw host-instruction counts:

```text
#167 base: 56,476,614,059  56,414,528,626  56,514,915,269
candidate: 55,961,053,451  55,930,939,682  56,054,854,493
aggregate delta: -0.861368%
median delta:    -0.912865%
```

The later SC2K timing runs encountered a changing machine/thermal state, but
the host-instruction reduction remained stable in every pair. This is why the
hardware work counter, rather than elapsed time, is the acceptance gate.

## Related depth-limit experiments (rejected)

The same investigation tested whether #167 was leaving useful compiled work
beyond the existing three-link execution limit.

### Admit exits even when the chain budget is exhausted

The current depth guard suppresses both recursive execution and exit-admission
evidence. A prototype continued counting/promoting the fourth-level target
without executing another nested trace. It passed the full tests but made
SC2K about **0.34% worse** in retired host instructions: the extra bookkeeping
found too little useful code to repay itself.

### Raise recursive chain depth from 3 to 32

This bounded proxy for removing the execution limit was neutral on SC2K:
**+0.004% by medians** and +0.020% by aggregate host instructions. Useful
compiled chains overwhelmingly fit inside the existing limit.

Literally removing the limit is unsafe in the current recursive design: a
compiled cycle can grow the host stack until the batch budget expires. A
constant-stack iterative dispatcher could remove that risk, but the 3-versus-32
result shows no substantive performance reason to make that larger refactor.

Both rejected prototypes remain preserved locally.

## Verification

- `cargo test --lib --features jit`: 275 passed
- `cargo test --lib --no-default-features`: 201 passed
- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- deterministic workloads above completed with equal ticks and identical
  final-frame hashes; EV used per-tick normalization as described above

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -762,6 +762,120 @@ struct TraceCodeSegment {
     len: u32,
 }
 
+/// Compare live guest instructions with the bytes captured at compilation.
+///
+/// The helper stays outlined: trace entry already pays one call per segment,
+/// and inlining the SIMD loop expands the iterator closure enough to disturb
+/// instruction-cache locality in workloads dominated by tiny segments. Short
+/// ranges use early-out native-word loads; on x86-64, genuinely wide ranges
+/// fold 16-byte differences into one final decision. Very long ranges retain
+/// the platform `memcmp`, whose startup cost is then well amortized.
+#[inline(never)]
+unsafe fn trace_code_eq(live: *const u8, expected: *const u8, len: usize) -> bool {
+    #[cfg(target_arch = "x86_64")]
+    if (48..=512).contains(&len) {
+        use std::arch::x86_64::{
+            __m128i, _mm_cmpeq_epi8, _mm_loadu_si128, _mm_movemask_epi8, _mm_or_si128,
+            _mm_setzero_si128, _mm_xor_si128,
+        };
+
+        let mut offset = 0;
+        // SAFETY: SSE2 is part of the x86-64 baseline ISA.
+        let zero = unsafe { _mm_setzero_si128() };
+        let mut difference = zero;
+        while len - offset >= 16 {
+            // SAFETY: each complete vector lies inside the caller-provided
+            // `len`-byte ranges. Unaligned loads are intentional.
+            let live_word = unsafe { _mm_loadu_si128(live.add(offset).cast::<__m128i>()) };
+            let expected_word = unsafe { _mm_loadu_si128(expected.add(offset).cast::<__m128i>()) };
+            // SAFETY: SSE2 is part of the x86-64 baseline ISA.
+            difference =
+                unsafe { _mm_or_si128(difference, _mm_xor_si128(live_word, expected_word)) };
+            offset += 16;
+        }
+
+        let mut tail_difference = 0u64;
+        if len - offset >= 8 {
+            // SAFETY: a complete eight-byte tail remains in both ranges.
+            let live_word = unsafe { std::ptr::read_unaligned(live.add(offset).cast::<u64>()) };
+            let expected_word =
+                unsafe { std::ptr::read_unaligned(expected.add(offset).cast::<u64>()) };
+            tail_difference |= live_word ^ expected_word;
+            offset += 8;
+        }
+        if len - offset >= 4 {
+            // SAFETY: a complete four-byte tail remains in both ranges.
+            let live_word = unsafe { std::ptr::read_unaligned(live.add(offset).cast::<u32>()) };
+            let expected_word =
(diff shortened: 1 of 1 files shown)


benletchford:

PR #169, opened 2026-09-02
perf(trace): extend indirect-call and immediate-memory coverage

rlanday:
> [!NOTE]
> This PR is stacked on #168, which is itself stacked on #167. The new work
> begins after `d7cb97c` and is kept in three independent commits:
> `b3934aa`, `6a80fe0`, and `4706bcc`. Once the dependencies merge, the diff
> collapses to those commits.

## Summary

Extend native and portable trace execution across three legal 68k forms that
were terminating heavily reused SimCity 2000 recordings:

- `JSR d16(An)` (`4EAD` in the measured path);
- `CMPI.B #imm,abs.W` (`0C38`);
- `ANDI.B #imm,d8(An,Xn)` (`0231`).

Together they reduce retired host instructions by **4.40%** in the standard
400M-instruction SC2K replay, with identical simulated ticks and framebuffer
output. Each step was also measured independently, and cross-application
controls are neutral.

The blocker profile is used here to prioritize implementation work, not to
define a permanent game-specific whitelist. These are extensions of existing
control-address and immediate-ALU families. Other legal forms continue to run
through the exact decoded/interpreter path, and the trace representation can
grow toward shared family lowerings as their fault/extension-word cases are
made atomic. Admission—not semantic support—should ultimately decide whether
a particular trace is worth compiling.

## Why unsupported forms cost host CPU

The trace recorder stops at an instruction it cannot lower. The cost is not
just dispatching that one instruction: the supported prefix before it cannot
become a reusable native loop. The pre-change 400M opportunity census
projected:

| Blocker | Projected prefix dispatches |
|---|---:|
| `JSR d16(An)` | 1,474,272 |
| `CMPI.B #imm,abs.W` | 1,286,754 + 708,462 |
| `ANDI.B #imm,d8(An,Xn)` | 1,474,725 |

The ANDI result makes the effect especially visible. After adding it, the
rejected head at `$00221014` becomes a 10-op native trace. It retires
2,943,680 guest operations in only 2,803 native entries—1,050 operations per
entry. Across the whole replay:

| Trace counter | Before ANDI | After ANDI | Delta |
|---|---:|---:|---:|
| JIT-retired guest ops | 341,714,179 | 349,981,182 | +8,267,003 |
| rejected hits | 1,825,305 | 1,520,925 | -304,380 |
| native calls | 15,895,996 | 15,586,123 | -309,873 |

This is why implementing one blocker saves much more than one decoded opcode:
it recovers the surrounding long-running loop.

### Displacement indirect JSR

The existing terminal `IndirectJsr` op now carries a predecoded `JitEa`, so
the same path supports `(An)` and `d16(An)`. The displacement word is captured
and validated, while the base address register remains live at execution.

The native path computes the target, checks the complete four-byte stack write,
and only then commits SP, the big-endian return address, flow state, and PC. A
window miss therefore re-executes the call with no partial state. The return
address is derived from the recorded instruction extent (four bytes for the
new form), and the M68000 bound is the exact 18 cycles rather than the 16-cycle
`(An)` value. The portable path delegates to the decoded memory/control
executor and has the same atomic fallback.

### Absolute-byte CMPI

`CmpiByteAbs` captures the byte immediate and sign-extended absolute-short
address while retaining both raw extension words for code validation. Native
execution performs one checked byte load and subtraction, updates NZVC, and
preserves X. It never writes guest memory; a window miss occurs before flags
change. Portable execution reuses `DecodedMemOp::AluImm`.

This is deliberately a separate internal op from the established
`CmpiWordMem`. A first candidate generalized that hot op to carry runtime size.
Although lowering consumes size at compile time, the representation/code-layout
change correlated with small repeatable regressions in 3 in Three and EV
Override. Keeping the fixed word op byte-for-byte unchanged retained the SC2K
gain and removed those regressions; see "Rejected and retained experiments."

### Indexed-byte ANDI

`AndImmByteMem` captures the immediate and decoded brief-indexed EA. The native
path computes the live base/index address, then performs all window and
self-modifying-code overlap guards before the read-modify-write or condition
flags commit. It writes one byte, sets logical NZVC, preserves X, and uses the
exact 22-cycle M68000 maximum. Portable execution again delegates to the
decoded memory-ALU implementation.

Tests cover decoder extents, signed displacement and live-register behavior,
exact M68000 state/cycle parity, native/portable parity, short-window bailout,
and code-overlap bailout with no partial state.

## Host-CPU measurements

All SC2K arms execute 400M guest instructions and finish at the same 185,810
ticks and framebuffer SHA-256. Values are aggregate retired host instructions
from three interleaved A/B pairs:

| Incremental step | Baseline | Candidate | Delta |
|---|---:|---:|---:|
| add `JSR d16(An)` | 168,092,724,217 | 167,247,845,394 | **-0.503%** |
| add isolated `CMPI.B abs.W` | 167,208,637,560 | 165,382,873,087 | **-1.092%** |
| add indexed `ANDI.B` | 165,827,019,725 | 160,877,832,958 | **-2.985%** |
| all three vs pre-JSR | 168,352,532,780 | 160,946,270,841 | **-4.399%** |

The direct cumulative pairs are independently -4.218%, -4.429%, and -4.550%.
The direct -4.399% agrees with the -4.42% compounded prediction from the
incremental steps.

Cross-workload retired-instruction controls for each incremental step:

| Workload | JSR | isolated CMPI | ANDI |
|---|---:|---:|---:|
| Lemmings | +0.026% | -0.017% | +0.001% |
| 3 in Three | +0.046% | +0.0006% | +0.0003% |
| System's Twilight | +0.013% | -0.005% | -0.045% |
| EV Override (per tick) | -0.016% | +0.0059% | +0.018% |

EV is normalized by aggregate simulated ticks because its scripted route is
not tick-deterministic. All other controls have identical ticks and framebuffer
hashes. The small values above are noise-level neutrality, not claimed wins.

## Why the old ANDI rejection changed

The same ANDI implementation measured +0.016% on a single September 1 pair
and was preserved rather than published. That test used the PR-#162-era stack.
Parking had removed the cheap spin loops that were the trace JIT's former
customers; on the then-current SC2K run, native traces retired only 0.07% of
guest instructions, so admitting this opcode had almost nothing to act on.

The later admission and exit-topology work restored a reusable native trace
graph. The current pre-ANDI census retires about 85% of the fixed guest budget
through native traces. Reapplying the unchanged opcode implementation now
recovers the 1,050-op-per-entry loop described above and saves 2.985% host
instructions. Opcode-performance rejections are therefore conditional on the
admission/topology baseline; they should be revisited after structural changes.

## Rejected and retained experiments

- **Generalized CMPI representation:** SC2K improved 1.084%, but 3 in Three
  regressed 0.167% and EV regressed 0.132%/tick. The isolated `CmpiByteAbs`
  control improved SC2K 1.092% while 3 in Three moved only +0.0006% and EV
  +0.0059%/tick. The generalized source and both binaries remain preserved for
  follow-up assembly/layout analysis.
- **`JMP (An)`:** six SC2K pairs measured -0.046%, noise-level neutral. Source
  and binary are retained locally in case another topology change makes it
  useful.
- **`LEA d16(PC),An`:** three SC2K pairs measured +0.042%, also neutral. It is
  retained as instruction-family coverage work, but is not included as a
  performance claim here.
- **Cranelift verifier removal:** not part of this work. Verification remains
  enabled; this PR does not trade safety for compile speed.

## Validation

- `cargo test --lib --features 'jit trace-profile'` — 318 passed
- `cargo test --lib --no-default-features` — 205 passed
- `cargo clippy --lib --all-features -- -D warnings` — clean
- deterministic headless A/B hardware-counter replays on SC2K, Lemmings,
  3 in Three, System's Twilight, and EV Override

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -762,6 +762,120 @@ struct TraceCodeSegment {
     len: u32,
 }
 
+/// Compare live guest instructions with the bytes captured at compilation.
+///
+/// The helper stays outlined: trace entry already pays one call per segment,
+/// and inlining the SIMD loop expands the iterator closure enough to disturb
+/// instruction-cache locality in workloads dominated by tiny segments. Short
+/// ranges use early-out native-word loads; on x86-64, genuinely wide ranges
+/// fold 16-byte differences into one final decision. Very long ranges retain
+/// the platform `memcmp`, whose startup cost is then well amortized.
+#[inline(never)]
+unsafe fn trace_code_eq(live: *const u8, expected: *const u8, len: usize) -> bool {
+    #[cfg(target_arch = "x86_64")]
+    if (48..=512).contains(&len) {
+        use std::arch::x86_64::{
+            __m128i, _mm_cmpeq_epi8, _mm_loadu_si128, _mm_movemask_epi8, _mm_or_si128,
+            _mm_setzero_si128, _mm_xor_si128,
+        };
+
+        let mut offset = 0;
+        // SAFETY: SSE2 is part of the x86-64 baseline ISA.
+        let zero = unsafe { _mm_setzero_si128() };
+        let mut difference = zero;
+        while len - offset >= 16 {
+            // SAFETY: each complete vector lies inside the caller-provided
+            // `len`-byte ranges. Unaligned loads are intentional.
+            let live_word = unsafe { _mm_loadu_si128(live.add(offset).cast::<__m128i>()) };
+            let expected_word = unsafe { _mm_loadu_si128(expected.add(offset).cast::<__m128i>()) };
+            // SAFETY: SSE2 is part of the x86-64 baseline ISA.
+            difference =
+                unsafe { _mm_or_si128(difference, _mm_xor_si128(live_word, expected_word)) };
+            offset += 16;
+        }
+
+        let mut tail_difference = 0u64;
+        if len - offset >= 8 {
+            // SAFETY: a complete eight-byte tail remains in both ranges.
+            let live_word = unsafe { std::ptr::read_unaligned(live.add(offset).cast::<u64>()) };
+            let expected_word =
+                unsafe { std::ptr::read_unaligned(expected.add(offset).cast::<u64>()) };
+            tail_difference |= live_word ^ expected_word;
+            offset += 8;
+        }
+        if len - offset >= 4 {
+            // SAFETY: a complete four-byte tail remains in both ranges.
+            let live_word = unsafe { std::ptr::read_unaligned(live.add(offset).cast::<u32>()) };
+            let expected_word =
+                unsafe { std::ptr::read_unaligned(expected.add(offset).cast::<u32>()) };
+            tail_difference |= u64::from(live_word ^ expected_word);
+            offset += 4;
+        }
+        if len - offset >= 2 {
+            // SAFETY: a complete two-byte tail remains in both ranges.
+            let live_word = unsafe { std::ptr::read_unaligned(live.add(offset).cast::<u16>()) };
+            let expected_word =
+                unsafe { std::ptr::read_unaligned(expected.add(offset).cast::<u16>()) };
+            tail_difference |= u64::from(live_word ^ expected_word);
+            offset += 2;
+        }
+        if len != offset {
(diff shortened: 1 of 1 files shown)


benletchford:

PR #170, opened 2026-09-03
perf(trace): compile indexed-dispatch paths across call spans

rlanday:
> [!NOTE]
> This PR is stacked on #169. The new work begins after `4706bcc` and is
> kept in four commits: `4eb51b9`, `5bffbbe`, `5cc2718`, and `212748f`.
> Once the dependency merges, the diff collapses to those commits.

## Summary

Compile a hot table-dispatch path that SimCity 2000's recorder could observe
but previously threw away:

- admit `LEA d16(PC),An`;
- represent and execute indirect `JMP` through `(An)`, `d16(An)`, brief
  `d8(An,Xn)`, and brief `d8(PC,Xn)`;
- record through an indirect jump by guarding the observed target rather than
  pretending that a dynamic target is static;
- when a later second call would make the bounded callee-code interval too
  broad, retain the completed prefix through its last safe terminal—but only
  when that prefix belongs to an indexed-dispatch path.

The complete stack reduces retired host instructions by **1.19%** in the
standard 400M-instruction SC2K replay, with identical simulated ticks and
framebuffer output. 3 in Three, System's Twilight, Lemmings, and EV Override
are neutral.

This does not skip guest work, loosen code validation, or widen the allowed
self-modifying-code intervals. It moves an already-observed, repeatedly
executed guest prefix from decoded execution into the existing guarded native
trace machinery.

## The failed SC2K topology

The hot head at guest PC `$002891A8` executes this shape:

```text
$002891A8  setup and compare operations
     ...
$002891BE  LEA d16(PC),An
$002891C2  JMP d8(An,Xn)          ; table dispatch, target observed at runtime
     ...   selected dispatch case
             BSR                  ; first bounded callee
             ... callee ...
             RTS                  ; checked return to the recorded caller
             BRA                  ; another completed control-flow boundary
     ...
$00289288  BSR                    ; second, far-away callee
```

Before this stack, the PC-relative LEA and indexed jump stopped recording. The
new control-flow representation lets the recorder follow the observed table
case, but exposes a second structural limit: one trace carries one caller code
interval and one callee code interval for self-modifying-code validation. The
first call fits. Adding the second far callee would make the single callee
interval exceed `CALL_THROUGH_MAX_SPAN`.

That span cap exists for a good reason. Treating two small, distant callees as
one enormous interval would make each trace entry validate unrelated memory
between them and would increase both false invalidation and steady-state host
work. The recorder therefore must refuse the second call.

The old refusal, however, discarded too much. It closed the recording as an
ordinary region while the last recorded operation was not a terminal.
Compilation returned `NoTraceTerminal`; repeated identical observations then
made the head structurally rejected. In a 400M parent profile this one head
had 183,834 hits, 183,623 rejected hits, five recording attempts, and zero
compiled calls. The second call poisoned the independently valid prefix before
it.

## The rule

At a call-span refusal, an out-of-line helper now:

1. verifies that the incomplete recording is not already terminal;
2. finds its last completed trace terminal;
3. requires a guarded brief-indexed or PC-indexed `JMP` at or before that
   terminal;
4. retains the existing `SALVAGE_MIN_OPS` profitability floor;
5. truncates to that last terminal and hands the prefix back to the ordinary
   linear admission/deferral path.

The indexed jump is a **topology anchor**, not necessarily the truncation
point. That distinction accounts for most of the win: SC2K performs useful
work after the dispatch, including the first call/return, before reaching the
later direct branch. The final trace keeps that work and leaves the impossible
second call plus its tail to full dispatch.

The structural restriction matters. A generic “call span means salvage” rule
also matched cold function thunks and data-dependent alternate paths in other
programs. An indexed dispatch says something stronger: this is a multi-way
control-flow join whose observed case is worth guarding. A plain `JMP (An)` or
ordinary conditional branch does not establish that topology.

The helper is `#[cold]` and `#[inline(never)]`. The vector scan and policy code
therefore do not change the common trace-entry, native-execution, or recording
layout; they run only when a recording has already reached this uncommon
second-callee span refusal.

## Guarded indirect JMP semantics

`IndirectJmp` carries a decoded `JitEa` plus an optional observed target.
During recording, the jump first decodes without a prediction. Once the guest
executes it, the recorder stores the actual successor and continues along that
case.

Native and portable execution both:

- recompute the target from the live address/index registers;
- apply 68000/68020 index-width and scale semantics;
- commit the architectural jump (`pc` and `change_of_flow`);
- compare the committed target with the recorded successor;
- continue natively on a match or side-exit with exact `ppc`/`ir` state on a
  mismatch.

Thus the trace specializes an observed dispatch edge without assuming the
jump is immutable. Existing guarded-exit accounting can re-record if another
table case becomes dominant. Raw opcode extension words remain in the trace's
code snapshot, and every discontiguous caller/callee segment is validated
before native execution.

`LEA d16(PC),An` records its PC-derived base as a constant plus the signed
displacement and preserves condition codes, matching the interpreter. The
indirect-JMP family uses the existing checked EA representation rather than a
game-specific address or opcode whitelist.

## Actual host-CPU measurements

All fixed-endpoint runs alternate A/B order. The primary metric is retired
**host** instructions: actual CPU work performed by Systemless, not emulated
68k cycles and not merely the number of guest instructions. Each SC2K arm runs
the same 400M guest instructions, ends at exactly 185,810 simulated ticks, and
has the same final framebuffer SHA-256.

Complete PR stack versus the exact #169 artifact, three interleaved pairs:

| Workload | Budget | #169 host instructions | Candidate | Change |
|---|---:|---:|---:|---:|
| SimCity 2000 | 400M | 160,600,336,338 | 158,688,715,569 | **-1.1903%** |
| 3 in Three | 200M | 284,822,684,554 | 284,763,977,068 | -0.0206% |
| System's Twilight | 100M | 304,968,183,095 | 305,000,953,454 | +0.0107% |
| Lemmings | 400M | 121,473,897,905 | 121,423,993,241 | -0.0411% |
| EV Override | 200M | normalized per tick | normalized per tick | +0.0759% |

SC2K's individual pairs were approximately -1.27%, -1.08%, and -1.23%; all
three were favorable. Ticks and screenshots are identical for every workload
except EV Override, whose scripted endpoint is intentionally tick-variable;
EV is therefore normalized by aggregate ticks. The control-workload changes
are noise-scale neutrality, not claimed wins.

To isolate only the final salvage policy from its indirect-JMP parent, a
second three-pair SC2K comparison measured **-1.2410%** host instructions
(160,879,368,743 -> 158,882,823,422). This shows the gain comes from retaining
the safe native prefix, not from holding guest work or simulated time fixed in
a misleading way.

### 1. Salvage every call-span recording

The first prototype treated every call-span stop like an unsupported-opcode
blocker and salvaged through the last terminal. It improved SC2K by **1.0816%**
but regressed 3 in Three by **+0.1506%** and System's Twilight by **+0.1529%**.
Lemmings was -0.0822%; EV Override was -0.0240% instructions/tick.

Profiles explained the conflicting results:

- SC2K has the hot indexed table-dispatch path described above.
- 3 in Three's superficially similar failures end in plain `JMP (A0)` function
  thunks. The observed heads had only 5-9 hits, far below the 208-hit
  independent-linear threshold, and produced no useful compiled shape.
- System's Twilight's `$003A2B06` head already had a productive 43-op trace
  (94 native calls, 3,948 retired guest ops). A data-dependent alternate path
  reached the span refusal after an ordinary conditional branch, so broad
  salvage could replace the useful long shape with a shorter alternate one.

Even where the broad rule compiled nothing, adding a hot-path enum case and
policy branch correlated with a repeatable ~0.15% representation/layout cost.
Moving the narrow rule out of line removed that effect.

### 2. Keep only the indexed-JMP prefix

A conservative version required the right topology but truncated exactly at
the indexed jump—an eight-op SC2K trace. It removed the other-game regressions
but saved only **0.4252%** on SC2K. It threw away the useful first call/return
and later direct branch. The final rule uses the indexed jump as proof of the
topology and the last later terminal as the safe endpoint.

### 3. Refuse salvage after any previous compile

The first indexed-only version also refused heads that had compiled at any
point in the past, to protect an existing shape. Removing that history guard
produced only **0.2768%** SC2K improvement in a fresh three-pair run, so blocked
adaptive re-recording was not the missing opportunity. The indexed topology
test already excludes Twilight's conditional alternate path; a durable
compiled-history exception is unnecessary and would interfere with the
existing dominant-target adaptation policy.

### 4. Follow the second call too

Another prototype appended a `CallExit` for the second far call, compiled the
parent through it, and seeded that callee separately. Against prefix salvage,
it was **0.6948% worse on SC2K** over three 400M pairs. It duplicated an
already-productive downstream trace graph and added validation/chaining work.
The second call remains outside this trace.

### 5. Judge the new control forms in isolation

Earlier isolated tests found `LEA d16(PC),An` at +0.042% and terminal
`JMP (An)` at -0.046% on SC2K—both neutral. They were retained locally rather
than claimed as wins. The present result shows why opcode admission and trace
topology cannot always be evaluated independently: those forms make the hot
dispatch path recordable, while call-span salvage lets the newly visible path
compile. The complete stack directly measured -1.1903% against #169.

## Validation

The new end-to-end test constructs:

```text
indexed dispatch -> first far BSR -> callee -> checked RTS
                 -> direct branch -> second far BSR
```

It runs the guest loop to a fixed budget, verifies that both callees execute in
lockstep, and inspects the compiled head. The resulting 12-op trace must contain
the guarded indexed dispatch, bounded first call, checked return, and final
direct branch, and must exclude the second call.

Also completed:

- `cargo test --features jit`
- `cargo test --no-default-features`
- `cargo fmt --check`
- `git diff --check`
- deterministic headless hardware-counter replays on SC2K, 3 in Three,
  System's Twilight, Lemmings, and EV Override

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -2172,6 +2172,59 @@ impl TraceJit {
         }
     }
 
+    /// Close a recording that cannot add another callee without making its
+    /// bounded code intervals overly broad, retaining a completed prefix
+    /// when it belongs to an indexed-dispatch path.
+    #[cold]
+    #[inline(never)]
+    fn finish_recording_at_call_span(&mut self, cpu: &mut CpuCore, call_pc: u32) {
+        // A trace carries one caller and one callee SMC interval. SC2K's
+        // table-driven dispatch can cross an indexed JMP, finish one
+        // call/return, and then reach a second far callee. The second call
+        // cannot join this trace without making its callee interval overly
+        // broad, but the completed control-flow prefix before it is
+        // independently useful.
+        //
+        // Keep this salvage deliberately structural. An indexed jump is a
+        // multi-way dispatch boundary; a plain JMP (An) is commonly a cold
+        // function thunk, while an ordinary conditional branch can belong to
+        // an alternate path of an already-productive longer trace. Once the
+        // indexed dispatch proves this topology, retain through the LAST
+        // completed terminal: SC2K does useful work after the dispatch,
+        // including one bounded call/return, before a direct branch leads to
+        // the impossible second call.
+        let salvage = self.recording.as_ref().and_then(|recording| {
+            if recording.ops.last().is_some_and(|op| op.op.ends_trace()) {
+                return None;
+            }
+            let last_terminal = recording.ops.iter().rposition(|op| op.op.ends_trace())?;
+            let has_indexed_dispatch = recording.ops[..=last_terminal].iter().any(|op| {
+                matches!(
+                    op.op,
+                    JitTraceOp::IndirectJmp {
+                        target: JitEa::Index { .. } | JitEa::PcIndex { .. },
+                        expected_target: Some(_),
+                    }
+                )
+            });
+            if !has_indexed_dispatch || last_terminal + 1 < SALVAGE_MIN_OPS {
+                return None;
+            }
+            let continuation = recording.ops.get(last_terminal + 1)?.pc;
+            Some((last_terminal + 1, continuation))
+        });
+        if let Some((prefix_len, continuation)) = salvage {
+            self.recording
+                .as_mut()
+                .expect("salvage came from an active recording")
+                .ops
+                .truncate(prefix_len);
+            self.finish_recording(cpu, continuation, RecordingEnd::Region);
+        } else {
+            self.finish_recording(cpu, call_pc, RecordingEnd::Region);
+        }
+    }
+
     /// `call_retry_pending` marks the one case where a rescued prefix is
     /// worse than no trace at all: a recording stopped by its first
     /// unpermitted call. Salvaging there installs a compiled prefix that
@@ -2578,7 +2631,7 @@ impl TraceJit {
                     if caller_hi.wrapping_sub(caller_lo) > CALL_THROUGH_MAX_SPAN
                         || callee_hi.wrapping_sub(callee_lo) > CALL_THROUGH_MAX_SPAN
                     {
-                        self.finish_recording(cpu, executed_pc, RecordingEnd::Region);
+                        self.finish_recording_at_call_span(cpu, executed_pc);
                         return;
                     }
                     recording.pending_return = Some(return_pc);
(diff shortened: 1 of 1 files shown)


benletchford:

PR #171, opened 2026-09-04
perf(trace): specialize validation segment counts and hot byte lengths

rlanday:
> [!NOTE]
> This PR is stacked on #170. The new work begins after `212748f` and is
> kept in two commits: `2a08fdf` and `58dccde`. Once the dependency merges,
> the diff collapses to those commits.

## Summary

Every native trace entry compares the trace's captured guest-code bytes with
live memory before running the body, so that code loaded, relocated, patched
or self-modified after compilation cannot execute stale native semantics. The
check is correct and stays exactly where it was; this PR only makes its two
nested levels cheaper: dispatching over the trace's code segments, and
comparing the bytes of one segment.

On the same guest work and identical output (same simulated ticks and
framebuffer hash in every pair), the two commits together retire **1.84%
fewer host instructions on SC2K** and about 2.9% fewer on Lemmings, with the
other workloads neutral to slightly better.

## Change 1: spell out the common segment counts (`2a08fdf`)

The validator expressed all segments as `segments.iter().all(...)`. In this
hot caller LLVM retained iterator state, loop control and abstraction work on
every entry. Real traces overwhelmingly have one, two or three discontiguous
code segments:

- SC2K: 1 = 21.96%, 2 = 56.26%, 3 = 12.39%, 4+ = 9.40% of validation entries.
- Lemmings: 1 = 75.80%, 2 = 23.99%, 3+ = 0.20%.

The validator now handles 0, 1, 2 and 3 segments directly and keeps the
iterator as the general fallback for 4 or more. The point is not that
iterators are slow in general; a very hot routine with a tiny, measured shape
distribution gets simpler machine code when those shapes are written out.
An outlined special loop for 4+ segments was also tried and measured only
-0.034% on SC2K, below the bar for the extra code.

Host instructions versus the #170 baseline: SC2K **-1.178%**, Lemmings
**-2.595%**, 3 in Three -0.082%, System's Twilight -0.071%, EV Override
-0.131% per tick.

## Change 2: exact-length comparisons for the hot byte lengths (`58dccde`)

After segment dispatch, the shared outlined byte comparator remained visible.
Its generic scalar loop compares native words but pays induction, remaining-
length tests and a per-word early-out branch, and on the overwhelmingly common
unchanged-code path every early-out is a wasted branch.

The x86-64 comparator gains exact cases for the measured hot lengths 6, 8,
10, 12, 28 and 36 bytes. Each loads native-width words unaligned, XORs live
against expected, ORs all differences together and makes one equality
decision: branchless, and a mismatch in any byte makes the combined
difference nonzero. Properties:

- the cases live in the one shared outlined helper, not in generated bodies;
- no native trace ABI change; validation happens at the same times over the
  same bytes;
- the portable generic fallback is unchanged, and the existing SSE2 path
  still takes 48 through 512 bytes (hot wide lengths such as 98 and 106 keep
  using it);
- the test covers every byte position of every selected length, including an
  unaligned live pointer and the scalar/SIMD boundaries.

Optimized assembly was inspected: 6/8/10/12 become straight loads and
compares or load-XOR-OR sequences; LLVM vectorizes 28 and 36 with the scalar
tail folded into the single result.

Host instructions versus `2a08fdf` (every pair improved; identical ticks and
hash):

| Workload | Budget / pairs | Host instructions |
|---|---:|---:|
| SC2K | 400M / 3 | **-0.660%** |
| Lemmings | 200M / 3 | **-0.371%** |
| 3 in Three | 200M / 2 | -0.018% |
| System's Twilight | 100M / 2 | -0.026% |
| EV Override | 200M / 3 | -0.061% per tick |

Cumulatively the two commits are **-1.842% SC2K host instructions** versus
the #170 baseline.

## Method

Alternating headless A/B runs of release (fat-LTO) Systemless binaries that
differ only in the m68k commit, fresh fixture per run, deterministic input
scripts, macOS hardware `instructions retired` from `/usr/bin/time -l`.
Deterministic workloads must reach the same tick count and screenshot hash in
every pair; EV Override is clock-seeded and is compared per tick. Host
instructions are the acceptance metric here; cycle and time counters showed
thermal noise (Lemmings cycles rose while all three instruction counts fell).

## Verification

- `cargo test --features jit --lib`: 287 passed
- `cargo test --lib` (portable): 207 passed
- `cargo fmt --check`: passed

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -2172,6 +2172,59 @@ impl TraceJit {
         }
     }
 
+    /// Close a recording that cannot add another callee without making its
+    /// bounded code intervals overly broad, retaining a completed prefix
+    /// when it belongs to an indexed-dispatch path.
+    #[cold]
+    #[inline(never)]
+    fn finish_recording_at_call_span(&mut self, cpu: &mut CpuCore, call_pc: u32) {
+        // A trace carries one caller and one callee SMC interval. SC2K's
+        // table-driven dispatch can cross an indexed JMP, finish one
+        // call/return, and then reach a second far callee. The second call
+        // cannot join this trace without making its callee interval overly
+        // broad, but the completed control-flow prefix before it is
+        // independently useful.
+        //
+        // Keep this salvage deliberately structural. An indexed jump is a
+        // multi-way dispatch boundary; a plain JMP (An) is commonly a cold
+        // function thunk, while an ordinary conditional branch can belong to
+        // an alternate path of an already-productive longer trace. Once the
+        // indexed dispatch proves this topology, retain through the LAST
+        // completed terminal: SC2K does useful work after the dispatch,
+        // including one bounded call/return, before a direct branch leads to
+        // the impossible second call.
+        let salvage = self.recording.as_ref().and_then(|recording| {
+            if recording.ops.last().is_some_and(|op| op.op.ends_trace()) {
+                return None;
+            }
+            let last_terminal = recording.ops.iter().rposition(|op| op.op.ends_trace())?;
+            let has_indexed_dispatch = recording.ops[..=last_terminal].iter().any(|op| {
+                matches!(
+                    op.op,
+                    JitTraceOp::IndirectJmp {
+                        target: JitEa::Index { .. } | JitEa::PcIndex { .. },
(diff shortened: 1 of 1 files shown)


benletchford:

PR #172, opened 2026-09-04
perf(trace): validate generated traces once per entry

rlanday:
> [!NOTE]
> This PR is stacked on #171 (validation cost). The new work begins after
> `58dccde` and is kept in two commits: `b5bd062` and `2b04412`. Once the
> dependency merges, the diff collapses to those commits.

## Summary

For eligible traces, Cranelift now compiles a second small function, a
*checked entry*, that bounds-checks and compares the trace's fixed code
segments as constants and then calls the ordinary unchecked body. Rust
selects the checked entry only for the first native call of a `try_execute`
entry; repeated loop-body calls within the same entry keep using the
unchecked body. Validation therefore still happens exactly once, at the same
semantic boundary as before, while the segment addresses, lengths and
expected bytes are compile-time constants instead of data the shared
comparator has to load and loop over.

Eligibility is deliberately narrow: native x86-64, one to three segments, no
segment longer than 47 bytes, plus one profile-guided exception, a single
contiguous segment of exactly 50 bytes. Wider or more fragmented traces and
all portable builds keep the shared validator.

Against the exact-length baseline, identical ticks and hash in every pair:
**SC2K -0.71% host instructions** (both commits combined), Lemmings -1.53%,
others neutral. The full validation series (with the validation-cost PR) is
about **-2.55% SC2K host instructions** versus the #170 baseline.

## Why a separate entry and not validation inside every body

The more ambitious design emitted each trace's byte checks into its Cranelift
body. It was correct and helped SC2K, but it moved validation from once per
Rust trace entry to once per native body call, and those differ:

- EV Override: 1,471,241 native calls / 460,851 validation entries = 3.19
  native calls per host validation;
- SC2K: 15,953,550 / ~16,000,394 = ~1.0;
- Lemmings: 56,403,039 / 11,023,125 = 5.12.

So SC2K won while EV regressed, because the in-body check repeats inside one
already-validated entry. Every variant that tried to keep it in the body was
rejected on EV:

| Variant | SC2K host instr | EV host instr/tick | Why rejected |
|---|---:|---:|---|
| Inline all, extra ABI argument | -1.545% | +0.197% | EV regression and global call overhead |
| Inline <=47 bytes, extra ABI argument | no win retained | +0.196% | EV regression remains |
| <=47 bytes, original ABI | no win retained | +0.418% | repeated validation is worse |
| Overlapping tail words | no win retained | +0.248% | EV regression remains |
| Only 16-47-byte segments | no win retained | +0.178% | EV regression remains |
| 16-47 bytes, bodies called once per entry | -1.222% | +0.239% | topology filter insufficient |

The useful lesson was to attach one generated validation entry to the whole
dispatch unit, not to every body, so the specialized code executes once at
the same point as the host validation did. That is this design.

## How the checked entry works

Each generated segment check first proves the complete guest range lies in
fast memory, then loads the live bytes in native-width chunks, XORs them with
expected-byte immediates, ORs the differences and branches once. Overlapping
final loads cover non-word tails without reading outside the segment. All
bounds are checked before any live byte is loaded.

If the entry cannot prove validity (a byte differs, or a segment is outside
the current fastmem window) it returns a reserved validation-needed result
before touching guest state. The existing bus-aware slow validator then
distinguishes: unchanged code reachable through the bus (call the unchecked
body); changed code (invalidate the trace, preserving the original
first-changed-instruction behavior); unavailable code (decline and return to
the interpreter). An undersized instruction budget cannot enter the wrapper
and runs the shared validator first, preserving validation-before-budget
ordering. Stores made by an executing trace keep the existing code-span
guards, so repeated unchecked iterations remain safe after the once-per-entry
proof.

A true tail call from the checked entry to the body was tried: Cranelift
rejects `return_call` under the system calling convention, and switching the
trace functions to its `Tail` convention would remove the C-ABI entry point
Rust relies on, so the entry uses an ordinary call and return.

## The exact-50-byte exception (`2b04412`)

The 47-byte limit stops before the shared comparator's 48-byte SSE2 path.
Entry-weighted census instrumentation showed SC2K's remaining medium-length
validation is one stable shape:

| Census | All validation entries | Max segment 48-64 B | of which single-segment | Exactly one 50 B segment |
|---|---:|---:|---:|---:|
| SC2K 400M | 16,004,173 | 657,298 | 522,563 | **497,471** |
| EV Override 200M | 2,683,910 | 53,769 | 21,496 | **401** |

A contiguous 50-byte, 19-operation non-loop region accounts for 497,412
entries by itself. The policy therefore keeps the general limit and generates
a checked entry additionally for a single segment of exactly 50 bytes;
neighboring 48, 52 and 64-byte shapes stay on the shared SIMD path. Selected
once at compile time, no runtime policy branch.

Broader cutoffs were measured and rejected because they regress EV:

| Rule versus 1-3 segments <=47 B | SC2K | EV /tick | Decision |
|---|---:|---:|---|
| Add four-segment traces | -0.020% (3 pairs) | not run | neutral; code unjustified |
| 1-3 segments through 64 B | -0.186% (3 pairs) | **+0.098%** (6 pairs) | real EV regression |
| One segment through 64 B | -0.157% (6 pairs) | **+0.111%** (3 pairs) | EV regression remains |

Increasing generated-validation coverage is not a sound goal in itself: the
extra functions raise compile work and code footprint in a workload that gets
little validation benefit.

## Measurements

Checked entry (`b5bd062`) versus the exact-length baseline `58dccde`,
identical ticks and hash in every deterministic pair:

| Workload | Budget / pairs | Host instructions | Cycles |
|---|---:|---:|---:|
| SC2K | 400M / 3 | **-0.611%** | -0.070% |
| Lemmings | 200M / 2 | **-1.528%** | -4.742% |
| 3 in Three | 200M / 2 | -0.002% | -3.551% |
| System's Twilight | 100M / 2 | -0.020% | -2.992% |
| EV Override | 200M / 3 | -0.155% per tick | -0.497% per tick |

Exact-50 (`2b04412`) versus `b5bd062`:

| Workload | Budget / pairs | Host instructions |
|---|---:|---:|
| SC2K | 400M / 6 | **-0.103%** (all six pairs improved) |
| EV Override | 200M / 3 | -0.117% per tick (treated as neutral) |
| Lemmings | 200M / 3 | +0.033% (neutral) |
| 3 in Three | 200M / 2 | -0.009% |
| System's Twilight | 100M / 2 | +0.016% |

The fixed instruction cap includes JIT compilation, so these already include
compiling the second function. Method as in the validation-cost PR
(alternating headless A/B, release binaries differing only in the m68k
commit, hardware instruction counters, tick/hash equivalence, EV per tick).

## Verification

- `cargo test --features jit --lib`: 288 passed
- `cargo test --lib` (portable): 207 passed
- The generated-entry test mutates every byte at every eligible even segment
  length from 4 through 46 and at exactly 50, and asserts rejection before PC
  or registers change; the neighboring wide lengths are asserted to retain
  shared validation.
- The discontiguous-segment test verifies changed-code invalidation and the
  bus fallback when only the later segment lies outside fast memory.

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -2172,6 +2172,59 @@ impl TraceJit {
         }
     }
 
+    /// Close a recording that cannot add another callee without making its
+    /// bounded code intervals overly broad, retaining a completed prefix
+    /// when it belongs to an indexed-dispatch path.
+    #[cold]
+    #[inline(never)]
+    fn finish_recording_at_call_span(&mut self, cpu: &mut CpuCore, call_pc: u32) {
+        // A trace carries one caller and one callee SMC interval. SC2K's
+        // table-driven dispatch can cross an indexed JMP, finish one
+        // call/return, and then reach a second far callee. The second call
+        // cannot join this trace without making its callee interval overly
+        // broad, but the completed control-flow prefix before it is
+        // independently useful.
+        //
+        // Keep this salvage deliberately structural. An indexed jump is a
+        // multi-way dispatch boundary; a plain JMP (An) is commonly a cold
+        // function thunk, while an ordinary conditional branch can belong to
+        // an alternate path of an already-productive longer trace. Once the
+        // indexed dispatch proves this topology, retain through the LAST
+        // completed terminal: SC2K does useful work after the dispatch,
+        // including one bounded call/return, before a direct branch leads to
+        // the impossible second call.
+        let salvage = self.recording.as_ref().and_then(|recording| {
+            if recording.ops.last().is_some_and(|op| op.op.ends_trace()) {
+                return None;
+            }
+            let last_terminal = recording.ops.iter().rposition(|op| op.op.ends_trace())?;
+            let has_indexed_dispatch = recording.ops[..=last_terminal].iter().any(|op| {
+                matches!(
+                    op.op,
+                    JitTraceOp::IndirectJmp {
+                        target: JitEa::Index { .. } | JitEa::PcIndex { .. },
+                        expected_target: Some(_),
+                    }
+                )
+            });
+            if !has_indexed_dispatch || last_terminal + 1 < SALVAGE_MIN_OPS {
+                return None;
+            }
+            let continuation = recording.ops.get(last_terminal + 1)?.pc;
+            Some((last_terminal + 1, continuation))
+        });
+        if let Some((prefix_len, continuation)) = salvage {
+            self.recording
+                .as_mut()
+                .expect("salvage came from an active recording")
+                .ops
+                .truncate(prefix_len);
+            self.finish_recording(cpu, continuation, RecordingEnd::Region);
+        } else {
+            self.finish_recording(cpu, call_pc, RecordingEnd::Region);
+        }
+    }
+
     /// `call_retry_pending` marks the one case where a rescued prefix is
     /// worse than no trace at all: a recording stopped by its first
     /// unpermitted call. Salvaging there installs a compiled prefix that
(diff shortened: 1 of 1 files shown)


benletchford:

PR #173, opened 2026-09-04
perf(trace): report guarded exits and specialize watch checks

rlanday:
> [!NOTE]
> This PR is stacked on #172 (generated checked entry). The new work begins
> after `2b04412` and is kept in two commits: `ff582bf` and `8aa8956`. Once
> the dependency merges, the diff collapses to those commits.

## Summary

This removes two pieces of redundant Rust-side work from the hot compiled-
trace entry path:

1. A generated conditional-branch or computed-jump guard now reports its own
   side exit in the existing packed return value. Rust no longer scans the
   trace's guarded operations and reconstructs the branch outcome from `PPC`
   and `PC` after every partial native return.
2. The interior-watch check spells out the measured zero-through-four-watch
   cases and retains the generic iterator fallback for larger embedding watch
   lists. Each watch is rejected cheaply by the trace's two code intervals
   before the exact operation list is inspected.

Both changes preserve the interpreter/native boundary and all architectural
state. They only make information that is already known at the point of use
cheaper to consume.

On the longer SC2K confirmation, the complete change reduces actual host work
by **2.313% retired instructions, 1.596% CPU cycles, and 1.708% user CPU
time**, at the same simulated tick and framebuffer hash. Lemmings improves by
**2.405% retired instructions**. 3 in Three, System's Twilight, and EV
Override show no substantive regression.

## Motivation

Recent trace work has increased the amount of useful guest code that reaches
the JIT. That makes the fixed work around each native trace invocation more
important. An SC2K host sample showed both of these helpers directly under
`TraceJit::try_execute`:

- the monomorphized `Iterator::any` used in the watch path: 122 top-of-stack
  samples;
- `CompiledTrace::is_guarded_branch_exit`: 100 top-of-stack samples.

That sample had 7,253 main-thread samples. Sampling attribution is only a lead,
not the acceptance metric, so both changes were isolated with production
hardware-counter A/B runs below.

### Previous behavior

A native trace returns one packed `u64`:

```text
63                              32 31                 0
+--------------------------------+--------------------+
| retired guest instruction count| status + cycles    |
+--------------------------------+--------------------+
```

Bit 31 of the low word already meant "the requested trace execution completed."
A return without that bit could mean several different things:

- a recorded conditional branch went the other way;
- a guarded indirect jump produced a different target;
- a fast-memory/window check failed;
- an alignment or memory-operation precondition failed; or
- a store might overlap the trace's guest code.

Only the first two are adaptive control-flow exits. They feed trace-exit
candidacy and branch-bias re-recording; treating a memory bailout as one would
change JIT policy.

The old code therefore called `CompiledTrace::is_guarded_branch_exit` after a
partial native return. That helper:

1. walked a `u128` bitmap of guarded trace operations;
2. used each set bit to load a `TraceBuildOp` from the retained operation
   vector;
3. compared the operation's captured PC with the CPU's post-return `PPC`;
4. matched conditional branch versus indirect jump;
5. reconstructed the branch target and fallthrough; and
6. compared the committed `PC` with the recorded prediction.

Retirement count alone cannot identify the exiting operation. `CondSkip`
makes retirement data-dependent, and a guard on the last operation can retire
the trace's full numeric operation count while still being a side exit. The
old PPC/PC reconstruction was correct, but it recovered information the
generated guard had known at the exact return site.

### New packed-result layout

The change reserves bit 30 of the low word:

```text
63                              32 31 30 29           0
+--------------------------------+--+--+---------------+
| retired guest instruction count| C| G| guest cycles  |
+--------------------------------+--+--+---------------+

C = completed request
G = partial return caused by a guarded control-flow mismatch
```

Generated guarded conditional branches and guarded indirect jumps OR `G` into
their packed result only on the mismatch edge. Memory, window, alignment, and
self-modification bailouts leave it clear. The portable executor sets the same
bit, keeping native and portable policy behavior aligned.

Rust classifies a guard exit with two bit tests. The production
`CompiledTrace` no longer retains the `guarded_ops` bitmap solely for
post-return reconstruction. The old classifier and bitmap remain in native
JIT tests as an independent oracle for the new generated tag.

### Interaction with generated code validation

The parent checked-entry change already reserves a distinctive
validation-needed sentinel. It now uses both status bits (`C | G`) and is
recognized by exact equality before the native result is interpreted.
`trace_return_guarded_branch_exit` additionally requires `C` to be clear, so a
validation miss cannot be mistaken for an adaptive side exit.

Two low-word metadata bits leave 30 bits for cycles. A single counted native
call is therefore capped at `(1 << 30) - 1` cycles. This is the same ceiling
already used by normal `run_batch` calls (`i32::MAX / 2`); the explicit cap
also makes direct callers with a larger positive cycle budget safe. If more
work remains, the existing outer loop invokes the body again. Guest cycle
accounting and the 32-bit retirement payload are unchanged.

### Direct tests

Native tests inspect the unmasked packed result and prove that:

- a completed `CondSkip` trace does not carry `G`;
- a mismatching generated conditional branch does carry `G`;
- every tested guarded indirect-jump addressing form agrees with the retained
  PPC/PC classifier for both matching and mismatching targets; and
- the validation-needed sentinel is never classified as a guard exit.

Existing tests also prove that memory bailouts do not create exit candidacy
and that last-operation guard exits remain distinguishable even when their
retirement count equals the trace length.

## Change 2: specialize the tiny interior-watch list

`run_batch` must not let a compiled multi-operation trace jump across a watched
PC. Systemless normally supplies a tiny list:

- PC zero, used as the clean top-level RTS exit;
- optionally an interrupt callback resume PC; and
- zero or more pending native trap-handler return PCs.

Both Systemless call sites reuse a `Vec` preallocated for four entries. The
public m68k API remains a general slice and can legally receive more.

Previously every compiled-trace entry used nested generic iteration:

```rust
watch_pcs.iter().any(|&watched| {
    let masked = cpu.address(watched);
    let in_code = (masked >= trace.code_start && masked < trace.code_end)
        || (masked >= trace.callee_start && masked < trace.callee_end);
    in_code && trace.ops.iter().skip(1).any(|op| op.pc == watched)
})
```

The overwhelmingly common sole watch is zero, which is outside application
code, yet it still entered the generic iterator/closure machinery on every
trace attempt.

The new `contains_interior_watch` operation first performs the same two compact
code-interval comparisons. Only an overlapping address scans exact operation
PCs. The caller matches slice lengths 0, 1, 2, 3, and 4 directly; lengths above
four retain the original iterator fallback.

This is not a watch-set approximation:

- no watch is dropped or truncated;
- the trace entry PC remains intentionally excluded, matching the old
  `skip(1)` behavior;
- an interior PC must still match an actual captured operation, not merely a
  byte in a broad range;
- address masking behavior is unchanged; and
- the general fallback preserves arbitrary embedder watch-list sizes.

## Relationship to the earlier 1/2/3-segment specialization

This is complementary to, not a replacement for, the earlier validation
specialization.

The earlier work optimized `Iterator::all` over captured guest-code segments:
one, two, and three segments are validated directly, with a four-or-more
fallback. That saved 1.178% host instructions in SC2K and 2.595% in Lemmings.

This PR operates at two different points:

- it removes post-return control-flow reconstruction; and
- it optimizes `Iterator::any` over embedding watch PCs.

The tested baseline already contains the 1/2/3 segment cases and all later
generated-validation work. The new gains therefore stack on them. In isolated
SC2K 400M measurements, guard tagging saved 0.882% and watch specialization
saved another 0.756%; the complete direct comparison saved 1.614%. Their
independent multiplicative expectation is 1.632%, so the effects are nearly
additive within normal production-run variation.

### Method

- Host: Intel Core i9-9880H, macOS 26.5.1, x86-64.
- Embedder: Systemless 0.28.5, release profile with fat LTO and one codegen
  unit, using the current open m68k PR stack.
- Baseline m68k head: `2b04412` (`generate the hot 50-byte validator`).
- Candidate m68k head: `8aa8956`.
- Each arm starts from a fresh copy of the same application fixture and save
  directory.
- Deterministic input scripts are used where needed.
- A/B order alternates between pairs.
- `/usr/bin/time -l` supplies retired host instructions, CPU cycles, and user
  CPU time.
- Retired host instructions are the primary low-noise measure of actual host
  work. The longer runs additionally establish that the reduction translates
  to cycles and CPU time.
- Fixed-instruction runs also include JIT compilation cost.

Exact artifacts:

- baseline binary SHA-256:
  `27dcc9dbee7c9dbe0add46437b408272c115d81bc15699a38202555d0a9a0487`
- candidate binary SHA-256:
  `460ab5dd1b09e954255d57e557912fe9e40e1a31a2b64af1c79fe3df18106407`

### Complete candidate versus parent

| Workload | Budget / pairs | Retired host instructions | CPU cycles | User CPU time | Correctness endpoint |
| --- | ---: | ---: | ---: | ---: | --- |
| SC2K, normal city | 400M / 3 | **-1.614%** | -0.580% | 0.000% | identical 185,810 ticks/hash |
| SC2K, normal city, long | 2B / 2 | **-2.313%** | **-1.596%** | **-1.708%** | identical 359,148 ticks/hash |
| Lemmings | 200M / 3 | **-2.405%** | -0.768% | -1.220% | identical 36,409 ticks/hash |
| 3 in Three, long | 1B / 2 | -0.113% | -0.679% | -0.391% | identical 4,116,775 ticks/hash |
| System's Twilight | 100M / 3 | -0.087% | -0.382% | -1.891% | identical 685,705 ticks/hash |
| EV Override | 200M / 3 | +0.081%/tick | +0.311%/tick | -2.112%/tick | variable ticks; normalized by aggregate ticks |

EV is conservatively neutral. Its raw retired-instruction total decreased by
0.042%, but the candidate arms advanced 0.123% fewer simulated ticks, yielding
the reported +0.081% per-tick value. That is below the repeatable effects in
the target workloads and is not treated as a win.

The 400M SC2K run is startup-heavy. At 2B, the unchanged startup/compile share
is diluted and the steady-state savings become clearer. Both long pairs agree:

| Pair | Retired host instructions | CPU cycles | User CPU time |
| --- | ---: | ---: | ---: |
| A then B | -2.299% | -1.601% | -1.732% |
| B then A | -2.327% | -1.591% | -1.684% |

### Component isolation on SC2K 400M

| Delta | Pairs | Retired host instructions | CPU cycles | User CPU time |
| --- | ---: | ---: | ---: | ---: |
| Guard-exit tag versus parent | 3 | **-0.882%** | -0.364% | -0.359% |
| Small-watch specialization versus guard tag | 3 | **-0.756%** | -0.819% | -0.232% |
| Both versus parent | 3 | **-1.614%** | -0.580% | 0.000% |

Short-run CPU time is quantized/noisy; the 2B comparison above is the relevant
confirmation that fewer retired instructions translate into lower CPU use.

### Interesting negative-looking short result

The first 3 in Three 200M matrix retired 0.116% fewer host instructions but
showed +0.232% cycles and +0.585% user time. Rather than hiding or accepting
that ambiguity, the workload was extended to two alternating 1B pairs. The
retired-instruction result reproduced (-0.113%), while cycles (-0.679%) and
user CPU time (-0.391%) both moved in the favorable direction. The short
timing movement was measurement noise, not a sustained microarchitectural
regression.

### Reconstruct the exiting operation from retirement count

This is not correct. Conditional skip operations retire a data-dependent
number of guest instructions, and a guarded last operation can report a count
equal to the complete trace length. PPC/PC reconstruction was required before
the new tag.

### Return a larger struct or write an out-parameter

Both would perturb the native trace ABI and add register or memory traffic to
every call. The existing low word has provably unused status space, so a
single bit carries the information without another memory dependency.

### Truncate or force a fixed-size watch API

Systemless's observed lists are tiny, but the m68k embedder API is general.
Changing the API or silently limiting watches would be a semantic regression.
The specialized cases therefore retain the exact arbitrary-length fallback.

### Replace byte validation with page generations

Page versioning was prototyped separately through five designs: eager page
counters, a lazy dirty-page journal, a guarded bounding interval, and a sparse
two-hash page filter. They were correct but put work on every native guest
store in order to save occasional trace-entry comparisons.

On SC2K, eager/lazy barriers regressed retired instructions by roughly 6--8%.
The guarded-range prototype was +3.2% even before useful reuse and became
+34.3% when a sparse 6.3 MiB false-positive interval was active. The sparse
two-hash filter still cost +2.38% disabled and +10.21% active. Those
experiments remain preserved on `experiment/page-versioned-trace-validation`;
they are not part of this PR.

A future invalidation design must avoid a software barrier on ordinary stores.
Plausible materially different sources are authoritative loader/resource
generations or OS page protection/write-watch with a portable fallback. Until
one can cover native JIT stores, decoded/interpreter stores, and host/HLE bus
writes without a hot universal tax, generated byte validation remains safer
and faster.

## Verification

- `cargo test --lib`: 207 passed
- `cargo test --features jit`: all unit, integration, architecture, batch, and
  doc suites passed; 288 JIT library tests passed
- Systemless `cargo test`: 4,762 library tests passed (3 ignored), 79 runner
  tests passed, and binary, toolbox-showcase, and doc tests passed
- Deterministic application A/B coverage: SC2K, Lemmings, 3 in Three,
  System's Twilight, and EV Override

Raw benchmark logs are retained under:

- `.work/steady/logs/game-cpu-guard-exit-tag-sc2k-400m`
- `.work/steady/logs/game-cpu-guard-exit-watch-small-sc2k-400m`
- `.work/steady/logs/game-cpu-guard-exit-tag-watch-small-vs-exact50-sc2k-400m`
- `.work/steady/logs/game-cpu-guard-exit-watch-small-sc2k-2b`
- `.work/steady/logs/game-cpu-guard-exit-watch-small-lemmings-200m`
- `.work/steady/logs/game-cpu-guard-exit-watch-small-three-1b`
- `.work/steady/logs/game-cpu-guard-exit-watch-small-twilight-100m`
- `.work/steady/logs/game-cpu-guard-exit-watch-small-ev-200m`

### DIFF
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -1381,6 +1381,18 @@ impl CompiledTrace {
         }
         false
     }
+
+    /// Whether a watch lies on an instruction the trace would execute after
+    /// its entry. Checking the two compact code intervals first keeps the
+    /// operation scan off the normal path (most embedding watches are return
+    /// sentinels outside application code).
+    #[inline(always)]
+    fn contains_interior_watch(&self, cpu: &CpuCore, watched: u32) -> bool {
+        let masked = cpu.address(watched);
+        let in_code = (masked >= self.code_start && masked < self.code_end)
+            || (masked >= self.callee_start && masked < self.callee_end);
+        in_code && self.ops.iter().skip(1).any(|op| op.pc == watched)
+    }
 }
 
 fn guarded_op_mask(ops: &[TraceBuildOp]) -> u128 {
@@ -1638,12 +1650,32 @@ impl TraceJit {
             // entry PC is intentionally excluded: run_batch does not check
             // watches on entry, and self-loop entry watches are handled by
             // `single_iter` after one complete iteration.
-            if watch_pcs.iter().any(|&watched| {
-                let masked = cpu.address(watched);
-                let in_code = (masked >= trace.code_start && masked < trace.code_end)
-                    || (masked >= trace.callee_start && masked < trace.callee_end);
-                in_code && trace.ops.iter().skip(1).any(|op| op.pc == watched)
-            }) {
+            // Systemless normally supplies one to four watches (the zero-PC
+            // sentinel plus callback/native-trap returns). Spell those small
+            // cases directly: generic Iterator::any was measurable even when
+            // the sole watch was zero and failed the cheap code-range test.
+            let interior_watched = match watch_pcs {
+                [] => false,
+                [a] => trace.contains_interior_watch(cpu, *a),
+                [a, b] => {
+                    trace.contains_interior_watch(cpu, *a) || trace.contains_interior_watch(cpu, *b)
+                }
+                [a, b, c] => {
+                    trace.contains_interior_watch(cpu, *a)
+                        || trace.contains_interior_watch(cpu, *b)
+                        || trace.contains_interior_watch(cpu, *c)
+                }
+                [a, b, c, d] => {
+                    trace.contains_interior_watch(cpu, *a)
+                        || trace.contains_interior_watch(cpu, *b)
+                        || trace.contains_interior_watch(cpu, *c)
+                        || trace.contains_interior_watch(cpu, *d)
+                }
+                watches => watches
+                    .iter()
+                    .any(|&watched| trace.contains_interior_watch(cpu, watched)),
+            };
+            if interior_watched {
                 return None;
             }
             if trace.needs_window && cpu.fm_len == 0 {
--- a/src/core/trace_jit.rs
+++ b/src/core/trace_jit.rs
@@ -102,17 +102,29 @@ pub(crate) const TRACE_PC_NONE: u32 = u32::MAX;
 /// (`i32`), so cycles never need this bit. The upper 32 retirement bits stay
 /// fully available for data-dependent retirement such as `CondSkip`.
 const TRACE_RETURN_COMPLETE: u64 = 1 << 31;
+/// A partial return caused by a recorded control-flow prediction missing.
+/// The generated branch knows this at the return site; carrying the reason
+/// avoids recovering it in Rust by scanning the trace's guarded operations.
+const TRACE_RETURN_GUARDED_BRANCH_EXIT: u64 = 1 << 30;
 /// A checked native entry returns this before touching guest state when its
 /// generated code-byte comparison cannot prove the trace current. Rust then
 /// uses the bus-aware slow path to distinguish changed code from an
 /// unavailable fastmem mapping.
 #[cfg(all(feature = "jit", not(target_family = "wasm")))]
-const TRACE_RETURN_VALIDATION_NEEDED: u64 = TRACE_RETURN_COMPLETE | (1 << 30);
-const TRACE_RETURN_CYCLES_MASK: u32 = i32::MAX as u32;
+const TRACE_RETURN_VALIDATION_NEEDED: u64 =
+    TRACE_RETURN_COMPLETE | TRACE_RETURN_GUARDED_BRANCH_EXIT;
+#[cfg(any(not(feature = "jit"), target_family = "wasm", test))]
+const TRACE_RETURN_METADATA: u64 = TRACE_RETURN_COMPLETE | TRACE_RETURN_GUARDED_BRANCH_EXIT;
+const TRACE_RETURN_CYCLES_MASK: u32 = TRACE_RETURN_GUARDED_BRANCH_EXIT as u32 - 1;
 
 #[inline]
 fn trace_return_complete(packed: u64) -> bool {
     packed & TRACE_RETURN_COMPLETE != 0
+}
+
+#[inline]
+fn trace_return_guarded_branch_exit(packed: u64) -> bool {
+    !trace_return_complete(packed) && packed & TRACE_RETURN_GUARDED_BRANCH_EXIT != 0
 }
 
 #[inline]
@@ -1282,6 +1294,7 @@ struct CompiledTrace {
     /// Trace entry checks this before inspecting `ops`, so the common
     /// guard-free return is constant-time; guarded traces visit only their
     /// guarded operations instead of scanning the complete recorded path.
+    #[cfg(all(test, feature = "jit", not(target_family = "wasm")))]
     guarded_ops: u128,
     /// A recorded interior branch is a path prediction eligible for adaptive
     /// rerecording. Cleared after the one allowed rerecord so completed traces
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Reviewed the complete cumulative stack #166–#173, including each incremental diff: exact instruction-byte comparisons and their bounds/tails, nested-call stack writes and retirement, live indirect-jump guards, indexed memory updates and self-modification checks, prefix salvage, generated checked entries and fallback, and packed exit/watch handling.

The local ARM64 run passes 323 library tests and 45 batch tests with jit/serde/trace-profile, plus 207 library tests and 45 batch tests with default features. The nested-CPU callback regression test also passes here. GitHub's x86-64 Build & Test passes formatting, clippy, documentation, default/all-feature tests and packaging; that run covers the x86-only generated validators which this ARM64 host cannot execute.

I will integrate this cumulative stack through this PR using rebase so its individual Conventional Commits remain intact, then close #166–#172 as included. #174 remains separate because of the reproduced default-path reentrancy regression. #175's independent inlining changes will be rebased off that blocked dependency.

